PR #37, opened 2018-08-05
fix(statObj): Makes lookahead objects compatible with the statObj alg…

jesperstarkar:
…orithm


jesperstarkar:
Setting lookahaead objects to isAbstract is _one_ way of making it compatible witht the statObj calculations. @Julusian please confirm that this works with lookAhead.

(spoiler alert: it does - in xpro)

PR #51, opened 2018-11-05
feat: unsynced mode

jstarpl:
This pull request makes a ro delete request not delete an RO from the database, but instead marks it as "unsynced". If an RO is unsynced, all subsequent MOS pushes are ignored until the user "syncs" the RO again, which causes a pull from MOS.

As a result, ROs need to be manually removed from DB using a button in the RO list.


jstarpl:
OK, I've created a method called isAvailableForMOS which checks for unsynced status and logs if RO is unsynced. Unsynced mode will only be used if the RO is active.

PR #52, opened 2018-11-06
Feature/manual casparcg record

Julusian:


meteor/client/ui/Settings/StudioSettings.tsx
```diff
@@ -1206,6 +1207,9 @@ class StudioMappings extends React.Component<Translated<IStudioMappingsProps>, I
 
 		return (
 			_.map(this.props.studioInstallation.mappings, (mapping: Mapping , layerId: string) => {
+				// If an internal mapping, then hide it
+				if (mapping.internal) return <React.Fragment key={layerId}></React.Fragment>
```
jstarpl: This probably performs better, but it would be cleaner to just filter the `.mappings` array

meteor/client/ui/TestTools/RecordingView.tsx
```diff
@@ -0,0 +1,86 @@
...
+			this.subscribe('studioInstallations', {
+				_id: this.props.match.params.studioId
+			})
+		}
+	}
+
+	renderRecordingView () {
+		const { t, file, studio } = this.props
+
+		if (!file) return <React.Fragment></React.Fragment>
```
jstarpl: Can also be `return null`, which is cleaner and will perform better.

meteor/client/ui/TestTools/RecordingsList.tsx
```diff
@@ -0,0 +1,290 @@
...
+		return true
+	}
+
+	renderControlPanel () {
+		const { t } = this.props
+
+		if (!this.isStudioConfigured()) {
+			return <React.Fragment>
+				<p>{t('A required setting is not configured')}</p>
+			</React.Fragment>
```
jstarpl: No need to return a `React.Fragment`, just return the `p`. `React.Fragment` is only needed when returning multiple, same-depth elements.

meteor/client/ui/TestTools/RecordingsList.tsx
```diff
@@ -0,0 +1,290 @@
...
+								</th>
+							</tr>
+						</thead>
+						<tbody>
+							{this.renderRecordingList()}
+						</tbody>
+					</table>
+				</div>
+			</div>
+		</React.Fragment>
```
jstarpl: Like before, just return the `div`. I don't think we expect to return any elements at the same level as the `div`.

meteor/client/ui/TestTools/RecordingsList.tsx
```diff
@@ -0,0 +1,290 @@
...
+					<td className='recorded-file-list-item__stopped'>
+						{this.props.file.stoppedAt && <Moment format='YYYY/MM/DD HH:mm:ss'>{this.props.file.stoppedAt}</Moment>}
+					</td>
+					<td className='actions'>
+						<button className='action-btn' onClick={(e) => this.props.onDeleteRecording(this.props.file)}>
+							<FontAwesomeIcon icon={faTrash} />
+						</button>
+					</td>
+				</tr>
+			</React.Fragment>
```
jstarpl: See previous comment.

meteor/client/ui/Settings/StudioSettings.tsx
Julusian: Mappings is an object not an array, so _.filter mangles the result so that we no longer have the layerId

jstarpl: OK, I guess we can leave this as-is.

PR #55, opened 2018-11-08
Feature: runtime arguments arguments

jstarpl:
This feature allows to set runtime arguments (as key: value items) on a segment line. Setting them is triggered using a shortcut key and changing the runtime arguments causes a re-run of the SL blueprint using the cached MOS data.

If a runtime argument is used, a given segment line is marked as `dirty` and the blueprint will be run again on any SL reset (RO reset, set as next, etc.) and the `dirty` flag will be cleared.

### DIFF
--- a/meteor/server/api/playout.ts
+++ b/meteor/server/api/playout.ts
@@ -1635,6 +1635,9 @@ methods[PlayoutAPI.methods.roTake] = (roId: string) => {
 methods[PlayoutAPI.methods.userRoTake] = (roId: string) => {
 	return ServerPlayoutAPI.userRoTake(roId)
 }
+methods[PlayoutAPI.methods.roToggleSegmentLineArgument] = (roId: string, slId: string, property: string, value: string) => {
+	return ServerPlayoutAPI.roToggleSegmentLineArgument(roId, slId, property, value)
+}
 methods[PlayoutAPI.methods.roSetNext] = (roId: string, slId: string) => {
 	return ServerPlayoutAPI.roSetNext(roId, slId, true)
 }
--- a/meteor/server/api/playout.ts
+++ b/meteor/server/api/playout.ts
@@ -44,11 +44,11 @@ import * as _ from 'underscore'
 import { logger } from '../logging'
 import { PeripheralDevice,PeripheralDevices,PlayoutDeviceSettings } from '../../lib/collections/PeripheralDevices'
 import { PeripheralDeviceAPI } from '../../lib/api/peripheralDevice'
-import { IMOSRunningOrder, MosString128 } from 'mos-connection'
+import { IMOSRunningOrder, MosString128, IMOSROFullStory } from 'mos-connection'
 import { PlayoutTimelinePrefixes, LookaheadMode } from '../../lib/api/playout'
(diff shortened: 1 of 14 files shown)


meteor/server/api/playout.ts
```diff
@@ -1505,6 +1548,53 @@ export namespace ServerPlayoutAPI {
 
 		updateTimeline(runningOrder.studioInstallationId)
 	}
+	export const roToggleSegmentLineArgument = syncFunction(function roToggleSegmentLineArgument (roId: string, slId: string, property: string, value: string) {
```
nytamin: a toggle-function is not ideal, because that requires the caller to know something about the value of the property, before it's about to change it.
Could we remove the "toggle" functionality (and if you want to remove an runtime-argument, you pass the value null instead)?

meteor/server/api/templates/templates.ts
```diff
@@ -70,6 +70,9 @@ export type SegmentLineItemOptional = Fix<SegmentLineItem>
 export type SegmentLineItemOptional = Optional<SegmentLineItem>
 export type SegmentLineAdLibItemOptional = Optional<SegmentLineAdLibItem>
 export type RunningOrderBaselineItemOptional = Optional<RunningOrderBaselineItem>
+export type TemplateRuntimeArguments = {
+	[key: string]: string
```
nytamin: (optional/question) could this be a number also? or _any_?

meteor/server/api/templates/templates.ts
```diff
@@ -70,6 +70,9 @@ export type SegmentLineItemOptional = Fix<SegmentLineItem>
 export type SegmentLineItemOptional = Optional<SegmentLineItem>
 export type SegmentLineAdLibItemOptional = Optional<SegmentLineAdLibItem>
 export type RunningOrderBaselineItemOptional = Optional<RunningOrderBaselineItem>
+export type TemplateRuntimeArguments = {
```
nytamin: this looks like it should be an _interface_ instead of a _type alias_
https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c

jstarpl: Yeah, good point!

meteor/server/api/templates/templates.ts
jstarpl: I guess so. I don't know enough about the new blueprint system to know if that makes sense. With plain JavaScript & the value of the Runtime argument being essentially user input (it's just two text fields in the settings), I think assuming this is a string is a safer choice.

nytamin: agreed :)

meteor/server/api/playout.ts
jstarpl: I guess, though this better maps to the action performed by the user - pretty much like roTake does various things depending on the holdState of the current segment line. The user just expects to have a single action that toggles a given argument. Is your argument that the current state check should be done against what is in the GUI, because that is the state that the user is actually toggling (as to what's server-side, which just might be something else)?

jstarpl:
I've merged develop with this branch so it should now be easily mergeable with develop.

PR #57, opened 2018-11-15
Fix/migrations

jesperstarkar:
**NOT READY TO BE MERGED**

_Look, I did something..._

Opening PR to have review and critique going.

See comments in my self-review for aspects I'm not sure about

### DIFF
--- a/meteor/lib/collections/PeripheralDevices.ts
+++ b/meteor/lib/collections/PeripheralDevices.ts
@@ -64,7 +64,17 @@ export enum PlayoutDeviceType { // to match DeviceType in TSR
 	PANASONIC_PTZ = 5,
 	// TCPSEND = 6, // to be implemented
 	HYPERDECK = 7,
-	PHAROS = 8
+	PHAROS = 8,
+	SCANNER = 9,
+	LAUNCHER = 10
+}
+export const DefaultPlayoutDevicePort = {
+	1: 5250,	// CASPARCG
+	2: 9910,	// ATEM
+	3: 9000,	// LAWO
+	4: 9993,	// HYPERDECK
+	9: 8000,	// SCANNER
+	10: 8005	// LAUNCHER
 }
 
 export interface PlayoutDeviceSettings {
--- a/meteor/lib/collections/StudioInstallations.ts
+++ b/meteor/lib/collections/StudioInstallations.ts
@@ -34,8 +34,8 @@ export interface Mappings {
 export interface Mapping {
 	device: PlayoutDeviceType,
 	lookahead: LookaheadMode,
-	deviceId: string
-	// [key: string]: any
+	deviceId: string,
+	[key: string]: any
 }
 export interface MappingCasparCG extends Mapping {
 	device: PlayoutDeviceType.CASPARCG,
--- a/meteor/server/databaseMigration.ts
+++ b/meteor/server/databaseMigration.ts
@@ -22,7 +22,7 @@ import { logger } from '../lib/logging'
 import { Optional } from '../lib/lib'
 
 /** The current database version, x.y.z */
-export const CURRENT_SYSTEM_VERSION = '1.0.0'
+export const CURRENT_SYSTEM_VERSION = '0.16.0'
 /** In the beginning, there was the database, and the database was with Sofie, and the database was Sofie.
  * And Sofie said: The version of the database is to be GENESIS_SYSTEM_VERSION so that the migration scripts will run.
  */
--- a/meteor/server/databaseMigration.ts
+++ b/meteor/server/databaseMigration.ts
@@ -76,12 +76,14 @@ export function addMigrationStep (step: MigrationStep) {
  * @param version
  * @param steps
  */
-export function addMigrationSteps (version: string, steps: Array<MigrationStepBase>) {
+export function addMigrationSteps (version: string, steps: Array<MigrationStepBase | null>) {
 	_.each(steps, (step) => {
-		addMigrationStep(_.extend(step, {
-			id: version + step.id,
-			version: version
-		}))
+		if (step) {
+			addMigrationStep(_.extend(step, {
+				id: version + step.id,
+				version: version
+			}))
+		}
 	})
 }
 
--- a/meteor/server/databaseMigration.ts
+++ b/meteor/server/databaseMigration.ts
@@ -15,11 +15,11 @@ import {
 	MigrationStepInputResult,
 	GetMigrationStatusResultNoNeed,
 	GetMigrationStatusResultMigrationNeeded,
-	RunMigrationResult
+	RunMigrationResult,
 } from '../lib/api/migration'
 import { setMeteorMethods } from './methods'
 import { logger } from '../lib/logging'
-import { Optional } from '../lib/lib'
+import { Optional, objectPathSet } from '../lib/lib'
 
 /** The current database version, x.y.z */
 export const CURRENT_SYSTEM_VERSION = '0.16.0'
@@ -247,7 +247,7 @@ export function runMigration (
 			let stepInput: MigrationStepInputFilteredResult = {}
 			_.each(inputResults, (ir) => {
 				if (ir.stepId === step.id) {
-					stepInput[ir.attribute] = ir.value
+					objectPathSet(stepInput, ir.attribute, ir.value)
 				}
 			})
 
--- a/meteor/server/databaseMigration.ts
+++ b/meteor/server/databaseMigration.ts
@@ -22,7 +22,7 @@ import { logger } from '../lib/logging'
 import { Optional, objectPathSet } from '../lib/lib'
 
 /** The current database version, x.y.z */
-export const CURRENT_SYSTEM_VERSION = '0.16.0'
+export const CURRENT_SYSTEM_VERSION = '1.0.0'
 /** In the beginning, there was the database, and the database was with Sofie, and the database was Sofie.
  * And Sofie said: The version of the database is to be GENESIS_SYSTEM_VERSION so that the migration scripts will run.
  */
--- a/meteor/server/databaseMigration.ts
+++ b/meteor/server/databaseMigration.ts
@@ -79,6 +79,7 @@ export function addMigrationStep (step: MigrationStep) {
 export function addMigrationSteps (version: string, steps: Array<MigrationStepBase>) {
 	_.each(steps, (step) => {
 		addMigrationStep(_.extend(step, {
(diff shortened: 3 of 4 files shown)


meteor/lib/collections/PeripheralDevices.ts
```diff
@@ -64,7 +64,17 @@ export enum PlayoutDeviceType { // to match DeviceType in TSR
 	PANASONIC_PTZ = 5,
 	// TCPSEND = 6, // to be implemented
 	HYPERDECK = 7,
-	PHAROS = 8
+	PHAROS = 8,
+	SCANNER = 9,
+	LAUNCHER = 10
+}
+export const DefaultPlayoutDevicePort = {
```
jesperstarkar: @nytamin I don't like this pattern. Unsure what will be the cleanest way.

meteor/lib/collections/StudioInstallations.ts
```diff
@@ -34,8 +34,8 @@ export interface Mappings {
 export interface Mapping {
 	device: PlayoutDeviceType,
 	lookahead: LookaheadMode,
-	deviceId: string
-	// [key: string]: any
+	deviceId: string,
+	[key: string]: any
```
jesperstarkar: @nytamin you had commented out this, I added it back. Can't see better ways of getting typings for all device types.

meteor/server/databaseMigration.ts
```diff
@@ -76,11 +76,14 @@ export function addMigrationStep (step: MigrationStep) {
...
  */
-export function addMigrationSteps (version: string, steps: Array<MigrationStepBase>) {
+export function addMigrationSteps (version: string, steps: Array<MigrationStepBase | null>) {
 	_.each(steps, (step) => {
-		addMigrationStep(_.extend(step, {
-			version: version
-		}))
+		if (step) {
+			addMigrationStep(_.extend(step, {
+				id: version + step.id,
```
jesperstarkar: @nytamin as discussed

meteor/server/databaseMigration.ts
```diff
@@ -244,7 +247,7 @@ export function runMigration (
 			let stepInput: MigrationStepInputFilteredResult = {}
 			_.each(inputResults, (ir) => {
 				if (ir.stepId === step.id) {
-					stepInput[ir.attribute] = ir.value
+					objectPathSet(stepInput, ir.attribute, ir.value)
```
jesperstarkar: @nytamin THIS I'm a bit afrais of. Basically, any device IDs with `.` in them will break, I guess? Scary

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -102,6 +102,127 @@ function ensureCollectionProperty<T = any> (
...
+ * Convenience function to generate basic test
+ * øparam type
+ */
+function ensurePlayoutDevicesHostAndPort (
+	deviceType: PlayoutDeviceType = -1
+): MigrationStepBase | null {
+	let collection: Mongo.Collection<any> = Collections.PeripheralDevices
+	if (!collection) throw new Meteor.Error(404, `Collection PeripheralDevices not found`)
+
+	let allDevices = {}
```
jesperstarkar: @nytamin Since we have this nested structure, things get a bit tangled........

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -102,6 +102,127 @@ function ensureCollectionProperty<T = any> (
...
+			devices = _.pick(devices, (device) => device.type === deviceType)
+		}
+		allDevices[playoutGw._id] = devices
+	})
+
+	// iterate over devices for gateways
+	_.each(allDevices, (devices, key) => {
+		allDevices[key] = _.pick(devices, (device) => {
+			let needsMoreData = false
+			let foundHostProp = objectPathGet(device, 'options.host', undefined)
```
jesperstarkar: @nytamin same at :132 The MAJOR question is to require host and port fields to be present at all in the database before failing the validation. As of now only non-empty strings passes hosts, and only numbers > 0 passes ports.

If we DON'T require an initial entry to even consider validating the device for host or port, we would need a map of device types to search for. This will align better now that I'm introducing the DefaultPlayoutDevicePort constant, which I didn't have as I originally wrote this validation check.

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -102,6 +102,127 @@ function ensureCollectionProperty<T = any> (
...
+		},
+		input: () => {
+			// let objects = collection.find(selector).fetch()
+			let inputs: Array<MigrationStepInput> = []
+			
+			_.each(allDevices, (devices: {[key:string]: PlayoutDeviceSettingsDevice}, gatewayId: string) => {
+				_.each(devices, (device: PlayoutDeviceSettingsDevice, deviceId: string) => {
+
+					// should have host added
+					let foundHostProp = objectPathGet(device, 'options.host', undefined)
```
jesperstarkar: @nytamin Repeating myself here, but I don't see better patterns as we have this nested structure with potentially multiple gateways fucking up better structures. And: repeating things twice is not an anti pattern to me ;)

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -102,6 +102,127 @@ function ensureCollectionProperty<T = any> (
...
+					// should have port added
+					let foundPortProp = objectPathGet(device, 'options.port', undefined) 
+					// port not number or <= 0
+					if(foundPortProp !== undefined && (isNaN(foundPortProp) || foundPortProp <= 0)) {
+						inputs.push({
+							label: `Playout device "${deviceId}" on gateway "${gatewayId}" is missing port`,
+							description: `Port for device "${deviceId}"`,
+							inputType: 'int',
+							attribute: `${gatewayId}.${deviceId}.options.port`,
+							defaultValue: DefaultPlayoutDevicePort[device.type] || 0
```
jesperstarkar: @nytamin see asana task on default values not working (never making it to the migration)

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -102,6 +102,127 @@ function ensureCollectionProperty<T = any> (
...
+		migrate: (input: MigrationStepInputFilteredResult) => {
+			_.each(input, (newDevices, objectId: string) => {
+				let gateway = collection.findOne(objectId)
+				
+				let newOptions = _.pick(newDevices, newSettings => !_.isEmpty(newSettings))
+
+				
+				let devices = objectPathGet(gateway, 'settings.devices', {})
+				_.each(newOptions, (options, deviceId) => {
+					let mergedDevice = _.extend(devices[deviceId], options)
```
jesperstarkar: @nytamin a bit unsure here. there's probably better ways.

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -597,102 +735,102 @@ addMigrationSteps( '0.1.0', [
...
 					PeripheralDevices.update(device._id, {$set: {
-						'settings.devices.casparcg0': {
+						'settings.devices.caspar01': {
 							type: PlayoutDeviceType.CASPARCG,
 							options: {
-								host: '127.0.0.1',
-								port: 5250,
-								launcherHost: '127.0.0.1',
-								launcherPort: 8010, // todo: change this
+								host: '',
```
jesperstarkar: @nytamin ref question on initiating values or if we could omit this in total (would need us to have a map for validation)

meteor/server/databaseMigrationSystemSteps.ts
```diff
@@ -910,48 +1033,82 @@ addMigrationSteps( '0.16.0', [
...
-		assignHotkeysToGlobalAdlibs: false,
-		unlimited: true
-	}),
-	ensureMapping('hyperdeck0', literal<MappingHyperdeck>({
-		device: PlayoutDeviceType.HYPERDECK,
-		deviceId: 'hyperdeck0',
-		mappingType: MappingHyperdeckType.TRANSPORT,
+
+	// ensures all playoutdevices have hosts and ports
+	// @todo: add launcher and scanner host and port checks
```
jesperstarkar: @nytamin waiting for launcher and scanner to become new devices.

meteor/lib/collections/PeripheralDevices.ts
nytamin: Yeah, I believe we should add the scanner and launcher as peripheral-devices. That gives us the flexibility of running Sofie with or without launcher & media-scanner

meteor/lib/collections/StudioInstallations.ts
nytamin: I believe the reason I commented it out was to keep a strict typing. If you want to do a specific typing you should use the MappingLawo, MappingAtem etc, not the Mapping itself.
If you add a `[key: string]: any` you essentially disable type checks for properties in that interface, which is bad.

jesperstarkar: I know, but I didn't make the effort. 
You still enforce the deviceID and other strict ones, so that part of the interface still holds.

meteor/server/databaseMigration.ts
nytamin: I'm not convinced this should be added at the "framework"-level.
I'd rather have it at the "user"-level (ie where the steps are defined)

meteor/server/databaseMigration.ts
```diff
@@ -76,11 +76,14 @@ export function addMigrationStep (step: MigrationStep) {
  * @param version
  * @param steps
  */
-export function addMigrationSteps (version: string, steps: Array<MigrationStepBase>) {
+export function addMigrationSteps (version: string, steps: Array<MigrationStepBase | null>) {
```
nytamin: Irregular whitespace not allowed, thats a &amp;nbsp;, not a space ;)

meteor/lib/collections/PeripheralDevices.ts
nytamin: This looks strange and has a non-typed dependency to PlayoutDeviceType.
It would be better to replace it with an enum, or perhaps a
```
function getDefaultPort (deviceType: PlayoutDeviceType) {
  switch(deviceType) {
    case PlayoutDeviceType.CASPARCG: return 5250
    [...]
  }
}
```

nytamin:
There's a big refactoring going on in #60, which has made this PR obsolete.
I believe it would be easiest to open a new fresh PR based on that in a few days time, when things have calmed down.

PR #64, opened 2019-01-07
fix: Migration for stk&full audio remapping

jesperstarkar:
Adds a missing commit from R4 to master.

The current 18.1 on master is missing this commit.


jesperstarkar:
Also, the git tree becomes much cleaner by merging this.

Julusian:
This wasn't merged as in R5 the migration is in the blueprints instead. 
Part of https://github.com/nrkno/tv-automation-sofie-blueprints/commit/afd067d62739cd49318cf20d2dad0d2de1c121c6

jesperstarkar:
So this is invalid and should be closed?

Julusian:
I didn't realise this was to master and not develop. it probably should be merged to master

jesperstarkar:
Yeah, it is to actually release R4, which we haven't fully done

PR #66, opened 2019-01-23
fix: F5/F6 adlib lib

Julusian:


Julusian:
@nytamin Any problems with the new fields added to SegmentLineItem? They are set on existing sli when running an adlib, and cleared as part of reset

PR #75, opened 2019-03-12
Feature/http watcher

stephan-nordnes-eriksen:
This PR should go together with https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/70

When the State Timeline Resolver is published, we need to update the state-timeline-resolver-types library to whatever the newest will be.


stephan-nordnes-eriksen:
This PR still needs the newest version of TSR to be published before it will work. Please don't merge before that is in.

jstarpl:
I've updated TSR types to the latest version but I still get:
```
client/ui/Settings/components/PlayoutDeviceSettingsComponent.tsx (195, 28): Property 'HTTPWATCHER' does not exist
on type 'typeof DeviceType'.
```

Is there a relase of the types that we still need to publish first?

PR #77, opened 2019-03-18
Feature/better evaluations

jesperstarkar:
Asana tasks:
- "1353-Better-looking slack messages"
- "User evaluation of shows as a metric for graphing in grafana"


jesperstarkar:
Might wait for the "better looking slack messages" subtask before merging.

jesperstarkar:
@nytamin this should be merged to R8

PR #91, opened 2019-07-08
Fix broken migration scripts.

bevand10:
Without the fix, anything involving snapshots (e.g. database migrations and therefore a functional system) is broken.

There's probably a better, more "Sofie", way to fix this, but as I don't know no better, this works for me.

### DIFF
--- a/meteor/server/api/snapshot.ts
+++ b/meteor/server/api/snapshot.ts
@@ -314,6 +314,9 @@ function storeSnaphot (snapshot: {snapshot: SnapshotBase}, comment: string): str
 	if (!system) throw new Meteor.Error(500, `CoreSystem not found!`)
 	if (!system.storePath) throw new Meteor.Error(500, `CoreSystem.storePath not set!`)
 
+	// Override the default value of 'undefined'.
+	system.storePath = process.env.STOREPATH
+	
 	let fileName = fixValidPath(snapshot.snapshot.name) + '.json'
 	let filePath = Path.join(system.storePath, fileName)
 


nytamin:
Thank you for this PR!
I believe the fix you proposed won't make the snapshots working, the issue you're having might instead be related to the value you have set in your Mongo-db, in the CoreSystem collection, the storePath property.

I instead suggest that the fixes to be made are:
1) make the storePath optional in the typings for CoreSystem
2) making changes to allow snapshots to work without having the storePath set.
3) add an UI to edit the CoreSystem settings.

I'll leave this PR open during the summer (as most of the team is on vacation, we won't act on this right now), and we'll put something together in August.

bevand10:
Hi.

The fix is working. See discussion in slack in the bbc_rd channel. Without it I have a completely broken Sofie - it refuses to run any migrations, so as soon as you load blueprints of any form, errors occur.

I can see the value in step 2 of your suggestion, but that would involve `system.storePath` having a sensible default value of something other than `undefined`.

I did look in `meteor/server/coreSystem.ts` and it usefully says: https://github.com/nrkno/tv-automation-server-core/blob/a416b0d2d5701221b6f319c8eef0244d61f1a18f/meteor/server/coreSystem.ts#L43

But the value isn't then set anywhere else to the best of my knowledge.

nytamin:
Ah, yes it seems like we are in a catch 22 situation, and should definitely take action to remedy this.

Julusian:
I am a little confused as to how this fix works, as 2 lines above it is throwing an error if the value is not set. But if it works then go with it for now :)

storePath is normally configured in the very first round of migrations, and I am sure that it would get correctly last I tried. But I can see that there are ways it could be not set, so I agree that it should be made more durable.

@nytamin As a default we could use a temp folder `/tmp/sofie-snapshots` or platform equivalent? Either that it just skips doing snapshots.

nytamin:
I think that it would be better (as I suggested above) to just make it optional.

bevand10:
Going back to the very 1st migration log line I encountered:

`Error when taking snapshot:Error: ENOENT: no such file or directory, open 'undefined/System_2019-06-24_10_57_33.json'`

So it's been with me from the very beginning. I'll drop mongo and start over to see if the err persists in a from-nothing scenario.

bevand10:
For completeness of this log, I deleted my mongodb and created a fresh one.

On Sofie load, I was prompted to run the very first migration (note, no blueprints have been loaded yet).

1st off, I get these messages when using the 'Migrate' button:
```
Step "Playout-gateway exists": Something went wrong, validation didn't approve of the changes. The changes have been applied, but might need to be confirmed. (Validation error: Playout-device is missing on studio0)
Step "Mos-gateway exists": Something went wrong, validation didn't approve of the changes. The changes have been applied, but might need to be confirmed. (Validation error: Mos-device is missing on studio0)
Step "studioConfig.slack_evaluation": Something went wrong, validation didn't approve of the changes. The changes have been applied, but might need to be confirmed. (Validation error: slack_evaluation is missing on studio0)
Step "Studios.settings.mediaPreviewsUrl": Something went wrong, validation didn't approve of the changes. The changes have been applied, but might need to be confirmed. (Validation error: settings.mediaPreviewsUrl is missing on studio0)
Step "Studios.settings.sofieUrl": Something went wrong, validation didn't approve of the changes. The changes have been applied, but might need to be confirmed. (Validation error: settings.sofieUrl is missing on studio0)
```

Checking *mongo*, this is where the fault starts, and then we're broken for all time:

```> db.coreSystem.find()
{ "_id" : "core", "created" : 1562606948427, "modified" : 1562606948427, "version" : "0.0.0", "previousVersion" : null, "storePath" : "undefined" }
```

jstarpl:
Bumping this since the vacations are over and this should be a fairly easy fix. I'm wondering if this is maybe actually an issue with the migrations input system. If the input from the input system is undefined, this would actually serialize as "undefined" and would pass the validation for the storePath step in 0_16_0.ts:

```validate: () => {
let system = getCoreSystem()
if (!system) return 'CoreSystem not found!'
if (!system.storePath) return 'CoreSystem.storePath not set!'
if (!_.isString(system.storePath)) return 'CoreSystem.storePath must be a string!'
if (system.storePath.slice(-1) === '/') return 'CoreSystem.storePath must not end with "/"!'
   return false
}
```

I think that it could be as simple as doing:
```
migrate: (input) => {
   setCoreSystemStorePath(input.storePath || '')
},
```
in line 28-30

nytamin:
I believe what @jstarpl suggested above is correct. I pushed a [fix for this](https://github.com/nrkno/tv-automation-server-core/commit/34a693ff804407d6f5f7b147030b2df6824e7ffd) that will be released in R14, that should resolve the issue.

Please reopen this if the issue if the problem persists after R14.

PR #100, opened 2019-09-06
Update README.md with nodes version info

olzzon:
It can be hard to find out that the reason for Sofie not running is the Node version.
So i´ve updated the README with this info


olzzon:
After looking at latest documentation:
https://sofie.gitbook.io/sofie-tv-automation/documentation/installation/installing-sofie-server-core

It would make better sense to reference to that page in the read me.

nytamin:
Yeah, we've done some major updates to the documentation the last few days. We'll get around updating the README in this repo with proper references shortly :)

olzzon:
Closing this PR as major documentation updates are underway.

PR #103, opened 2019-10-01
fix: database migration step for service messages

gundelsby:


gundelsby:
> 
> 
> Hmm, I don't think this is in the right migration file - this is supposed to be a part of R13+ and 1_0_0.ts file is for R12. There is a X_X_X.ts file for "next release" migrations, so maybe put it there?

Right 🤦‍♂ 
Good thing I made a PR for this :D

gundelsby:
Also, @jstarpl already fixed this ten days ago when he merged the other PR, so I'll just close this without merging as there's no need for this.

gundelsby:
Also, @jstarpl already fixed this ten days ago when he merged the other PR, so I'll just close this without merging as there's no need for this.

PR #104, opened 2019-10-14
Fix/segment removal mos delete

nytamin:
This PR primarily contains a fix for the issue:
_When the first story in a segment is floated, the segment is prompty deleted (along with the parts), then recreated.
The currently playing Part could then be removed and recreated but without the timings, which cause things like auto-next to fail._

Along with the fix, I added some snapshot checks in the tests for the mos-methods, to avoid any regressions.


meteor/server/api/ingest/mosDevice/ingest.ts
```diff
@@ -464,109 +465,182 @@ function diffAndApplyChanges (
...
+			..._.keys(segmentDiff.changed)
+		]
+	)
+}
 
-	afterIngestChangedData(rundown, _.map(segmentDiff.rankChanged, i => getSegmentId(rundown._id, newIngestRundown.segments[i.newRank].externalId)))
+export interface SegmentEntries {
+	[segmentExternalId: string]: SegmentEntry
+}
+export interface SegmentEntry extends IngestSegment {
```
Julusian: this interface should be dropped or tidied

meteor/server/api/ingest/mosDevice/ingest.ts
```diff
@@ -464,109 +465,182 @@ function diffAndApplyChanges (
...
+		if (segmentEntries[ingestSegment.externalId]) {
+			throw new Meteor.Error(500, `compileSegmentEntries: Non-unique segment external ID: "${ingestSegment.externalId}"`)
+		}
+		segmentEntries[ingestSegment.externalId] = {
+			// id: ingestSegment.externalId,
+			// name: ingestSegment.name,
+			...ingestSegment,
+			// rank: rank // QUESTION: use rank or ingestSegment.rank?
+			// parts: _.map(ingestSegment.parts, p => p.externalId)
+		}
```
Julusian: this can be tidied

PR #107, opened 2019-10-29
Release 13 patch

nytamin:
* ~fix issue with timings in GUI~ fix issue with distributing timing information in the GUI
* fix issue with disappearing part when updating some part durations
* fix issue with mediaObjects, causing missing media notifications in GUI


jstarpl:
Seems to be fine, my only objection would be to the "fix issue with timings in GUI" which may be misleading. I'd rephrase it as "fix issue with distributing timing information in the GUI".

PR #111, opened 2019-11-12
fix: change UI settings to a Meteor-standard way

jstarpl:
**What is happening?**
UI needs some installation-specific settings to be provided. So far, we essentially used a static object (the rest of the code was dead for all intents and purposes), hardcoded in the code for providing the system framerate to be used for displaying things like Timecodes and such.

**What should happen?**
It should be possible to modify these settings on a per-installation basis, without modifying the sourcecode.

**What this PR changes?**
This PR removes the dead code and relies on a standard Meteor.settings functionality to provide Settings to the UI. The Settings object is now also strongly typed.


meteor/lib/Settings.ts
```diff
@@ -1,25 +1,22 @@
...
-} catch (e) {
-	localSettingsEnv = {}
+const DEFAULT_SETTINGS: ISettings = {
+	"frameRate": 25,
+	"defaultToCollapsedSegments": false,
+	"autoExpandCurrentNextSegment": false
 }
 
-export const Settings = _.extend({}, defaultSettings, localSettings, localSettingsEnv)
+Settings = _.extend({}, DEFAULT_SETTINGS)
```
nytamin: Just a minor note, this isn't type safe as _.extend returns an `any` type. What you probably want to do instead:

`Settings = _.clone(DEFAULT_SETTINGS)`

jstarpl: Good tip! I'll change that to `_.clone`.

PR #112, opened 2019-11-13
Feature/r13 prompter tyholt

jesperstarkar:
Feature that adds support for using devices like the Contour Shuttle and Xkeys with jog/shuttle inputs as prompter controllers. This implementation uses keyboard (ascii) inputs from the devices just like a normal keyboard, hence operating like a normal keyboard and can easily be controlled the same way from any device that outputs the same set of shortcuts.

A list of shortcuts are added to the readme. Also the new usage of the mode-setting has it's new changes documented.

* **Other information**:


nytamin:
@jesperstarkar I've added these changes:
* Changed name of "mode" parameter to "controlmode"
* Added "new" parameter "mode" to switch which devices to use
* Updated docs

See https://github.com/nrkno/tv-automation-server-core/tree/feature/r13-prompter-tyholt#operating-the-prompter-screen

Could you have a go and see that it still works as intended?

PR #113, opened 2019-11-18
System migrations

nytamin:
This PR adds the feature to run system migrations before blueprint-blueprints.

It also deals with several issues that came up during implementation:
* Added tests for initial migrations
* Remove requirement of MOS-gateway. Since there are other ingest gateways that could be used, it doesn't make sense to require this one during installation.

TODO:
- [x] Add to tests: mocked blueprints (system, studio, showStyle) which exposes migrations
- [x] Make Core/system migrations run before blueprints
- [x] Define in which order blueprints migrations are run
- [x] Tests


nytamin:
Note: The amount of commits in this PR looks a bit wierd, due to it being based on the /release15 branch, which hasn't been merged into master yet at the time of writing. Hopefully all that will be better when it has been merged.

Julusian:
Reading through a comparison of this branch against the release15 branch makes it very easily readable :)

PR #117, opened 2019-11-24
Part._ranks unique within segment

nytamin:
This PR changes how the `Part._rank` is defined.

Before: 
`Part._rank` was unique within the whole rundown, so when moving a segment, the `Part._rank`s within had to be updated as well.

After:
`Part._rank` is unique only within the segment.
When moving a Segment, the parts doesn't have to be updated.

Todo:

- [ ] Test this feature in test environment


Julusian:
I don't see lookahead as being updated here, so I expect that will have been broken

Julusian:
@nytamin missed another case https://github.com/nrkno/tv-automation-server-core/pull/117/files?file-filters%5B%5D=.ts#diff-ea81f831abd74b065e9c2e8d151e4ca4R873

meteor/lib/lib.ts
```diff
@@ -771,15 +782,15 @@ export const caught: <T>(v: Promise<T>) => Promise<T> = (f => p => (p.catch(f),
  * Blocks the fiber until all the Promises have resolved
  */
 export const waitForPromiseAll: <T>(ps: Array<Promise<T>>) => Array<T> = Meteor.wrapAsync(function waitForPromises<T> (ps: Array<Promise<T>>, cb: (err: any | null, result?: any) => T) {
```
Julusian: This function now freezes execution.
There is a call to Meteor.wrapAsync still and its callback is never called, so it never returns.

Perhaps this is what was intended?
```
export function waitForPromiseAll<T>(ps: Array<Promise<T>>): Array<T> {
	return waitForPromise(Promise.all(ps))
}
```

nytamin:
Thanks @Julusian, I added a few places in 3ee38fe252cc0c807ac102a6d2cbff983dc9c5b2 that I must have missed before.

Julusian:
@nytamin You need to cherry pick https://github.com/nrkno/tv-automation-server-core/commit/bd8b893a9d8158fcd10018c1409541aea95fba35
I was finding in the part instances branch that every segment was showing every part. I narrowed it down to an optimisation you made which this commit fixes.

Julusian:
The failing test can be fixed with https://github.com/nrkno/tv-automation-server-core/pull/150/commits/f75eeed55d09b983004d3c2f7050a78079bf211a

PR #122, opened 2019-11-28
Automatically rewind segment that stopped being live

jstarpl:

This PR introduces an optional automatic rewind of the segment timeline when leaving a given segment.

Currently, when leaving the segment, the timeline is left in the state as-is, until the SegmentTimeline component is destroyed (when it is outside of the viewport), when the entire component is rebuilt with the scroll position at 0.

When `autoRewindLeavingSegment` setting in the public settings is turned on, the segment will be rewound after it stops being live.

* **Other information**:


jesperstarkar:
I LOVE THIS

PR #123, opened 2019-11-28
Update ReadMe

jstarpl:

This PR updates docs in the `README.md` file.

Some of the functionality is not described in the ReadMe & some of the settings are no longer used. 

This PR updates the `README.md` file with information about new features and changes in the UI settings.

* **Other information**:


jstarpl:
Hmm... That is an interesting discussion. I agree that it would make documenting it much easier, but then it requires you to look at the source code, even if you don't want to do any development. I suppose, since we publish the docker images on GitHub, it wouldn't be a bad thing if all the necessary documentation would be available 'on the outside', and you would be able to just download the image and run. I think that would be much more 'friendly' towards prospective users.

PR #128, opened 2019-12-03
Fix/next on float bug

nytamin:
This PR fixes an issue with the Next point being wrongly moved to the beginning of the rundown when floating a Part in the current playing segment.

* Having a segment with three parts: A, B & C.
* A is currently playing, B is Next.
* C is removed from the NRCS
* The Next point now is moved to the beginning of the rundown

The Next-point shouldn't be moved at all.

* **Other information**:
I've concluded that this bug was due to the `ensureNextPartIsValid` being called multiple times, one before the `updatePartRanks` had run, resulting in a bad operation on parts with wrong ranks.
This is a bug that would have been prevented by #117 , but I think it's still good to clean up the function calls a bit.

**Status**


jstarpl:
Should this be merged in?

PR #136, opened 2019-12-09
chore: make codeControl unit tests slower, decreasing the risk of them failing

nytamin:
This PR increases some durations for timeouts in the tests for codeControl.

It's of course no ideal to have to depend on actual times in unit tests, but since we're testing code that handles concurrency so...

* **Other information**:

**Status**


Julusian:
@nytamin why is it that we are using real timers in this test? Can we really not mock it all?

nytamin:
It might be possible to mock the timers, now that I know that the non-mocked tests pass 🙂

nytamin:
I'm merging this in now, and I've also added to our internal backlog to have a look at mocking the timers some time later.

PR #142, opened 2019-12-16
feat: ignore piece media status

Julusian:
Adds option to VT pieces to allow for media status checks to be ignored

Core assets will show a full status of meaningless issues (eg freeze frames during vignett). These are not useful or relevent to the producer and just add noise

The blueprints can tell core to skip these checks, and always say that the media is online and valid

* **Other information**:
blueprints-integration: https://github.com/nrkno/tv-automation-sofie-blueprints-integration/commit/a7fc8c8254009a14c8e1650675d18c0402395f1b

**Status**

### DIFF
--- a/meteor/lib/mediaObjects.ts
+++ b/meteor/lib/mediaObjects.ts
@@ -115,7 +115,8 @@ export function checkPieceContentStatus (piece: IBlueprintPieceGeneric, sourceLa
 	let metadata: MediaObject | null = null
 	let message: string | null = null
 
-	if (sourceLayer) {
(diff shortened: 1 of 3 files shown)


Julusian:
It appears to treat both of those values the same. Perhaps it should be changed though for clarity, and to be safer?

jstarpl:
Well, I think it's not the same, although for the sake of ease of implementation they are _for now_ displayed the same. However, I think we should keep UNKNOWN and OK as separate states, the same way we have OK and UNKNOWN separate for PeripheralDevices. I think by default, the returned code should be UNKNOWN, unless some layer-specific procedure gives a positive confirmation of availability or that a given piece is unavailable.

PR #143, opened 2019-12-18
feat: add validateBlueprintConfig and notifications in the rundown view

Julusian:
This will make core validate that all the required blueprint config fields are defined for the rundown being viewed.

It is up to the blueprints to do this check. This means that every segment in the show will end up logging a warning for each missing field.

Core now does the check at the same interval as checking if the configuration has changed since the rundown has been imported. It will then present up to 2 notifications to warn of the missing fields.

This is much less repetitive and easier to read then before. Additionally, any blueprint hooks wont log the same warnings to the core log, which will help clean those up a bit.

* **Other information**:

The config missing isnt a major problem, as the blueprints do define defaults for them all, but it indicates that something has been configured wrong. For this reason it is a warning, and doesnt want to be too intrusive as the show will likely work just fine without.

Partner change in blueprints: https://github.com/nrkno/tv-automation-sofie-blueprints/pull/28

**Status**

### DIFF
--- a/meteor/client/ui/RundownView/RundownNotifier.ts
+++ b/meteor/client/ui/RundownView/RundownNotifier.ts
@@ -3,7 +3,7 @@ import * as _ from 'underscore'
 import { Meteor } from 'meteor/meteor'
 import { Tracker } from 'meteor/tracker'
 import { NotificationCenter, NotificationList, NotifierHandle, Notification, NoticeLevel } from '../../lib/notifications/notifications'
-import { RundownAPI } from '../../../lib/api/rundown'
+import { RundownAPI, RundownPlaylistValidateBlueprintConfigResult } from '../../../lib/api/rundown'
 import { WithManagedTracker } from '../../lib/reactiveData/reactiveDataHelper'
 import { reactiveData } from '../../lib/reactiveData/reactiveData'
 import { checkPieceContentStatus, getMediaObjectMediaId } from '../../../lib/mediaObjects'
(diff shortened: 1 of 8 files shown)


meteor/client/ui/RundownView/RundownNotifier.ts
```diff
@@ -486,6 +490,48 @@ class RundownViewNotifier extends WithManagedTracker {
 				this._rundownImportVersionStatusDep.changed()
 			}
 		})
+
+		// Verify the showstyle & studio config look good
+		Meteor.call(RundownAPI.methods.validateBlueprintConfig, rundownId, (err: Error, configErrors: { studio: string[], showStyle: string[] } | undefined) => {
+			console.log(configErrors)
```
nytamin: no `console.log`s please ;)

Julusian:
This sounds like a larger discussion for how we want to treat things like this in general. Because if the config is going to be strict and fail if something looks wrong, then we should start doing the same for mappings and other things.
But also then we should not allowing playing a rundown if the config has changed since rundown import, as again that could result in broken playout

But I dont think we should be hard failing for this. Ultimately this config belongs to the blueprints, so I think that if it should fail, then the blueprints should be explicitly deciding that instead of this minimal check in core, as one not being set is no worse than being set to a bad value. The blueprints can already do that if they want to.
So if we do a strong invalid check, it should be done completely by the blueprints and should take some steps to decide if the values also look ok (which will likely be hard, messy and error prone).

Additionally, a hard fail could make things worse. In the unlikely event that something is changed between rundown import and running the show, this would cause any metadata sending to fail as well as rundown updates, making a potentially incorrect show be very broken part way through.

I am open to emitting a warning if the config is invalid at the time of execution, but it should be done in a way that is not shown to the user. This will be easy as the check is run by core.

With the current version of this PR, the blueprints are already not validating the config they receive. They do the exact same processing for optional and required fields.

From all this, I propose that we do:
* check-of-configs should affect systemStatus
* the check-of-configs have to be made before executing blueprints and log it

nytamin:
I agree, let's do that!

Julusian:
@nytamin I have reworked this to cover the points proposed

PR #144, opened 2019-12-19
Fix/ui performance

nytamin:
This PR makes some optimizations in `Part.getNotes`, `Segment.getNotes` in order to reduce the amount of DB calls. This gives two improvements:
   * Decrease the number of reactive dependencies (improves Tracker performance)
   * Make fewer, but larger calls to DB (increases performance)

A call to getResolvedSegment could take up to 60ms (when tested locally)

No behavioural change, a call to getResolvedSegment reduced to ~19ms.

* **Other information**:
This PR builds upon https://github.com/nrkno/tv-automation-server-core/pull/117

**Status**


nytamin:
I only benchmarked this client-side, and I have not tested the functionality yet either :)

meteor/lib/collections/Rundowns.ts
```diff
@@ -163,17 +163,64 @@ export class Rundown implements DBRundown {
...
+		}, { sort: { _rank: 1 } })
+
+		const segments = await pSegments
+		return {
+			segments: segments,
+			parts: Rundown._sortParts(await pParts, segments)
+		}
+	}
+	/** Synchronous version of getSegmentsAndParts, to be used client-side */
+	getSegmentsAndPartsSync (): { segments: Segment[], parts: Part[] } {
```
jstarpl: I'm starting to have second thoughts about this: won't such generic queries cause Trackers to refresh whenever there is any change in any of the parts, not just when Parts in the segment change? Also, wouldn't it make sense to filter parts at this stage instead of doing two passes though the Parts arrays first with the rundown filter and then with the segment filter?

meteor/lib/Rundown.ts
```diff
@@ -122,24 +122,23 @@ export function getResolvedSegment (showStyleBase: ShowStyleBase, rundown: Rundo
 
 	// fetch all the parts for the segment
 	let partsE: Array<PartExtended> = []
-	const parts = segment.getParts()
 
-	if (parts.length > 0) {
+	const segmentsAndParts = rundown.getSegmentsAndPartsSync()
+	const parts = segmentsAndParts.parts
+	// let segments = segmentsAndParts.segments
+	const partsInSegment = _.filter(parts, p => p.segmentId === segment._id)
```
jstarpl: Like this: this is essentially an unnecessary pass though the array, at least from the UIs point of view. This could have been filtered at the query stage.

Julusian:
@nytamin @jstarpl Is this pr still relevant or is it 'replaced' by https://github.com/nrkno/tv-automation-server-core/pull/220 ?

nytamin:
It is no longer relevant, as #220 implements most of the things addressed in this PR.

PR #146, opened 2019-12-23
Feature: automatic code formatting

gundelsby:
* Automatic code formatting
* **What is the current behavior?** Some formatting is done via tslint, but it's not automated on check-in. Therefore no codestyle is enforced on the code base.
* **What is the new behavior (if this is a feature change)?** Code will be automatically formatted when running tslint and/or via a pre-commit hook. This enforces a consistent codestyle across the project.
* **Other information**:
Code formatting is done by Prettier. Prettier will be integrated with tslint, meaning that formatting errors will be reported (and if possible autofixed) just as linting errors.

**Status**


Julusian:
I would recommend ignoring package.json (and I guess package-lock.json) from prettier, as prettier and yarn disagree (unsure about npm, but it likely will), and so they keep fighting over where to place a `}` in the file.

Didn't we discuss having a wider printWidth? I have been going for 120 in other projects, and am happy with that

gundelsby:
@Julusian yeah, ignoring the package*.json files is a good suggestion, I'll just do that.

The printWidth could be wider, but then again having them somewhat narrow makes it a better fit for people working with multi column layout in their IDEs. So I've been a bit back and forth on what I think is the best solution.

Julusian:
OK, so the best method is to probably just pick something and change it if later if we dont like it after using that for a while

Julusian:
https://github.com/nrkno/tv-automation-server-core/blob/3ad4e41bad76850f74210a425a12a7c832821f19/meteor/package.json#L133
The `tslint-config-standard` package is no longer needed here.

gundelsby:
> Formatting looks good, the only thing left is to revert the commit that applies this formatting to all of the files. This PR should only contain the config changes, not the resulting formatting changes.
Yeah, it's because of the merge from master again after enabling pre-commit auto-formatting. It formats (or rather lint --fix) all files that are staged.

I'm not gonna merge this branch ever, but when the time comes I'll copy the configuration files to a new branch and put everything in one commit for easy cherry picking to existing feature branches.r

PR #147, opened 2019-12-27
Automatic formatting test #1

gundelsby:


gundelsby:
At this point (f228dc138abbb38fd248bf269f7f4b4fe93d78e2), all test suites fail.

gundelsby:
Tests pass now, unsure why the coverage has gone down 🤔

Julusian:
Coverage decreasing makes sense to me as this increases the linecount by 23000.
It is likely because of all the throws that were written on the same line as an if statement

gundelsby:
> 
> 
> Coverage decreasing makes sense to me as this increases the linecount by 23000.
> It is likely because of all the throws that were written on the same line as an if statement

Yeah, that makes sense. Good thinking :)

gundelsby:
Closing this, created another branch for testing which is based on a newer version of master.

PR #148, opened 2020-01-02
fix: don't emit status.BAD for messages in queue. 

nytamin:
This PR changes what the system-status will be when there are unsent items in the externalMessageQueue.
We no longer consider these to be critical to the broadcast.

System status is BAD when there are more than 10 items on queue
System status is WARNING_MINOR when there are less than 10 items on queue

System status is WARNING_MAJOR when there are any items on queue.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/ExternalMessageQueue.ts
+++ b/meteor/server/api/ExternalMessageQueue.ts
@@ -202,9 +202,7 @@ function updateExternalMessageQueueStatus (): void {
 				const messagesOnQueueExample = messagesOnQueueCursor.fetch()[0]
 				status = {
 					statusCode: (
-						messagesOnQueueCount > 10 ?
-						StatusCode.BAD :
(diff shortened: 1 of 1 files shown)


nytamin:
Hmm, this is a tough one to figure out what's right or not...
I pushed some more situations to consider, namely the ones with NEGATIVE duration.
Let's discuss this further!

jesperstarkar:
@nytamin should we keep the check limited to >1 to avoid the alerts blinking every time some message gets queued before succeeding for that split second?

jesperstarkar:
...mind you, I haven't tested this, it is just a theory from reading the code.

nytamin:
@jesperstarkar I was unclear in my description, a message shows up in the queue only after 3 failed re-tries, so it _should™_ not flicker.

PR #151, opened 2020-01-14
Fix/meteor1.9

nytamin:
Since Meteor 1.9 has now been released, it's time for us to upgrade the platform to it.
This gives us several improvements, including Node 12 and Typescript 3.7+

This is should not affect any behaviour of the system.
However, when upgrading a system to this, mongodb will have to be updated.

* **Other information**:

**Status**


Julusian:
@jstarpl what did you find was wrong with moment? 
with your change moment is failing to import in all of the unit tests. https://circleci.com/gh/nrkno/tv-automation-server-core/2240?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

jstarpl:
> @jstarpl what did you find was wrong with moment?
> with your change moment is failing to import in all of the unit tests. https://circleci.com/gh/nrkno/tv-automation-server-core/2240?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


Well, during runtime, what was actually imported was undefined. The change in how moment module is imported allowed the GUI to run instead of crashing.

Julusian:
> > @jstarpl what did you find was wrong with moment?
> > with your change moment is failing to import in all of the unit tests. [circleci.com/gh/nrkno/tv-automation-server-core/2240?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link](https://circleci.com/gh/nrkno/tv-automation-server-core/2240?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)
> 
> Well, during runtime, what was actually imported was undefined. The change in how moment module is imported allowed the GUI to run instead of crashing.

Strange, because jest is now finding the import to be undefined.. I didnt try using sofie so I cant say if it was only jest having the problem. If so it could be something in tsconfig needs changing (meteor does ignore most/all of that)

Julusian:
Due to how meteor runs typescript, it is swallowing any type errors causing the build to pass even though they are numerous.
To help prevent this, I have added a new 'lint' job to circleci which runs `tsc --noEmit` (typecheck only) and then `tslint` (as that wasnt being run before)

This should help enforce validity and linting (especially relevant after #146)

meteor/tsconfig-base.json
```diff
@@ -0,0 +1,28 @@
...
+		],
+		"lib": ["dom", "es6", "dom.iterable", "scripthost", "es2017", "es2018"]
+	},
+	"include": [
+		"client/**/*",
+		"server/**/*",
+		"lib/**/*"
+	],
+	"exclude": [
+		"client/**/*",
```
Julusian: @nytamin this looks wrong..

nytamin: ah, yeah that's definitely wrong. I have probably pushed some temporary code..

Julusian:
@nytamin Instead of that mock wrapper around moment, have you tried adding `"esModuleInterop": true,` to tsconfig? I think that should resolve it.
In case you arent aware, meteor will ignore all (or close to it) of any tsconfig in the repo, as it needs control over the code generation portions to ensure it will work as expected. Which explains the difference experienced with jest

Julusian:
Replaced by https://github.com/nrkno/tv-automation-server-core/pull/239

PR #157, opened 2020-01-16
fix: improve UI performance

jstarpl:

This PR is supposed to limit the amount of re-renders by changing how props are distributed to timeline components, changing some components to PureComponents and doing shallow and deep comparisons in a couple of places for props in shouldComponentUpdate.

GUI rendering takes a lot of CPU time due to unnecessary re-renders.

Significant performance increase should be observable and framerate should be increased.

* **Other information**:

**Status**


jstarpl:
I think this is basically ready to be merged. The changes are quite substantial, so it would be good to get it in before any Meteor/deps updates, etc.

PR #160, opened 2020-01-20
Hotfix for looping between 2 parts

baltedewit:
This PR introduces two fixes in order to prevent a bad state where the system will keep looping between 2 parts. One is the prevention of doing a manual take shortly before an autotake and the other is to prevent agressive measures when the playout gateway reports the wrong part has started playback.

**Status**

### DIFF
--- a/meteor/server/api/playout/playout.ts
+++ b/meteor/server/api/playout/playout.ts
@@ -235,10 +235,11 @@ export namespace ServerPlayoutAPI {
 
 				const offset = currentPart.getLastPlayOffset()
 				if (start && offset && currentPart.expectedDuration) {
-					const t = Date.now() - start! + offset!
+					// date.now - start = playback duration, duration + offset gives position in part
+					const playbackDuration = Date.now() - start! + offset!
 
 					// If there is an auto next planned
-					if (currentPart.autoNext && t > currentPart.expectedDuration - AUTOTAKE_DEBOUNCE) {
+					if (currentPart.autoNext && Math.abs(currentPart.expectedDuration - playbackDuration) < AUTOTAKE_DEBOUNCE) {
 						return ClientAPI.responseError('Cannot take shortly before an autoTake')
 					}
 				}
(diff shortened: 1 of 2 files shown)


meteor/server/api/playout/playout.ts
```diff
@@ -212,17 +221,27 @@ export namespace ServerPlayoutAPI {
...
 						return ClientAPI.responseError('Cannot take during a transition')
 					}
 				}
+
+				const offset = currentPart.getLastPlayOffset()
+				if (start && offset && currentPart.expectedDuration) {
+					const t = Date.now() - start! + offset!
+
+					// If there is an auto next planned
+					if (currentPart.autoNext && t > currentPart.expectedDuration - AUTOTAKE_DEBOUNCE) {
```
nytamin: I would change this to
`if (currentPart.autoNext && Math.abs(t - currentPart.expectedDuration) > AUTOTAKE_DEBOUNCE)`
to ensure that you _can_ do a Take even after an auto-take, in the (however unlikely) case the auto-take didn't fire

meteor/server/api/playout/playout.ts
```diff
@@ -212,17 +221,27 @@ export namespace ServerPlayoutAPI {
...
-					const start = currentPart.getLastStartedPlayback()
+				if (allowTransition && currentPart.transitionDuration) {
 					if (start && now < start + currentPart.transitionDuration) {
 						return ClientAPI.responseError('Cannot take during a transition')
 					}
 				}
+
+				const offset = currentPart.getLastPlayOffset()
+				if (start && offset && currentPart.expectedDuration) {
+					const t = Date.now() - start! + offset!
```
nytamin: Shouldn't this be `const t = Date.now() - (start + offset)`? Perhaps also rename `t` to something more descriptive, like `actualDuration` or the like?

baltedewit: `Date.now()` and `start` are timestamps I believe. Offset is a duration. so `Date.now() - start` gives a duration, if you add the offset you get where in the part we are.

PR #164, opened 2020-01-27
fix: strip pre-release tags from version number when doing comparisons.

nytamin:
This fixes a bug reported by @gundelsby:
> Using nightly builds results in the following error message:
check blueprintCompability_system: Integration version: Version mismatch: blueprint.integrationVersion version: "1.7.0-nightly-20191216-161547-344775e.0" does not satisfy expected version range of core.tv-automation-sofie-blueprints-integration: ">=1.7.0 <2.0.0" (Blueprint has to be updated)

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/coreSystem.ts
+++ b/meteor/server/coreSystem.ts
@@ -177,7 +177,11 @@ function checkDatabaseVersion (
 	theyName: string
 ): { statusCode: StatusCode, messages: string[] } {
 
-	if (currentVersion) currentVersion = semver.clean(currentVersion)
+	if (currentVersion) {
+		// Strip any patches from the version:
+		const strippedVersion = semver.coerce(currentVersion)
+		if (strippedVersion) currentVersion = semver.valid(strippedVersion)
+	}
 
 	if (expectVersion) {
 		if (currentVersion) {


Julusian:
@nytamin I think we should discuss this. I am tempted to call the existing behaviour correct, even though it is a bit unfriendly.

gundelsby:
> 
> 
> @nytamin I think we should discuss this. I am tempted to call the existing behaviour correct, even though it is a bit unfriendly.

IMO, 1.7.0-nightlysomehash satisfies the requirement, which is 1.7.0 or higher, but lower than 2.0.0. Rejecting a nightly that is within the accepted range of version doesn't seem correct to me. But if it is, how can we test nightly builds of blueprints?

Julusian:
A 1.7.0-nightly is in the commit history before a 1.7.0 release. Once 1.7.0 has been tagged in master, future prereleases will be 1.7.1-nightly. So a 1.7.0-nightly is <1.7.0

As an example here, say we have compared to the previous 1.6.x
1.7.0-a which adds feature A
1.7.0-b which adds feature B
1.7.0 which combines feature A and B

With core is saying it needs >=1.7.0, then both nightlies are wrong as they are missing a feature that core expects. In theory this shouldnt be a problem (because semver) but this could also happen with a nightly in the middle of feature A where semver doesnt guarantee things to not change.

Additionally, if core is able to say it needs `^1.7.0`, from memory I think that must mean that 1.7.0 has been actually released and is being used by core, and so the blueprints should also be using that release

Finally, this is also just a warning. Core doesnt do anything differently based on this check, and I would expect to see this a lot in testing, but if it shows in a DK then something is using a wrong version.
I am now wondering if this check is even being done in the correct direction. Should it not be core that dictates an exact version of what it is using, and the blueprints state the range they support? The blueprints are the 'plugin' in our scenario, so they should be the ones that are trying to be more tolerant of version

gundelsby:
> 
> 
> A 1.7.0-nightly is in the commit history before a 1.7.0 release. Once 1.7.0 has been tagged in master, future prereleases will be 1.7.1-nightly. So a 1.7.0-nightly is <1.7.0
> 

Of course 🤦‍♂ 
I yield.

> Finally, this is also just a warning. Core doesnt do anything differently based on this check, and I would expect to see this a lot in testing, but if it shows in a DK then something is using a wrong version.

I didn't know, but in that case my original issue is non-existant. As long as it will still use the blueprints I'm happy, as my case indeed was testing.

nytamin:
I'll close this issue then :)

PR #165, opened 2020-01-28
Changes to core as media scanner is replaced and retired by media manager

sparkpunkd:

A feature to remove and retire media scanner.

Configuration settings for media scanner are via text file and command-line parameters.

Configuration settings for metadata and artifact generation are exposed in the Sofie settings GUI for media manager and stored in the Peripheral Devices collection. 

* **Other information**:

See related PR https://github.com/nrkno/tv-automation-media-management/pull/17 

**Status**


Julusian:
The placeholder property in the manifest needs some thought. For any playout sub-devices, there is now a placeholder and defaultVal property. These should probably be combined to be a single property

sparkpunkd:
The difference between placeholder vs defaultVal is subtle. Media scanner already had a default value mechanism that replaces an falesy value with a default. The placeholder gives a hint to the user as to what this would be. This is different from a default value being filled in explicitly. HTML makes this distinction - so I think we should to.

Julusian:
> HTML makes this distinction - so I think we should to.

Makes sense. We might want to look over the other gateways to check they use the correct ones at the correct points then.

PR #167, opened 2020-02-04
Optimization: Lookahead max search distance

nytamin:
This PR adds a setting for "lookahead maxSearchDistance".
This limits how far ahead the lookahead will look for content. Improving performance in long running shows.


jesperstarkar:
This is not primarily a performance fix, it is a feature to get a new lookahead behaviour like "as late as possible". It will improve logic for cuing cameras and lights, can drive a CUE-bus on Sisyfos and also used for clip loading logic.

PR #168, opened 2020-02-04
Various Bug fixes

nytamin:
This PR fixes various bugs:

* Second infinite on a layer within a part doesnt get extended
* Copy expectedDuration to inserted adlibs
* Don't throw error using github branches as packages
* Typings
* Bugfix: don't die if the previous part doesn't exist anymore
* Safeguard for first part(s) being invalid on reset and activate user action
* Proper handling of setting the first valid part on activation
* Use adlibPrerollDuration when converting adlib to a part
* Exception when converting adlib to part if preroll not defined
* Guard warning
* Infinite pieces started in an autonext part would disappear briefly between the take and the new timeline resolving
* Parsing versions
* Allow Next:ing past an empty segment.
* Bugfix: not able to reset rundown when first part was playing
* Remove leftover properties unlimited & onPGMClean
* Block alt key default action
* Change UI settings to a Meteor-standard
* Removes un-used properties from the source layers to clean up the UI and avoid confusions this introduce


Julusian:
41b696ec64d4d16b0567592284d1c9db244e63ae and b033b056f643a910904550235e5b97fed3535196 from this PR have been rewritten and pulled into #193. Hopefully this wont conflict or cause the original commits to be merged, as that will cause some work to be done twice

PR #172, opened 2020-02-04
Optimization when doing many Adlibs 

nytamin:
This PR adds an optimization for the situation where many adlibs are played while staying in one part.
This is fixed by adding a filter in Core that removes old-and-stopped-adlibs after a while, making calculations and timeline to run faster.
System slows down to a crawl after 15-30 adlibs (we experienced this when cutting between cameras)
* **What is the new behavior?**
Adds the property "definitelyEnded" to pieces to allow for efficient trimming of pieces that no longer need to be evaluated to calculate the current state, increasing performance.


Julusian:
@nytamin I think this PR should also contain https://github.com/olzzon/tv-automation-server-core/commit/786d478ac96a49cba22673853430d45d2a0e2c14

jstarpl:
@nytamin Also, I think olzzon@9cc4abcb3ffff8ac787371d49a0c585552ac4598 should be included here as well.

PR #173, opened 2020-02-04
Feature: Dashboard panel improvements: timeline view, external frame

thomasslee97:
Feature
- Improves dashboard / shelf layouts to allow for "toggling" adlibs.
- Provides a "timeline" view whereby an indicator is given of which adlibs belong to a particular piont in a show.
- Adds an option of using an "External frame" panel, that is an iframe with an API to exchange events with Sofie

Timeline view:
image

Adlib toggle:
image
image

### DIFF
--- a/meteor/client/lib/invalidatingTime.ts
+++ b/meteor/client/lib/invalidatingTime.ts
@@ -1,0 +1,17 @@
+import { Tracker } from 'meteor/tracker'
+import { getCurrentTime } from '../../lib/lib'
+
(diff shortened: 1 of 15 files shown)


jesperstarkar:
@thomasslee97 your fixes from yesterday, should they be added here too?

jstarpl:
Code from this PR has been included in commit 61dd000. I think that if `feat/adlib-mos-plugin` is merged into `master` then we should just close this one.

PR #174, opened 2020-02-04
Hidden Segments

nytamin:
Support of a new feature: "Hidden Segments".
A Segment can now be set as "hidden" in the blueprints.
A Hidden segment will not be shown in the RundownView, but will be shown in the Shelf.

This is useful for example for adding Adlibs without nessesarily having them related to a certain Segment.

### DIFF
--- a/meteor/client/ui/RundownView.tsx
+++ b/meteor/client/ui/RundownView.tsx
@@ -1088,7 +1088,12 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
 	return {
 		rundownId: rundownId,
 		rundown: rundown,
-		segments: rundown ? Segments.find({ rundownId: rundown._id }, {
+		segments: rundown ? Segments.find({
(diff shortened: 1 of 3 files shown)


Julusian:
This looks to leave parts inside a hidden segment to be playable. Perhaps all parts inside a hidden segment should be forced to invalid by core?

Also, I suspect the prompter will still show these parts too?

jstarpl:
I think prompter is supposed to show them, if content for the prompter is
available in them.

wt., 11 lut 2020, 11:27 użytkownik Julian Waller <notifications@github.com>
napisał:

Julusian:
> I think prompter is supposed to show them, if content for the prompter is available in them.

But as an empty and invalid part shows the part name, that is going to lead to a lot of noise

jstarpl:
OK, after discussing it we agree that:
- hidden segments are hidden in the prompter as well
- core should force all parts in a hidden segment to be invalid

PR #181, opened 2020-02-04
feat: Timeline grab to scroll

thomasslee97:
Feature
Allows the user to click and drag on the segment timeline to scroll left and right through a segment.


jstarpl:
I also have a branch called [`feat/zoomSegLatch`](https://github.com/olzzon/tv-automation-server-core/tree/feat/zoomSegLatch) that is both #181 and #175, but makes the grab to scroll functionality hidden enable-able with a Settings flag.

PR #184, opened 2020-02-04
feat: Queued adlibs

thomasslee97:
Feature
Allows an adlib to indicate that it should always be queued when added to the timeline, rather than being immediately taken on air.
* **Other information**:
An example usecase is allowing the inserting of transitions as adlibs, where they may need to be created as their own part for timing calculations
to work.


Julusian:
This gives the same effect as the `Shift+1` hotkeys that used to be in use.

As nrk are no longer using the SourceLayer.isQueueable property, perhaps it should be removed? Its behaviour can be emulated by creating a second adlib with this property set, but with the added benefit of full control over the hotkey used.

PR #186, opened 2020-02-04
vizMSE device

nytamin:
Feature
Adds support for vizMSE devices to be connected to the Playout-gateway.
(vizMSE-devices are Viz-Engines controlled through Media-Sequencer)


Julusian:
The UI panels changed here are deprecated and are only shown when an outdated playout-gateway is connected (before https://github.com/nrkno/tv-automation-playout-gateway/pull/58).
This has been passed back to the contributors to rework it in the new way

PR #187, opened 2020-02-04
feat: option for sticky items to only make 'original' pieces sticky

thomasslee97:
Feature
All pieces on sticky layers are sticky.
Adds an optional setting to showstyle such that adlibs are not sticky.

### DIFF
--- a/meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
+++ b/meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
@@ -723,6 +723,18 @@ const SourceLayerSettings = translate()(class SourceLayerSettings extends React.
 									</div>
 									<div className='mod mvs mhs'>
 										<label className='field'>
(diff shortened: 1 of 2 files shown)


jstarpl:
I wonder if this should be an attribute of the source layer or rather an attribute of the (AdLib) Piece saying that it should be excluded out of Sticky processing. It seems that if the blueprints have some kind of implementation-specific functionality which they need to hide from Sticky, it should be on a per-piece basis.

Julusian:
This behaviour wouldn't work the same if done by the blueprints, as it adds
`adlibSourceId: { $exists: false }` to the query, meaning it will not pick
up either explicit adlibs or other recalls of stickies. I don't know if
that latter part is desired/intentional, but I expect that is how it will
be working.

However, if the actions replacement for adlibs task gets done in the way I
expect, sticky items would likely get moved into that and then that in
effect would be done as you suggest.

PR #188, opened 2020-02-04
feat: iNews gateway

thomasslee97:
Feature

Adds support for the iNews gateway: https://github.com/olzzon/tv2-inews-ftp-gateway.git


baltedewit:
Is it possible to write the config page for this device as a config manifest instead?

nytamin:
Definitely, we'll rewrite it to match the config manifest before finally merging it in.

PR #190, opened 2020-02-07
Fix: DeviceConfigManifest: creating default table rows

jstarpl:

This is a bugfix

When creating "default" rows in tables, a truthy check is done on the `defaultVal` for each field before creating. This causes a problem if the default value is an empty string, 0 or false.

It uses a `!== undefined` to verify that the `defaultVal` is defined.

* **Other information**:

**Status**


jstarpl:
I think all the people that could review this are away right now, so since this is an extremely simple fix, I'm just going to merge it in.

PR #195, opened 2020-02-14
feat: Buckets & MOS plugins in ExternalFrames in Dashboard

jstarpl:

Adds the Bucket concept and enables ExternalFrame guests to use MOS HTML5 Plugin API

* **Other information**:

**Status**


jstarpl:
> 
> 
> Looks good from initial glance (I've only looked at the server-side parts).
> 
> It needs to be rebased onto `release20` and functionality-tested again before merging, I think.

Rebased onto `release20` and should be ready for feature-testing along with Ola's updated MOS plugin.

jstarpl:
I just realized we should wait for testing before we merge, so that we don't cause any more mess in the workflow. This should now be ready for merging.

PR #197, opened 2020-02-18
fix: dont call uploadFileToAtem as part of deviceMakeReady

Julusian:
Partner: https://github.com/nrkno/tv-automation-playout-gateway/pull/60

Remove the calls of uploadFileToAtem to playout-gateway.
Playout-gateway should internally handle ensuring the media loaded to the device is correct.

This code will trigger an upload of media based on old and incorrect data, overriding correct media with outdated ones.

The media upload process is left to playout-gateway

* **Other information**:

**Status**


Julusian:
Well, makeReady isnt handling this. But playout-gateway was already rechecking when device status changed. So it is being left completely up to that

PR #198, opened 2020-02-19
feat: Gap parts

jstarpl:

This implements Gap parts: parts that fill space within a show that is to be filled by other parts during playback.

We have invalid parts, but the assumption is that they contain some actual content.

On top of being `invalid`, the part can be shown as `gap`, which will enable a specific style in the GUI. Apart from that change, all other functionality should be the same. To enable the complete `gap` functionality, the `Part` needs to be a member of a `displayDurationGroup`.

* **Other information**:

**Status**


meteor/lib/collections/Parts.ts
```diff
@@ -100,6 +100,7 @@ export class Part implements DBPart {
         color?: string
     }
 	public floated?: boolean
+	public gap?: boolean
```
nytamin: Just a note; this PR is dependent on an addition of the `IBlueprintPart.gap`-property to the blueprint-integration repo as well

jstarpl: Yes, and now the blueprint-integration has been updated.

PR #199, opened 2020-02-24
feat: placeholder option for config manifests

baltedewit:


sparkpunkd:
Thanks for this pull request. This will be a welcome UX improvement for users editing several settings on a page.

I note that not all types of attribute in the `EditAttribute` component support placeholders. I would suggest that at least Int and Float also need `placeholder={this.props.label}` added in their `render()` method?

PR #205, opened 2020-03-25
fix: Part Instances contentMetaData broken

jstarpl:

This PR resolves an issue with `contentMetaData` not being attached to the correct object within `SourceLayerItemContainer`

Hoverscrub box doesn't show up, even if it should (mediaPreviewsUrl is set up and the MediaObject is found)

The `contentMetaData` is now placed in the correct position, where `VTRenderer` and other Piece renderers can pick it up.

* **Other information**:

**Status**


jstarpl:
> Looks fine.
> But why did this not trigger some typings errors? That concerns me

I think the problem was mainly with `overrides` being untyped and then also with `pieceCopy` being [untyped as well](https://github.com/nrkno/tv-automation-server-core/blob/4bf3e14f4a5e0d5b3382ebd113575a29bcb25491/meteor/client/ui/SegmentTimeline/SourceLayerItemContainer.tsx#L152). I've added stronger typing there, should now trigger errors, if anything is wrong.

PR #207, opened 2020-03-27
chore: refactor connection status notification strings

jstarpl:

Make connection status text generation more organized and easier to localize

There is some code-based concatenation of strings in `getStatusText` that could make translations impossible into some languages (non-VSO or -SVO word-order).

The Sofie Automation Server string is translated and then used as an input to the connection status strings, allowing full localization.

* **Other information**:

**Status**


jstarpl:
I'll make the typing stronger for the `status`, but changing that to an enum should be done at the `@types/meteor/ddp` library, otherwise we risk going out-of-sync between those two typing sources. I'm just going to leave that as-is for the time being.

PR #214, opened 2020-04-14
fix: make the userActions.take method be a syncFunction on a rundownPlaylist

jstarpl:

This PR fixes a bug with the `MINIMUM_TAKE_SPAN` protection.

There is a short window of opportunity where a slow response from the DB at the right moment, when switching between two non-autonext parts, can cause a double-take to happen.

The userActions take method is now a `syncFunction`, so that only a single `take` userAction can be executed at any time for a rundownPlaylist. This should ensure that the already existing `MINIMUM_TAKE_SPAN` check can work correctly.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/userActions.ts
+++ b/meteor/server/api/userActions.ts
@@ -645,7 +645,7 @@ class ServerUserActionAPI implements NewUserActionAPI {
 }
 registerClassToMeteorMethods(UserActionAPIMethods, ServerUserActionAPI, false, (methodContext: MethodContext, methodName: string, args: any[], fcn: Function) => {
 	const eventContext = args[0]
-	return ServerClientAPI.runInUserLog(methodContext, eventContext, methodName, args, () => {
+	return ServerClientAPI.runInUserLog(methodContext, eventContext, methodName, args.slice(1), () => {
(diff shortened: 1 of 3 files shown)


meteor/server/api/userActions.ts
```diff
@@ -54,9 +55,8 @@ export function setMinimumTakeSpan (span: number) {
 */
 
 // TODO - these use the rundownSyncFunction earlier, to ensure there arent differences when we get to the syncFunction?
-export function take (rundownPlaylistId: RundownPlaylistId): ClientAPI.ClientResponse<void> {
+export const take = syncFunction((rundownPlaylistId: RundownPlaylistId): ClientAPI.ClientResponse<void> => {
```
nytamin, line 58: A minor note;
This will create an anonymous function inside the syncFunction. By naming the function it'll be slightly easier to read call-stacks when troubleshooting:

`export const take = syncFunction(function take(rundownPlaylistId: RundownPlaylistId): ClientAPI.ClientResponse<void> {`

jstarpl: Good tip!

PR #216, opened 2020-04-16
fix: part notes and invalidReason notes not showing

Julusian:
Two parts to this:

1) Any notes generated by the blueprints for a part were not being saved against that part.

2) Any part notes were not being shown in the ui. Neither were any invalidReason notes.

**Status**


Julusian:
A related thought, when the notes are stored in the db they contain the rundownId, segmentId and partId. This feels excessive and a little dangerous, as we are not updating these when moving parts, so while the blueprints should be regenerating them, there is some risk the ids could fall out of sync and the notes will get lost by the ui.

@jstarpl @nytamin any objection to removing those ids from the notes and to add them in the ui if/where they are needed?

PR #217, opened 2020-04-20
Server routes and various fixes

nytamin:
**Refactoring:**
Collect server-side route setup into a single file, The middlewares should only be defined once. By putting them into a single file, we avoid the risk of defining several middlewares (overwriting eachother)
**Fixes**
* Increase upload limit, to allow for large rundown-playlists, etc
* when restoring from snapshots: set correct playlistId on rundowns
* when removing rundownPlaylist, actually remove the rundownPlaylist 

**Status**


meteor/server/api/rundownLayouts.ts
```diff
@@ -40,12 +37,11 @@ export function removeRundownLayout (layoutId: RundownLayoutId) {
 	RundownLayouts.remove(layoutId)
 }
 
-const postJsRoute = Picker.filter((req, res) => req.method === 'POST')
-postJsRoute.middleware(bodyParser.text({
+PickerPOST.middleware(bodyParser.text({
```
Julusian, line 40: should this middleware still exist here?

nytamin: good find! I fix

PR #222, opened 2020-04-29
[TV 2 Contributions] More contributions following next2020Q1

thomasslee97:
- Config Table Improvements
  - Sticky headers + better scrolling
    - Column headers remain visible as the user scrolls through the table, making editing easier.
  - Sorting tables by columns
    - Clicking on the column headers allows the user to sort tables in ascending and descending order.
  - Table column rank
    - Allows the blueprints to change the order of columns in a table by setting a rank; columns are ordered left-to-right according to ascending rank (i.e. 0 on left, 10 on right).
- Numeric field typing improvements
  - Improves the input validation of numeric fields to prevent the cursor from "jumping" while the user is typing.
- New dashboard panel type: list
  - Adds a new dashboard panel type that presents the panel buttons in a more compact list format.
image
- Queue adlibs with transitions
  - Allows adlibs to be given the properties `adlibPreroll`, `adlibAutoNext`, `adlibAutoNextOverlap`, `adlibDisableOutTransition`, `expectedDuration` which will be copied into the part properties when the adlib is queued. This allows an adlib to create a transition part, with transitions both into and out of the part.
- New adlib panel type: Multiview
  - This panel type allows an interactive multiview to be presented to the user. Multiview panels can have adlibs associated with them so that when the user clicks/taps on the multiview panel, they can trigger an action such as cutting to the source associated with that panel. When used in conjunction with the external frame panel and the image provider, live previews of the sources can be displayed within the multiview panels.
  - The panels also display Live / Preview highlights around the borders of each panel, following the same logic as dashboard buttons to determine if a source is on-air / set as next.
  - The image below shows a multiview made from 10 of these panels, where the "Preview" panel is assigned the "Take" action when clicked. Although this is a standard multiview configuration, panels can be placed anywhere in the dashboard following the rules of other panels for positioning.
image
- Dashboard panel In / Out buttons
  - Option to add "In" and "Out" buttons to a list type dashboard panel, configured via panel settings. This allows the user to select, for example, a graphic that they want to send on-air soon, then press "In" to activate the selected adlib. Pressing "Out" will clear the source layer of the taken adlib. This is intended for custom user interfaces, but could also be useful for a workflow where a user can select the next graphics / VT / other sources that they want during a break in activity, then rapidly take them / remove them during a busy segment - without needing to search for the adlibs they want when they need them immediately.
image
- DVE thumbnails in dashboard panel lists
  - List dashboard panels have the option to show thumbnail-style previews of splits, using the same previews as are displayed when hovering over a DVE piece in the rundown view. This allows a particular layout to be found at a glance.
  - Configured using the `Show DVE layouts as thumbnails` setting, which is available for List type dashboard panels.
image
- View shelf of inactive rundowns
  - This allows the user to open a rundown and add `/shelf` to the URL to view the shelf associated with this rundown as a standalone page. This is for applications where the shelf is being used as the primary user interface and needs to assume responsibility for activating / deactivating.
- Dropdown blueprint config manifest option, including source layer and layer mapping dropdowns.
  - Allows the blueprints to expose dropdowns to the settings page, both single-select and multi-select dropdowns.
  - Special "Source Layers" and "Layer Mappings" dropdowns can be exposed as blueprint settings, which are populated by core from the available source layers and layer mappings. These can also be exposed as either single-select or multi-select. This allows the user to customize things such as which audio layer mappings to assign to a camera source, or which source layer to use when creating a graphic of a particular type.
image
- ~~Add rundown identifier~~
  - ~~Adds an "identifier" field to segments, such that segments can have some external label placed in their header; e.g. Their story rank / ID in an NRCS.~~
  - ~~This currently requires custom css to be added to opt-in, the default display style is set to `display: none`, so systems that do not need the identifier will not see it.~~
  - ~~Set by blueprints as an optional property of the segment returned from `getSegment`~~
image
- Unsynced Segments
  - Adds a setting `allowUnsyncedSegments` to allow individual segments to become unsynced - disabled by default.
  - If a part within a live segment is modified, the live segment will become unsynced, but updates to other segments will still come through.
  - Upon leaving the unsynced segment, it will be resynced by resyncing against the attached ingest gateway (e.g. MOS, iNews).
  - Adds a resyncSegment method for ingest gateways.
    - For iNews gateway, this reloads an individual segment.
    - For MOS gateway, this falls back to reloading the entire rundown.
- Misc fixes
  - Scaling of external frame panels
    - External frame panels can be scaled according to some factor set in the settings, to allow a better fit of external content.
  - Add display rank to external frame panel
    - Allows the order in which external frame panels are displayed to be modified, in the same way as other panels.
  - Hide pieces in gap
    - Infinites that are copied into "gap" pieces are hidden from view using css. This declutters the gap pieces and ensures that they do not look like pieces that can be taken by the user.
  - Compress Gap
    - Gap parts will "compress" as time runs out in the segment, to give a visual indicator of the amount of adlib time left to the user. The gap will compress down to 1px wide, with a width of 1px indicating time has run out, and it is time to move on.
  - Rundown layout upload / subscribe to collection
    - Fixed a bug where the shelf layouts of only one show style would be subscribed to, so viewing the setting pages of other show styles would not show their shelf layouts.
    - Removed the requirement for uploaded shelf layouts to have a show style Id, they will instead take the Id of the show style they are uploaded to. This allows a shelf layout to be reused across multiple show styles by just uploading the file to each show style.
  - Filtering for only global adlibs
    - Fixed a bug where if any dashboard panel had the "Include Global AdLibs" setting set to "only", all dashboard panels would only see global adlibs.
  - Always crop infinites when adlibbing
    - Fixed a bug where future infinites would not be updated by queued adlibs.
  - Remove definitely ended infinites
    - Use the `definitelyEnded ` piece property to aid in cropping infinites.
  - ~~Fixed issue where start times of timeline objects could be offset from the start of a piece by the amount that the piece was offset from the start of the part; where the piece had been dynamically inserted by an adlib.~~

* **Other information**:
Requires [https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/60](https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/60)

**Status**


meteor/lib/collections/RundownLayouts.ts
```diff
@@ -63,6 +79,7 @@ export interface RundownLayoutFilterBase extends RundownLayoutElementBase {
 	label: string[] | undefined
 	tags: string[] | undefined
 	displayStyle: PieceDisplayStyle
+	dvePreviewsAsThumbnails: boolean
```
jstarpl, line 82: This seems very specific. Wouldn't it just make sense to have a flag to just enable/disable thumbnails for an entire filter? Or a list of SourceLayerTypes/SourceLayers to enable/disable thumbnails on?

meteor/server/api/playout/adlib.ts
```diff
@@ -192,11 +192,51 @@ export namespace ServerPlayoutAdLibAPI {
...
+			}
+
+			if (adLibPiece.expectedDuration !== undefined) {
+				PartInstances.update(partInstance._id, {
+					$set: {
+						expectedDuration: adLibPiece.expectedDuration
+					}
+				})
+			}
+
```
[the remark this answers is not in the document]
thomasslee97: https://github.com/nrkno/tv-automation-server-core/pull/222/commits/f89ee24f10deae5c361066cb07e254845a3564e4

meteor/lib/collections/RundownLayouts.ts
thomasslee97: Good point, I think I'll go for an entire filter for now - as that matches the use case this is being built for at TV 2.

meteor/server/api/playout/adlib.ts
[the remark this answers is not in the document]
thomasslee97: I would think so, as the adlibs are creating their own part they are starting from a "blank template" to begin with.

meteor/server/api/ingest/rundownInput.ts
```diff
@@ -417,24 +428,83 @@ function updateRundownFromIngestData (
...
-	if (!isUpdateAllowed(dbPlaylist, dbRundown, { changed: [{ doc: dbRundown, oldId: dbRundown._id }] }, prepareSaveSegments, prepareSaveParts)) {
-		ServerRundownAPI.unsyncRundown(dbRundown._id)
-		return false
+	if (Settings.allowUnsyncedSegments) {
+
+		if (!isUpdateAllowed(dbPlaylist, dbRundown, { changed: [{ doc: dbRundown, oldId: dbRundown._id }] })) {
+			ServerRundownAPI.unsyncRundown(dbRundown._id)
+			return false
+		} else {
+			const segmentChanges: SegmentChanges[] = splitIntoSegments(
```
Julusian, line 451: This feels to have to split the data back into segments. I suppose it has to be done though, unless we do the prepare step again.

Has this been tested for any cross-bleed issues?
As in, moving piece99 from segment1 to segment2. I assume this should be safe due to how we do ids, but that would be good to confirm.

meteor/lib/Settings.ts
```diff
@@ -41,7 +43,8 @@ const DEFAULT_SETTINGS: ISettings = {
 	autoRewindLeavingSegment: false,
 	disableBlurBorder: false,
 	defaultTimeScale: 1,
-	allowGrabbingTimeline: true
+	allowGrabbingTimeline: true,
+	allowUnsyncedSegments: true
```
Julusian, line 47: I suspect that with mos resyncing segments will get 'stuck' in a loop, so it might be better to disable by default at least until there is a fix for that.

This is because mos has to resync the whole rundown, so while resyncing a segment, it will likely unsync the newly playing segment.

meteor/lib/collections/RundownLayouts.ts
```diff
@@ -46,6 +47,21 @@ export interface RundownLayoutElementBase {
...
+
+export enum RundownLayoutMultiViewRole {
+	QUEUE = 'queue',
+	TAKE = 'take',
+	PROGRAM = 'program'
+}
+
+export interface RundownLayoutMultiView extends RundownLayoutElementBase {
+	type: RundownLayoutElementType.MULTIVIEW
+	windowNumber: number
```
Julusian, line 61: This property doesnt appear to be read anywhere, its only use is the ui input field.

Should this element be called a multiview? It doesnt look specific to that particularly, and looks to be a more generic name overlay with an action.

meteor/client/ui/Shelf/MultiViewPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+				break
+			case RundownLayoutMultiViewRole.PROGRAM:
+				break
+		}
+	}
+
+	render () {
+		const isTake = this.props.panel.role === RundownLayoutMultiViewRole.TAKE
+		const isProgram = this.props.panel.role === RundownLayoutMultiViewRole.PROGRAM
+		const isLarge = isProgram || isTake
```
Julusian, line 123: I don't like this isLarge behaviour being done based on the action. This should be made into a separate config field

Julusian:
@jstarpl could you have a read through the ui changes to make sure you are happy with it?

@thomasslee97 I've added some comments, could you look over them and let me know what you think

Julusian:
@thomasslee97 I have also reverted a couple of changes that I believe to be handled by other changes being made by nrk

meteor/server/api/ingest/rundownInput.ts
thomasslee97: I have not managed to test that use case, as the iNews gateway does not have that awareness of parts / pieces being moved / copied (it's not how iNews works as such). I can say that it handles parts within a segment changing order with the iNews workflow. I suspect an ENPS system will be needed to test every permutation.

meteor/lib/Settings.ts
thomasslee97: Addressed by https://github.com/olzzon/tv-automation-server-core/commit/bbf6d34e5fa148c2bad34d3aa22ec79079aa25bf

meteor/lib/collections/RundownLayouts.ts
thomasslee97: I've removed windowNumber (https://github.com/olzzon/tv-automation-server-core/commit/6ff7353b14fec1f87f84d11d0a347a480c5b609e), it was a leftover from the initial implementation.

Yeah, it's not specific to that as such. I don't have an alternate name myself, @jstarpl any suggestions?

meteor/client/ui/Shelf/MultiViewPanel.tsx
thomasslee97: I have moved this to the setting "Place label below panel" tied to the "labelBelowPanel" property https://github.com/olzzon/tv-automation-server-core/commit/8ac362f4db862137d5be13c0e9f89270b442221f

meteor/client/ui/RundownView/RundownTiming.tsx
```diff
@@ -332,7 +332,7 @@ withTracker<IRundownTimingProviderProps, IRundownTimingProviderState, IRundownTi
 					this.displayDurationGroups[partInstance.part.displayDurationGroup] =
 						(this.displayDurationGroups[partInstance.part.displayDurationGroup] || 0) + (partInstance.part.expectedDuration || 0)
 					displayDurationFromGroup = partInstance.part.displayDuration
-						|| Math.max(0, this.displayDurationGroups[partInstance.part.displayDurationGroup], this.props.defaultDuration || DEFAULT_DURATION)
+						|| Math.max(0, this.displayDurationGroups[partInstance.part.displayDurationGroup], partInstance.part.gap ? 1 : this.props.defaultDuration || DEFAULT_DURATION)
```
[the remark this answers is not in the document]
thomasslee97: This allows the gap to fold down to being essentially invisible (a value of 0 did not play nicely at all). So, the user can see their time running out entirely - without this check it would only compress down to the default duration and no further.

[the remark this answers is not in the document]
jstarpl: Aha, OK, that makes sense. OK, in that case I'm fine with getting this to a constant like `MINIMAL_NONZERO_DURATION` and perhaps a comment on why this is the way it is.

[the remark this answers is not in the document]
thomasslee97: Addressed by https://github.com/olzzon/tv-automation-server-core/commit/f0efaa08f1009f538b409ccccebb64beaf8894a2

meteor/lib/collections/RundownLayouts.ts
Julusian: We discussed 'adlib region'. Could you change this too?

meteor/client/ui/SegmentTimeline/SegmentTimeline.tsx
```diff
@@ -728,6 +728,9 @@ export class SegmentTimelineClass extends React.Component<Translated<IProps>, IS
 							hideOnZero={true}
 						/>
 					}
+					{
+						Settings.allowUnsyncedSegments && this.props.segment.unsynced && t('UNSYNCED')
```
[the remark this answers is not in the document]
thomasslee97: Added here https://github.com/olzzon/tv-automation-server-core/commit/bc77c8dcccc772f537c8ed4dd8648c6405162cff

meteor/client/ui/Shelf/DashboardActionButton.tsx
```diff
@@ -73,12 +76,8 @@ export class DashboardActionButton extends React.Component<IDashboardButtonProps
...
 						onMouseUp={(e) => this.props.onButtonUp(button, e)}
 						style={{
-							width: button.width ?
-								(button.width / 3 * DEFAULT_BUTTON_WIDTH) + 'em' :
-								undefined,
-							height: button.width ?
-								(button.height / 3 * DEFAULT_BUTTON_HEIGHT) + 'em' :
-								undefined
+							width: 'calc(100% - 8px)',
+							height: 'calc(100% - 8px)'
```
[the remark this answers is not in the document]
thomasslee97: Moved to CSS https://github.com/olzzon/tv-automation-server-core/commit/7f9a463f8a3e12bce7eb396b645f5b9b060b54d6

meteor/lib/collections/RundownLayouts.ts
thomasslee97: I've renamed them to AdLib Regions https://github.com/olzzon/tv-automation-server-core/commit/01e0301dc0e2d21ff48d6516a4d11a752532d0a8

meteor/client/ui/Shelf/DashboardPanel.tsx
```diff
@@ -527,16 +605,17 @@ export function getUnfinishedPieceInstancesReactive (currentPartInstanceId: Part
 	}
 
 	// Convert to array of ids as that is all that is needed
-	const unfinishedPieceInstanceIds: { [adlibId: string]: PieceInstanceId[] } = {}
-	_.each(_.groupBy(prospectivePieces, (piece) => piece.piece.adLibSourceId), (grp, id) => unfinishedPieceInstanceIds[id] = _.map(grp, instance => instance._id))
+	const unfinishedPieceInstanceIds: { [adlibId: string]: PieceInstance[] } = {}
+	_.each(_.groupBy(prospectivePieces, (piece) => piece.piece.adLibSourceId), (grp, id) => unfinishedPieceInstanceIds[id] = _.map(grp, instance => instance))
```
Julusian, line 609: @thomasslee97 whats the reason for this change? I cant find anywhere that doing anything with the contents, only some checks on the length of the array values

PR #224, opened 2020-04-30
fix: some pieces starting at non-zero time being forced to zero-time

Julusian:
This fixes a bug introduced in the TV2 contributions. They have fixed this in a different way in their fork, but the changes did not make it into the original batch of PRs.

This was causing some adlib pieces to think they start at 0 within the part, which causes the timings in them to be off. 

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/playout/pieces.ts
+++ b/meteor/server/api/playout/pieces.ts
@@ -21,7 +21,7 @@ import {
 	OnGenerateTimelineObj,
 	TSR
 } from 'tv-automation-sofie-blueprints-integration'
-import { transformTimeline } from '../../../lib/timeline'
+import { transformTimeline, TimelineContentObject } from '../../../lib/timeline'
 import { AdLibPiece } from '../../../lib/collections/AdLibPieces'
 import { Random } from 'meteor/random'
 import { prefixAllObjectIds } from './lib'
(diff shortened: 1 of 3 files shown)


meteor/server/api/playout/pieces.ts
```diff
@@ -296,8 +296,12 @@ export function getResolvedPiecesFromFullTimeline (playoutData: RundownPlaylistP
 	const pieceInststanceMap = normalizeArray(pieceInstances, '_id')
 
 	objs.forEach(o => {
-		if (o.enable.start === 'now') {
-			o.enable.start = now
+		if (o.enable.start === 'now' && o.isGroup) {
+			if ((o as any).isPartGroup) {
+				o.enable.start = now
+			} else if (o.metaData && o.metaData.pieceId) {
+				o.enable.start = 0
```
nytamin, line 303: As an uninformed reader, my first question would be "but what about if it's not a group that has `o.enable.start === 'now'`"?
Could we refactor the statements to be a little more clear, and perhaps add a comment or two on what situations we solve?

Julusian: My first reaction is that we don't really support doing that. We don't block it, but it gets really weird results. (eg, during an autonext the now gets set to a time 2 mins before the part it is in is on air) But also a few lines above we filter out anything which is not a group so thats not an issue here.

But you have a point. and I think that always setting a piece to 0 will cause problems, so I am rewriting it to handle this all better

PR #220, opened 2020-04-29
feat: Performance improvements on long shows

nytamin:
(This is based on the branch release20, to be merged first)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
This PR contains heavy performance improvements, both server-side and client-side.
The changes are focused on long-running shows with a large number of elements in them.

TODO:
- [x] Implement Database cache on server
- [x] Implement better granularity of reactivity on client
- [x] Have another discussion about obvious improvements
- [ ] Test for regressions

**Status**

- [ ] The functionality has been tested by the PR author
- [ ] The functionality has been tested by NRK

PR #227, opened 2020-05-14
Fix: wrong order of parts

nytamin:
This is a bug fix for the bug we've seen with part being in the wrong order.

It adds the propertys ingestData.modified and Segment.externalModified, to be used for better determining of what data to be updated.

This is a fix for an issue where a data update failed, and unsequent data updates didn't come through because the cached-ingestData and incoming-ingestData where in sync even though the Segments/Parts where not.

Note that this PR is based upon #220 

* **Other information**:

**Status**


meteor/server/api/ingest/ingestCache.ts
```diff
@@ -9,22 +9,29 @@ import { RundownId } from '../../../lib/collections/Rundowns'
...
 	_.each(segmentMap, objs => {
 		const segmentEntry = objs.find(e => e.type === IngestCacheType.SEGMENT)
 		if (segmentEntry) {
-			const ingestSegment = segmentEntry.data as IngestSegment
-			_.each(objs, e => {
-				if (e.type === IngestCacheType.PART) {
-					ingestSegment.parts.push(e.data as IngestSegment)
+
+			const ingestSegment = segmentEntry.data as LocalIngestSegment
+			ingestSegment.modified = segmentEntry.modified
```
Julusian, line 27: Shouldnt the various parts in this file use the makeNewIngest* functions instead of doing it manually?

Julusian:
Just the one query, other than that the 'wrong order of parts' commits look good (I havent read the base perf commits yet)

meteor/server/api/ingest/ingestCache.ts
nytamin: There is a difference:
The makeNewIngest**-functions (with emphasis on "New") always updates the `modified` property. In this case, I want to use it from the db to avoid unnecessary data-updates down the line.

PR #237, opened 2020-05-27
Accounts and security

nytamin:
_**Note: The description of this PR is to be updated/written**_

Note: This PR is based on #220 .

This is a fairly large PR that adds support for user accounts and a basic foundation for security policies.

Disclaimer: it is not 100% complete and will be complemented with more security fixes in later PR:s.

* **Other information**:

**Status**


meteor/client/ui/App.tsx
```diff
@@ -134,23 +176,38 @@ class App extends React.Component<InjectedI18nProps, IAppState> {
...
-							<Route path='/rundowns' component={RundownList} />
-							<Route path='/rundown/:playlistId' component={RundownView} />
-							<Route path='/activeRundown/:studioId' component={ActiveRundownView} />
-							<Route path='/prompter/:studioId' component={PrompterView} />
-							<Route path='/countdowns/:studioId/presenter' component={ClockView} />
-							<Route path='/status' component={Status} />
-							<Route path='/settings' component={Settings} />
+							{Settings.enableUserAccounts ?
+								[
+									<Route key='0' exact path='/' component={(props) => <LoginPage {...props}/>} />,
```
jstarpl, line 193: I would suggest not using arrays, if you are not actually providing an array of items. Use a <React.Fragment> or the shorthand `<></>` as that will perform way better and you don't have to set any dummy `key` attributes.

meteor/client/ui/AccountPage.tsx
```diff
@@ -0,0 +1,73 @@
...
+				undefined,
+				NoticeLevel.CRITICAL,
+				'Error deleting account',
+				'Account Page'
+			))
+		})
+	}
+
+	componentDidUpdate () {
+		if (this.state.user === undefined && getUser() !== undefined) {
```
jstarpl, line 47: Shouldn't this just be moved to the tracker part, since the getUser() is reactive?

meteor/server/api/playout/adlib.ts
```diff
@@ -22,28 +22,29 @@ import { rundownPlaylistSyncFunction, RundownSyncFunctionPriority } from '../ing
 
 import { PieceInstances, PieceInstance, PieceInstanceId } from '../../../lib/collections/PieceInstances'
 import { PartInstances, PartInstance, PartInstanceId } from '../../../lib/collections/PartInstances'
+import { initCacheForRundownPlaylist, CacheForRundownPlaylist } from '../../DatabaseCaches'
 
 export namespace ServerPlayoutAdLibAPI {
-	export function pieceTakeNow (rundownPlaylistId: RundownPlaylistId, partInstanceId: PartInstanceId, pieceInstanceIdOrPieceIdToCopy: PieceInstanceId | PieceId) {
-		return rundownPlaylistSyncFunction(rundownPlaylistId, RundownSyncFunctionPriority.USER_PLAYOUT, () => {
-			const rundownPlaylist = RundownPlaylists.findOne(rundownPlaylistId)
```
Julusian, line 29: Not doing a full review, just saw this while dealing with a merge conflict

I don't like the rundownPlaylist crossing the syncFunction boundary like this. It could cause all kinds of weird race conditions, as any other rundownPlaylistSyncFunction which are already executing or queued could change any of the properties set on the playlist.
Once inside the rundownPlaylistSyncFunction, it has the lock to the RundownPlaylist so can reload and be guaranteed that nothing else will change it (in any way that matters)

example timeline:
0.0s User does take to part 'b' (currently in 'a')
0.1s User runs adlib in part 'a'
0.2s Core starts take
0.4s Core queues 'run adlib' call (at this rundownPlaylistSyncFunction). 
5.0s Core completes take. current part is now 'b'
5.1s Core starts this 'run-adlib'.
	rundownPlaylist.currentPartInstance is now not in sync with the db, so does the wrong thing

meteor/client/ui/App.tsx
chrisryanouellette: I originally had a fragment but when the `Settings.enableUserAccounts` returns it, the Switch thinks it is the component to render instead of checking the sub-components.

PR #239, opened 2020-05-29
Update to meteor 1.10

Julusian:
Replaces https://github.com/nrkno/tv-automation-server-core/pull/151
Since Meteor 1.10 has now been released, it's time for us to upgrade the platform to it.
This gives us several improvements, including Node 12 and Typescript 3.8

This is should not affect any behaviour of the system.
However, when upgrading a system to this, mongodb will have to be updated.

## Key changes

1) `meteor run` no longer reports any typescript errors. Instead there are `check-types` and `watch-types` npm scripts which invoke tsc to check the typings. This is run as part of ci and will cause a failure if it does not pass.
Instead, you can use `meteor npm run dev` which will run them together and provide you labelled output from both in the same window

1) When meteor starts it no longer loads every file under the server or client folder. Files must be imported from the main.ts/main.tsx file. When adding some code that needs to be invoked at startup, it must be referenced explicitly of those files, to ensure it gets loaded. TODO: should we restructure the files to make this more obvious and clearer?

1) This bumps the version of mongo significantly. Meteor makes no attempt at startup to upgrade the data and will instead crash as mongo will fail to start. The easy fix is to take a system snapshot before upgrading, delete your local `.meteor/local` folder to wipe the db and then reimport the snapshot. See the upgrading section below for full steps

1) To ensure that your node_modules is all updated correctly, it is recommended to delete node_modules, run `meteor --version` then `meteor npm install`. This ensures that npm and all the dependencies are installed with the correct version of node and npm.
If you get `npm ERR! Unsupported URL Type "npm:": npm:fibers@4.0.3` this means that it is using an old version of npm

### Production
As mongodb needs upgrading, this might be possible to do online if you have a replicaset with more than one node, which is best described by the mongo documentation. I do not know how meteor will handle the mongo version changing, especially over such a big jump so we recommend doing an offline upgrade, but hopefully in future the jumps will be small enough for it to be possible while online

1) Stop sofie core
1) Upgrade mongodb. You need to first [upgrade to 3.6](https://docs.mongodb.com/v3.6/release-notes/3.6-upgrade-replica-set/), then [on to 4.0](https://docs.mongodb.com/v4.0/release-notes/4.0-upgrade-replica-set/), and finally [to 4.2](https://docs.mongodb.com/v4.2/release-notes/4.2-upgrade-replica-set/)
1) Startup the new sofie

### Development
The easiest option is to do a clean checkout of sofie and either import a system snapshot, or to reconfigure from scratch.

If you want to upgrade in place (after this, the db will not work in older meteor):

1) With a meteor 1.6 branch, do `meteor update --release 1.7`
1) Start sofie. This is expected to fail to startup, but will do the mongo upgrade to 3.6
1) Once mongo is running, in a mongo shell execute `db.adminCommand( { setFeatureCompatibilityVersion: "3.6" } )`
1) do `meteor update --release 1.8.1`
1) Start sofie. This is expected to fail to startup, but will do the mongo upgrade to 4.0
1) Once mongo is running, in a mongo shell execute `db.adminCommand( { setFeatureCompatibilityVersion: "4.0" } )`
1) checkout the meteor 1.10 branch
1) `meteor --version` to get the latest meteor tools setup
1) Delete your node_modules folder
1) `meteor npm install` to install dependencies
1) Startup sofie. This time it should be successful
1) Once running, in a mongo shell execute `db.adminCommand( { setFeatureCompatibilityVersion: "4.2" } )`

**Status**


Julusian:
@jstarpl What was the motivation to use legally here? After a quick look at it myself I am not impressed with it.	

Firstly, after installing and running  `yarn legally` I get `1: Syntax error: "(" unexpected`. I have to invoke it with `yarn node ./node_modules/.bin/legally` for it to work. Its missing the `#!/usr/bin/env node` line that scripts need to be runnable like that on *nix. That I could live with though

Secondly, if a license doesn't meet the criteria it does report that, but then it exits with OK rather than an error, so circleci will never fail.
eg:
```
yarn node ./node_modules/.bin/legally -r --filter GPL --plain --border ascii
yarn node v1.22.4
Working on "node_modules". It might take a while...
   

   +------------------------------------------------------------------------+
   |                                                                        |
   |                                Reports                                 |
   |                                                                        |
   +------------------------------------------------------------------------+
   | 100% of the dependencies are not "GPL" (951/951)                       |
   +------------------------------------------------------------------------+
   

Done in 4.62s.
```
vs
```
yarn node-license-validator -p -d --allow-licenses MIT 
yarn run v1.22.4
$ /home/julus/Projects/sofietv/circle-test/node_modules/.bin/node-license-validator -p -d --allow-licenses MIT
Identified licenses: 0BSD, MIT
Invalid license: tslib@1.11.2: 0BSD
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

jstarpl:
Hmm... that's weird, I thought that it was something wrong with node on my system that made the script broken. The motivation was that node-license-validator is dead and needs to be replaced with something that is still maintained. I guess I'll wrap legally in a wrapper script, that will resolve both the execution problem and the exit code problem.

Julusian:
that sounds like a good plan, we do need to do something about that issue eventually.
That wrapper sounds like something we could later bundle in the eslint & tsconfig config package we want to make.

nytamin:
Merged into https://github.com/nrkno/tv-automation-server-core/pull/245

PR #244, opened 2020-06-15
Maintenance/no focused test linting

gundelsby:
* **What kind of change does this PR introduce?** enforces a linting rule that disallows committing test files which includes focused tests. This hinders accidentally disabling all our tests but one.

* **What is the current behavior?** You are allowed to commit code that disables all tests but one.

**Status**

### DIFF
--- a/meteor/__mocks__/helpers/jest.ts
+++ b/meteor/__mocks__/helpers/jest.ts
@@ -18,6 +18,7 @@ export function testInFiber(testName: string, fcn: Function) {
 }
 
 export function testInFiberOnly(testName: string, fcn: Function) {
+	// tslint:disable-next-line:no-focused-test
 	test.only(testName, async () => {
 		await runInFiber(fcn)
 	})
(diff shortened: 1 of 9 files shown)


Julusian:
Has this rule been customised to detect uses of `testInFiberOnly`?

gundelsby:
> 
> 
> Has this rule been customised to detect uses of `testInFiberOnly`?

No, it's just the one line ignore. So I guess you can circumvent this safeguard by calling that function. Not entirely sure what to do about it, if anything. Adding the linting rule makes is less likely to disable the tests, but not entirely impossible.

Julusian:
hmm yeah this isnt even configurable to add that.
that makes the value of using this in this repo questionable, as most of the tests use testInFiber, so would need to use testInFiberOnly instead of test.only

nytamin:
converting this back into a draft, while you guys look for a more appropriate solution

Julusian:
I spent the 15 minutes this morning to do this.
Its a simple copy of the rule, with the regex changed

Julusian:
Manually merged in 24fa60c172d48707b61da2b386b731cf62a84849
This was causing some weird typings issues that went away with discarding the large package-lock changes

PR #251, opened 2020-06-19
fix: Unique hotkey group names

thomasslee97:
Allows multiple instances of a layout / bucket / panel of a given type to exist side-by-side

Currently, hotkeys are registered using group names like `DashboardPanel` and `BucketPanel`, which means that if you have more than one dashboard panel / bucket panel in a shelf, they will remove each other's hotkeys and could register conflicting hotkeys.

The group name used by each panel / bucket / shelf layout is generated from the name of that component, as assigned in the showstyle settings. This means that they will not interfere with each other as long as they have been given unique names.

**Status**


nytamin:

PR #253, opened 2020-06-19
feat: JSON config manifest entry type

thomasslee97:
Adds the JSON config manifest entry type, which is automatically checked to be valid JSON.

An example use case is to allow Sofie admins to add a new DVE layout by building the layout on an ATEM and using the atem connection library to pull the JSON representation of this layout. They can then copy the JSON into showstyle settings, where it will be validated to prevent copying errors.

* **Other information**:
Requires https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/65
**Status**

### DIFF
--- a/meteor/client/lib/EditAttribute.tsx
+++ b/meteor/client/lib/EditAttribute.tsx
@@ -8,7 +8,7 @@ import { Mongo } from 'meteor/mongo'
 
 import { MultiSelect, MultiSelectEvent } from './multiSelect'
 import { TransformedCollection } from '../../lib/typings/meteor'
-import * as classNames from 'classnames'
+import ClassNames from 'classnames'
 
 interface IEditAttribute extends IEditAttributeBaseProps {
 	type: EditAttributeType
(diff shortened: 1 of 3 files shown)


meteor/client/lib/EditAttribute.tsx
```diff
@@ -8,6 +8,7 @@ import { Mongo } from 'meteor/mongo'
 
 import { MultiSelect, MultiSelectEvent } from './multiSelect'
 import { TransformedCollection } from '../../lib/typings/meteor'
+import * as classNames from 'classnames'
```
Julusian, line 11: I think this will need to be `import ClassNames from 'classnames'`. This import was changed when the library was updated for meteor 1.10

thomasslee97: Updated: https://github.com/nrkno/tv-automation-server-core/pull/253/commits/4e6e97434ea45f668f1a4acd8c6b50bdd73458ba

PR #256, opened 2020-06-22
Feat: Shelf layout selection

thomasslee97:
This adds the ability for users to select a shelf layout to use when opening a rundown. Shelf layouts can be exposed as standalone layouts, which will open the shelf as a standalone page. The users' selection is remembered per-client so each user can open the shelf layout that they last used in a single click next time. This PR also introduces colour picker and icon picker components which allow each layout to be assigned an easily identifiable icon using the font-awesome icon set.

In this case, the commentator shelf layout has been set as a layout that can be selected by the user as either a standalone shelf or as the layout to use for the shelf in the normal rundown view.

image

The properties associated with this are set per layout in the shelf layout section of showstyle settings.

image

**Additional Information**
There are a large number of icons to select from, so they are presented in a searchable grid.
image

**Status**


thomasslee97:
Something like this (still working on it pushing the row)? I've set it to highlight options on hover as well as they act like buttons 
image

thomasslee97:
I've constrained the width (currently 150px) 
image

jstarpl:
That's certainly a step in the right direction. I guess we should drop the shadow around the button as well. My guess is that this is used on a touchscreen? It might be a good idea to then provide `pointer: coarse` media variant styling so that the `RundownList` is easily accessible there (with bigger buttons and taller rows) and have just shorter, regular buttons and normal rows when not running on a touch screen. But I don't know if your setup will match a `pointer: coarse` query (I guess a desktop Chrome might not?), so that's up for you to decide.

thomasslee97:
We won't have access to that, but I have changed the dropdown arrow from being a small link to a div 20% width of the button, I have tested on a touchscreen and it seems to be forgiving enough.

meteor/client/lib/splitDropdown.tsx
```diff
@@ -0,0 +1,59 @@
...
+	elements: Array<React.ReactNode>
+	selectedKey: string
+	className?: string
+}
+
+interface IState {
+	expanded: boolean
+}
+
+export class SplitDropdown extends React.Component<IProps, IState> {
```
jstarpl, line 18: I have only now realized that the file name for this component is `splitDropdown.tsx` and the component name is `SplitDropdown`. In order to keep in line with our naming convention, this should be renamed to `SplitDropdown.tsx`, since it's a React component.

jstarpl:
I have made some changes to resolve the look-and-feel issue:

Lobby view
Expanded

Also, the layout selection button will only show, if there are layouts exposed using the checkboxes in layout configuration. I think this is now ready to merge, pending @nytamin 's approval.

nytamin:
Following the discussion earlier this week I'm merging this as-is for now.

_Note: There was some massive edits to the translations json files which I'm omitting in the merge._

PR #259, opened 2020-06-24
Feat/gui rundowns in playlist

nytamin:
This PR build on https://github.com/nrkno/tv-automation-server-core/pull/258

Adds GUI a element that's inserted between rundowns in the RundownView.

image

* **Other information**:

**Status**


meteor/client/ui/RundownView.tsx
```diff
@@ -2009,39 +2010,54 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
...
-							</ErrorBoundary>
-						)
+				console.log('this.props.segments', this.props.segments)
+
+				let elements: (JSX.Element | undefined)[] = []
+
+				let previousRundownId = ''
+				this.props.rundowns.forEach((rundown, index) => {
+					if (this.props.rundowns.length > 1) {
+						elements.push(<RundownDividerTimeline key={`rundown${index}`} rundown={rundown} />)
```
jstarpl, line 2020: The `<RundownDividerTimeline>` component name is very confusing, since it doesn't produce any timeline. I think it should be renamed to something like `<RundownDividerMarker>` or `<RundownDividerHeader>`.

meteor/client/ui/RundownView.tsx
```diff
@@ -2009,39 +2010,54 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
...
 					}
+					elements = elements.concat(
+						this.props.segments
+							.filter((segment) => segment.rundownId === rundown._id)
```
jstarpl, line 2024: This will perform very poorly and result in a long loading time and bad performance on takes. The resulting operations will be:
1) Scan the Segments collection and select segments matching the rundowns in the playlist and order them correctly. This is a pretty costly operation.
2) For every rundown, scan the entire array resulting from step 1. and find the rundowns matching the current rundown in the playlist. Note that this will run every time there is a significant change in the playlist or a segment is added.

meteor/client/ui/RundownView.tsx
```diff
@@ -2452,6 +2468,8 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
 										? t('This rundown has been unpublished from Sofie.')
 										: !this.props.studio
 										? t('Error: The studio of this Rundown was not found.')
+										: !this.props.rundowns.length
+										? t('This playlist is empty')
```
jstarpl, line 2472: Since we never refer to playlists in User-facing code, I think this needs to be rephrased? Or we need to use the word *Playlist* more, so that users know what this is. As-is this will just be confusing, and - unlike a missing `showStyleBase` this is something that can easily happen during normal operation.

meteor/client/ui/SegmentTimeline/RundownTimeline.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
+import * as _ from 'underscore'
+import { Rundown } from '../../../lib/collections/Rundowns'
+
+interface IProps {
+	key: string
+	rundown: Rundown
+}
+interface IState {}
+export class RundownDividerTimeline extends React.Component<IProps, IState> {
```
jstarpl, line 10: See my coment above regarding the naming of this component. Also, I don't think it needs the state, so maybe this could be a Stateless Function Component?

meteor/client/ui/RundownView.tsx
nytamin: Oh, I misunderstood the naming convention, I thought the components was called ...Timeline because they where "in a timeline view" :)

jstarpl: The _Timeline View_ is I think something that was born very recently, as far as I know, this was called the Rundown View, and each segment had it's own Timeline (hence: `SegmentTimeline`).

PR #262, opened 2020-06-25
Fix/various fixes

nytamin:
This PR contains various smaller fixes that comes from https://github.com/nrkno/tv-automation-server-core/pull/237
the most prominent is some stricter typings in the Meteor interfaces.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/codeControl.ts
+++ b/meteor/server/codeControl.ts
@@ -52,6 +52,33 @@ function getFunctionName<T extends Function>(parent: Function, fcn: T): string {
 		return name
 	}
 }
+export function MeteorWrapAsync(func: Function, context?: Object): any {
+	// A variant of Meteor.wrapAsync to fix the bug
+	// https://github.com/meteor/meteor/issues/11120
+
+	return Meteor.wrapAsync((...args: any[]) => {
+		// Find the callback-function:
+		for (let i = args.length - 1; i >= 0; i--) {
+			if (typeof args[i] === 'function') {
+				if (i < args.length - 1) {
+					// The callback is not the last argument, make it so then:
+					const callback = args[i]
+					const fixedArgs = args
+					fixedArgs[i] = undefined
+					fixedArgs.push(callback)
+
+					func.apply(context, fixedArgs)
+					return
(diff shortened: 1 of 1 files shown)


meteor/server/codeControl.ts
```diff
@@ -79,7 +79,7 @@ function syncFunctionInner<T extends Function>(
 	return Meteor.wrapAsync((...args0: any[]) => {
 		let args = args0.slice(0, -1)
 		// @ts-ignore
-		let cb: Callback = _.last(args0) // the callback is the last argument
+		let cb: Callback = _.last(_.compact(args0)) // the callback is the last (non-falsy) argument
```
Julusian, line 82: What was the motivation for this?
The meteor docs state the the callback will always be the final parameter https://docs.meteor.com/api/core.html#Meteor-wrapAsync

If your change is valid, then the slice a few lines above should be updated too, as right now it will be not always be removing the callback when we execute fcn.

nytamin: gaah, I'm sure i _had_ a motivation for it when I wrote it, but I can't remember exactly where the issue was, and I seem to not be able to reproduce it now..
I'm going to pull this out for now, but leave this comment for posterity, in case the issue shows up again:
_If I remember correctly, I had an issue where the last argument was undefined for some reason. If I remember correctly (and I don't think I do, because I wasn't able to reproduce it now) the Meteor.wrapAsync was able to handle it so I made this to be as analogous to that as possible._

nytamin: Okay, I've had my morning tea now, and figured out the underlying issue.
It turns out there is a bug in Meteor.wrapAsync that causes `myWrappedFunction(1, undefined, undefined, undefined)` to generate the arguments `[1, CallbackFunction, undefined]` instead of the expected `[1, undefined, undefined, CallbackFunction]`. I have tracked down the issue and [submitted a bug report on it](https://github.com/meteor/meteor/issues/11120).

Meanwhile, I'll reimplement the "fix" and take your note on the slice into account, that was a good spot! :)

PR #266, opened 2020-06-30
Feat/multiple playout gateways

nytamin:
This PR intends to add support for having multiple playout-gateways connected to Core (in the same studio).

When having multiple gateways, the "Now"-feature in timeline cannot be used.
Instead this PR aims for continuous measuring what the roundtrip latency is for the timeline to be resolved in playout-gateways/TSR. Then the worst gateway gets to decide what the "Now" is going to be set to directly upon timeline generation in Core.

This depends on https://github.com/nrkno/tv-automation-playout-gateway/pull/66 in playout gateway

**Status**


Julusian:
A couple of small typos as discussed.
This should be enabled/disabled via a config instead, then I am happy for it to be merged to r23

nytamin:
Just because it was to easy, I made it write the reported playout-gateway executionTime and resolveTime into the userActions log, so it's viewable in the GUI:
image

The numbers mean
[Core execution time] + [time for playout-gateway to receive and resolve timeline] (resolve-time in PGW)

Julusian:
In server/api/peripheralDevice.ts `timelineTriggerTime()` is doing a `cache.Timeline.update()`, but will not updating the hash property. And there are another couple in server/api/manualPlayout.ts that should be updated too.

Other than that I am happy

nytamin:
> In server/api/peripheralDevice.ts `timelineTriggerTime()` is doing a `cache.Timeline.update()`, but will not updating the hash property. And there are another couple in server/api/manualPlayout.ts that should be updated too.
> 
> Other than that I am happy

Good find! I've fixed `timelineTriggerTime()`, but since manualPlayout is about to be removed I ignore that one.

PR #268, opened 2020-07-01
chore: Include vscode workspace file

nytamin:
This PR includes the vscode workspace files into the repo.
I intend this PR to be the basis of discussion on
1) if the file should be included at all
2) what settings should go in there

To be acted upon after the team has had a chance to discuss it (that'll be after the summer break, in August).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-	"typescript.tsdk": "node_modules\\typescript\\lib",
+	"typescript.tsdk": "./meteor/node_modules/typescript/lib",
 	"files.eol": "\n",
 	"editor.formatOnSave": true,
 	"[json]": {
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+	"typescript.tsdk": "node_modules\\typescript\\lib",
+	"files.eol": "\n",
+	"editor.formatOnSave": true,
+	"[json]": {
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+	"typescript.tsdk": "node_modules\\typescript\\lib",
+	"files.eol": "\n",
+	"editor.formatOnSave": true,
```
baltedewit, line 4: Note to self: what happens when a user does not have the prettier plugin installed and VS Code does the formatting?

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+	"typescript.tsdk": "node_modules\\typescript\\lib",
```
Julusian, line 2: This will need to be "meteor\\node_modules\\typescript\\lib"
Also should the slashes be \\ or / ? We should make sure to test on both windows and linux to make sure it is happy.

.vscode/settings.json
nytamin: Good point!
The docs for `typescript.tsdk` says
> Format a file on save. A formatter must be available, the file must not be saved after delay, and the editor must not be shutting down.

So I assume that if there's no formatter installed at all, no formatting is done.

In my latest commit I also added which default formatter to use

nytamin:
After a group discussion I'm dropping this. Everyone uses different types of workspaces so this doesn't make sense.

PR #269, opened 2020-07-01
fix: service messages types

jstarpl:

This is a bugfix for a typing/database serialization issue with Service Messages.

At some points, there will be an Error message logged to console:
```TypeError: message.timestamp.getMilliseconds() is not a function```

No error message should be thrown. Existing ServiceMessages should be upgraded by migrations.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/serviceMessages/serviceMessagesApi.ts
+++ b/meteor/server/api/serviceMessages/serviceMessagesApi.ts
@@ -1,7 +1,7 @@
 import { CoreSystem, getCoreSystem, ServiceMessage, ExternalServiceMessage } from '../../../lib/collections/CoreSystem'
 import { logger } from '../../logging'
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: meteor/server/api/serviceMessages/serviceMessagesApi.ts [hunk 1]
```diff
@@ -27,6 +27,13 @@ function readAllMessages(): Array<ServiceMessage> {
 	return messages
 }
 
+function convertExternalToServiceMessage(message: ExternalServiceMessage): ServiceMessage {
```
nytamin, line 30: It looks like this function is only ever used in the tests, perhaps it should be moved to `serviceMessagesApi.test.ts` instead?

jstarpl: I moved the function to `serviceMessagesApi.test.ts`.

PR #273, opened 2020-07-08
Feat: Pad piece start when created by action

thomasslee97:
This aims to change the behavior of the adLib action function `insertPiece` such that when you insert a piece into the currently playing part, it is inserted relative to the playhead position rather than the start of the part. i.e. a piece with `start: 0, end: 10` would become `start: 0 + playhead, end: 10 + playhead`.

* **Other information**:
Perhaps this should be behind a flag like `insertAtPlayhead: boolean`?

**Status**

### DIFF
--- a/meteor/server/api/blueprints/context/adlibActions.ts
+++ b/meteor/server/api/blueprints/context/adlibActions.ts
@@ -269,6 +269,20 @@ export class ActionExecutionContext extends ShowStyleContext implements IActionE
 			throw new Error(`Piece with id "${rawPiece._id}" already exists`)
 		}
 
+		const lastStartedPlayback = partInstance.part.getLastStartedPlayback()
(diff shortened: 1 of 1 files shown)


jstarpl:
Why not just use start: 'now'?

Julusian:
> Why not just use start: 'now'?

You cant do `now + 50` or anything, so using now would always give the current time.

This change ties into the infinites-rewrite work I am doing, as 'now' is giving me some pain to figure out how to handle. An `insertAtPlayhead: boolean` option would make my life simpler there, but then core will need to predict the latency of playout, to minimise the risk of missing the start of the inserted piece. cc @nytamin

thomasslee97:
No longer relevant, using `start: 'now'` works in the use cases we have

Julusian:
I think that this is still a relevant issue, but it needs a more flexible fix so something to revisit in the future

PR #274, opened 2020-07-08
Adlib actions in dashboard

thomasslee97:
This PR includes a number of changes to allow adLib actions to display and be executed withing dashboard layouts.

Changes include:
- Copying tags for adLib actions
- Fixed a bug where two memoizedIsolatedAutorun functions had the same name but different return types causing dashboards to crash
- Call the correct user action when taking an adLib action from a dashboard
- Sort adLibs and actions together according to rank

**Status**


thomasslee97:
I have found a bug in this PR involving the order in which hotkeys are assigned to shortcuts, it is not yet safe to merge

thomasslee97:
The above fix has fixed the issue with hotkeys not getting assigned properly

PR #282, opened 2020-07-27
Allow adlib actions to call take

thomasslee97:
This adds the function `takeAfterExecuteAction` which sets a flag on the action execution context of a given action. If this flag is set after the action has finished executing, take will be called.

Currently, the flag is set to the boolean value passed to `takeAfterExecuteAction` directly, allowing an action to request a take and then retract this request later in its execution. The value of the flag is returned to the caller, so if safety checks are added in the future an action can react to being unable to take.

* **Other information**:
Requires https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/70

**Status**

### DIFF
--- a/meteor/server/api/playout/playout.ts
+++ b/meteor/server/api/playout/playout.ts
@@ -93,7 +93,7 @@ import { rundownPlaylistSyncFunction, RundownSyncFunctionPriority } from '../ing
 import { ServerPlayoutAdLibAPI } from './adlib'
 import { PieceInstances, PieceInstance, PieceInstanceId } from '../../../lib/collections/PieceInstances'
 import { PartInstances, PartInstance, PartInstanceId } from '../../../lib/collections/PartInstances'
-import { ReloadRundownPlaylistResponse } from '../../../lib/api/userActions'
+import { ReloadRundownPlaylistResponse, UserActionAPIMethods } from '../../../lib/api/userActions'
 import {
 	initCacheForRundownPlaylist,
 	CacheForRundownPlaylist,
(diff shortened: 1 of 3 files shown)


meteor/server/api/playout/playout.ts
```diff
@@ -1483,7 +1486,13 @@ export namespace ServerPlayoutAPI {
 			}
 
 			waitForPromise(cache.saveAllToDatabase())
+
+			takeAfterExecute = context.takeAfterExecute
 		})
+
+		if (takeAfterExecute) {
+			Meteor.call(UserActionAPIMethods.take, 'e', rundownPlaylistId)
```
jstarpl, line 1494: Since this is called from the server-side, there's no reason why this should use the UserActionAPI, the 'e' string as userAction description will only be confusing (as no such action is actually made by the user). I think this just be a straight call to `ServerPlayoutAPI.rundownTake()`.

nytamin:
I like the idea of being able to do this from an adlib-action.

I'd like to discuss this a little bit more before we pull it in, let's make a call some time.
My thoughts are that I thing that the implementation should perhaps be a little more generic, and include calls to other methods as well.

Ideas for questions to discuss:
* Should the BP be able to call any method? or just a few?
* Should the BP be able to call multiple methods, to be run after eashother?
* Should the BP be able to call another adlib-action after the method(s) have been executed?

If it's important to get this in quickly, I'm fine with this implementation as a first step. 👍

Julusian:
@nytamin do you know what other methods might be wanted?
We need to be very careful with what we expose here, and how they get executed as I expect that some will want to be run at the end, but others might want to be executed immediately and work on the same cache (eg moveNext). So to me exposing them individually like this would be preferable, it will also let us more clearly document the requirements/limitations and exectution order for each.

Some other of the other bits of the adlib-action api are also user actions, but they are useful building blocks so were made available from the beginning (but integrated closely, rather than via the user actions)

I think a call to discuss would be a good idea.

Julusian (approved):

nytamin:

PR #286, opened 2020-07-30
feat: i18n bundles from blueprints

gundelsby:
i18n data contained in blueprint bundles is added to the existing translator function in core, making it possible to have localised content in blueprint code

* **Other information**:
Changes in the blueprints project:
https://github.com/nrkno/tv-automation-sofie-blueprints/pull/65

Changes in the blueprints-integration project:
https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/71

**Status**


meteor/client/ui/i18n.ts
```diff
@@ -2,49 +2,76 @@ import * as i18n from 'i18next'
...
+			.use(reactI18nextModule)
+			.init(i18nOptions, (err, t) => {
+				if (err) {
+					console.error('Error initializing i18Next', err)
+				} else {
+					this.i18nTranslator = t
+				}
+			})
+
+		this.subscribe(PubSub.translationsBundles, null)
```
jstarpl, line 57: Considering the possible amount of data this may load (if we have a lot of various translations), it might be a good idea to narrow this subscription down. If my memory serves me right, Meteor publications use up some memory per connection/subscription depending on the size of that subscribed data set. So it's a good idea to keep the subscriptions to a minimum. Perhaps this could be integrated as an additional i18next-backend so that only the locales requested by i18next (based on the used detection scheme) would be subscribed to? Just a thought.

gundelsby: I'm not entirely sure of what constitutes as a large amount of data in this context, but the complete set of translations for core right now (which is the two variants of Norwegian) comes in at <90 kB of JSON data.
I'd expect the translations coming from blueprints to be significantly less than that, given that most of the text content should be in core. But I could be wrong :)

gundelsby:
> 
> 
> Yeah, this looks right. I don't know if this (https://github.com/nrkno/tv-automation-server-core/pull/286/files#diff-dad48b273d5c4671b0fa2c012182194fR61) is going to work, because I don't know how react-i18next reacts to injected bundles after a string has been rendered, but I would guess that there is some way to trigger a rerender of the UI when the bundles are injected.

I would imagine it simply does nothing? The translation function just works with the data it has at call time, and then returns a string. If the component that called doesn't render again then it won't be called again, and everything just stays as it is.
Do we not re-render after importing blueprints today?

gundelsby:
Closed due to branching out from the wrong parent. New PR here: https://github.com/nrkno/tv-automation-server-core/pull/287

PR #287, opened 2020-07-31
Blueprint contents localisation

gundelsby:
i18n data contained in blueprint bundles is added to the existing translator function in core, making it possible to have localised content in blueprint code

* **Other information**:
Changes in the blueprints project:
https://github.com/nrkno/tv-automation-sofie-blueprints/pull/65

**Status**


gundelsby:
Closed due to branching out from the wrong parent. New PR here: https://github.com/nrkno/tv-automation-server-core/pull/287

meteor/server/api/ingest/bucketAdlibs.ts
```diff
@@ -29,7 +29,12 @@ export function updateBucketAdlibFromIngestData(
 		studio,
 		showStyle._id,
 		showStyle.showStyleVariantId,
-		new NotesContext('Bucket Ad-Lib', 'bucket-adlib', false)
+		new NotesContext(
+			'Bucket Ad-Lib',
+			'bucket-adlib',
+			false,
+			[blueprintId, studio.blueprintId].map(unprotectString).filter((id): id is string => id !== undefined)
```
jstarpl, line 36: What's this about? You can use `unprotectStringArray` if you need to pass some protectedStrings array as a string array.

jstarpl (changes requested):
I think it might be worth to convert the subscription/reactive query to a periodic method call, as this might be a pretty substantial subscription (in terms of memory size).

meteor/client/ui/i18n.ts
```diff
@@ -2,52 +2,125 @@ import i18n, { TFunction } from 'i18next'
...
-				console.error('Error initializing i18Next', err)
+				console.error('Error initializing i18Next:', err)
 			} else {
-				i18nTranslator = t
+				this.i18nTranslator = t
+				console.debug('i18nTranslator init complete')
 			}
+		})
+
+		this.subscribe(PubSub.translationsBundles, {})
```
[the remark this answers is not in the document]
gundelsby: Yeah, that's fair critique. This change has been put on hold until after the performance sprint anyway, but I don't think we should release this until we've had a proper look at this either.

meteor/server/api/blueprints/api.ts
```diff
@@ -149,6 +154,16 @@ export function uploadBlueprint(
 		newBlueprint.studioConfigManifest = blueprintManifest.studioConfigManifest
 	}
 
+	// extract and store translations bundled with the manifest if any
+	logger.debug(`blueprintManifest for ${newBlueprint.name} translations`, {
+		translations: (blueprintManifest as any).translations,
+		type: typeof (blueprintManifest as any).translations,
+	})
+	if ((blueprintManifest as any).translations) {
```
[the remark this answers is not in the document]
gundelsby: It's only added to the StudioBlueprintManifest and ShowStyleBlueprintManifest, not the SystemBlueprintManifest. SomeBlueprintManifest  can be either of those. The changes are in a nightly version which is built for this branch like you suggested.

We could keep it as is, I could explicitly check for those types, or we could add it to the Some-type. I don't really feel strongly about any of the options.

meteor/server/api/ingest/bucketAdlibs.ts
gundelsby: That's a good tip, thanks. I'll push an update 👍

jstarpl: I was just hoping that there's a way of not running the map and filter. I don't think either `blueprintId` or `studio.blueprintId` can be undefined (as per typings), so just a simple overload should be fine.

Julusian: this has been changed in #336, which will be merged in at the same time, so isnt worth worrying about

meteor/server/api/blueprints/api.ts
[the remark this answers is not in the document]
jstarpl: Let's have @nytamin decide, he's the "strict typings" ombudsman.

meteor/client/ui/i18n.ts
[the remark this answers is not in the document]
nytamin: I agree, since this is going to be basically static data, a method call will be much lighter than setting up a subscription

meteor/lib/api/rundown.ts
```diff
@@ -81,9 +81,13 @@ export function runInRundownContext<T>(rundown: Rundown, fcn: () => T, errorInfo
 function handleRundownContextError(rundown: Rundown, errorInformMessage: string | undefined, error: any) {
 	rundown.appendNote({
 		type: NoteType.ERROR,
-		message:
-			(errorInformMessage ? errorInformMessage : 'Something went wrong when processing data this rundown.') +
-			`Error message: ${(error || 'N/A').toString()}`,
+		message: {
+			key: `${errorInformMessage ||
+				'Something went wrong when processing data this rundown.'} Error message: {{error}}`,
```
nytamin, line 86: typo

meteor/server/api/translationsBundles.ts
```diff
@@ -0,0 +1,46 @@
...
+				const dbCursor = TranslationsBundles.find({})
+				const availableBundles = dbCursor.count()
+				const bundles = dbCursor.fetch().map(({ _id, namespace, language }) => ({ _id, namespace, language }))
+				logger.debug(`${availableBundles} bundles in database:`, { bundles })
+			}
+		)
+	}
+}
+
+function getExistingId(namespace: string | undefined, language: string): TranslationsBundleId | null {
```
nytamin, line 42: This only creates an implicit (&loose) connection between the namespace-language pair and the _id. I would very much prefer that to be an explicit one (which also should include the _id of the blueprint that created it, to ensure uniqueness).
Example: `return getHash('${blueprint}_${namespace}_${language}')`.

meteor/lib/collections/TranslationsBundles.ts
```diff
@@ -0,0 +1,17 @@
...
+import { registerCollection, ProtectedString } from '../lib'
+
+import { TranslationsBundle as BlueprintTranslationsBundle } from 'tv-automation-sofie-blueprints-integration'
+import { createMongoCollection } from './lib'
+
+/** A string identifying a translations bundle */
+export type TranslationsBundleId = ProtectedString<'TranslationsBundleId'>
+
+export interface TranslationsBundle extends BlueprintTranslationsBundle {
+	_id: TranslationsBundleId
```
nytamin, line 11: I'm missing a `blueprint: BlueprintId` here, so that a bundle is explicitly tied to the blueprint that created it.

meteor/server/api/blueprints/api.ts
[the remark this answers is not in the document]
nytamin: I'm in favor of the stricter types solution, as that decreases the risk of future bugs during refactoring.

meteor/server/api/playout/playout.ts
```diff
@@ -1209,14 +1208,19 @@ export namespace ServerPlayoutAPI {
...
-						currentPartInstance._id
-					},execution=${getRandomId()}`,
-					false
+				const actionContext = new ActionExecutionContext(
+					{
+						name: `${rundown.name}(${playlist.name})`,
+						identifier: `playlist=${playlist._id},rundown=${rundown._id},currentPartInstance=${
+							currentPartInstance._id
+						},execution=${getRandomId()}`,
+						tempSendUserNotesIntoBlackHole: true, // TODO-CONTEXT store these notes
```
nytamin, line 1217: todo

meteor/server/api/translationsBundles.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function upsertBundles(bundles: TranslationsBundle[], parentBlueprintId: BlueprintId) {
+	for (const bundle of bundles) {
+		const { type, language, data } = bundle
+
+		if (type !== TranslationsBundleType.I18NEXT) {
+			throw new Error(`Unknown bundle type ${type}`)
+		}
+
+		const namespace = unprotectString(parentBlueprintId)
```
nytamin, line 15: So I'm confused about the "namespace". If it is equal to the _id of the blueprint that created it, I think it should be named `blueprintId`

meteor/server/api/translationsBundles.ts
```diff
@@ -0,0 +1,46 @@
...
+					logger.info(`Stored translation bundle ([${insertedId || _id}]:${namespace}:${language})`)
+				} else {
+					logger.error(`Unable to store translation bundle ([${_id}]:${namespace}:${language})`, {
+						error: err,
+					})
+				}
+				const dbCursor = TranslationsBundles.find({})
+				const availableBundles = dbCursor.count()
+				const bundles = dbCursor.fetch().map(({ _id, namespace, language }) => ({ _id, namespace, language }))
+				logger.debug(`${availableBundles} bundles in database:`, { bundles })
```
nytamin, line 36: This is just for debugging during dev, and should be removed before merging, right?

meteor/server/api/translationsBundles.ts
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		const namespace = unprotectString(parentBlueprintId)
+		const _id = getExistingId(namespace, language) || getRandomId<'TranslationsBundleId'>()
+
+		TranslationsBundles.upsert(
+			_id,
+			{ _id, type, namespace, language, data },
+			{ multi: false },
+			(
```
nytamin, line 22: Is there a reason for using the callback pattern here?
If there is an error it wont be displayed to the user, which might be unhelpful.

meteor/server/publications/translationsBundles.ts
```diff
@@ -0,0 +1,16 @@
+import { Meteor } from 'meteor/meteor'
+
+import { TranslationsBundles } from '../../lib/collections/TranslationsBundles'
+import { TranslationsBundlesSecurity } from '../security/translationsBundles'
+import { meteorPublish } from './lib'
+import { PubSub } from '../../lib/api/pubsub'
+
+meteorPublish(PubSub.translationsBundles, (selector, token) => {
```
nytamin, line 8: to be removed, if we use a method instead

nytamin (changes requested):
A few notes, most of which I have talked to @gundelsby about before (but to keep track I put them in writing now as well).

meteor/server/api/translationsBundles.ts
gundelsby: Yup

gundelsby:
This is now working, so I've converted the PR from a draft and will start working on the QA taking into consideration the previous comments from @jstarpl and @nytamin that are still valid.

jstarpl (approved):

meteor/client/ui/i18n.ts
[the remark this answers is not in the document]
nytamin: I'm bumping this up again, as I still think it would be better to use a method than a subscription, since the data is going to be static.

meteor/lib/api/userActions.ts
```diff
@@ -160,6 +161,7 @@ export interface NewUserActionAPI extends MethodContext {
 	restartCore(userEvent: string, token: string): Promise<ClientAPI.ClientResponse<string>>
 	guiFocused(userEvent: string, viewInfo?: any[]): Promise<ClientAPI.ClientResponse<void>>
 	guiBlurred(userEvent: string, viewInfo?: any[]): Promise<ClientAPI.ClientResponse<void>>
+	getTranslationBundle(bundleId: TranslationsBundleId): Promise<ClientAPI.ClientResponse<TranslationsBundle>>
```
nytamin, line 164: I don't think it's a good idea to use the UserActions API for fetching the translations, as that is tied to user-facing things (since it logs an entry to the userActions-log). For this I would instead just expose a plain Meteor method. It's not as type-safe, but it is what we have.

meteor/server/api/translationsBundles.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export function upsertBundles(bundles: BlueprintTranslationsbundle[], originBlueprintId: BlueprintId) {
+	for (const bundle of bundles) {
+		const { type, language, data } = bundle
+
+		if (type !== TranslationsBundleType.I18NEXT) {
+			throw new Error(`Unknown bundle type ${type}`)
+		}
+
+		const _id = getExistingId(originBlueprintId, language) || createBundleId(originBlueprintId, language)
```
nytamin, line 31: Question: Is there a specific reason for the lookup with `getExistingId`? Why not just use createBundleId right away? (that would be more performant). If there is a reason, plz add it as a comment in the code.

meteor/server/api/blueprints/context/context.ts
```diff
@@ -184,30 +146,60 @@ export class StudioConfigContext implements IStudioConfigContext {
...
 
-export class ShowStyleContext extends StudioContext implements IShowStyleContext {
-	readonly notesContext: NotesContext
+	constructor(contextInfo: UserContextInfo, studio: Studio) {
+		super(contextInfo, studio)
+		this.tempSendNotesIntoBlackHole = contextInfo.tempSendUserNotesIntoBlackHole ?? false
+	}
+
+	notifyUserError(message: string, params?: { [key: string]: any }): void {
+		if (this.tempSendNotesIntoBlackHole) {
```
nytamin, line 165: Should this temporary feature be removed?

meteor/server/api/userActions.ts
```diff
@@ -1058,6 +1066,9 @@ class ServerUserActionAPI extends MethodContextAPI implements NewUserActionAPI {
 	guiBlurred(_userEvent: string, _viewInfo: any[]) {
 		return traceAction('userAction.noop', noop, this)
 	}
+	getTranslationBundle(bundleId: TranslationsBundleId): Promise<ClientAPI.ClientResponse<TranslationsBundle>> {
```
nytamin, line 1069: See my other comment; I think this should be a MeteorMethod, not a UserAction.

nytamin (changes requested):

meteor/server/api/blueprints/context/context.ts
Julusian: we dont have a way of getting these to the user currently, but they really should be.
so this is a hack to make them go somewhere (rather than into the void) until a follow up task is written and done

Julusian: but I am thinking I will end up making the blueprints do a log* everytime they do a notifyUser* so that we have a record of the note with more technical details, so maybe this isnt necessary?

meteor/lib/api/userActions.ts
Julusian: I agree, the purpose of the UserActions API is to provide a record of what the user did, and this does not want to be logged like that

jstarpl:
@gundelsby Looks good to me!

PR #288, opened 2020-08-03
Feat/nrcs name

thomasslee97:
This adds the meteor/core setting `nrcsName` which allows the changing of text containing "ENPS" or "MOS" to the name of other systems e.g. "iNews". The default name is "ENPS". This is an intermediary step to allow easy customization, in the future a better approach would be for ingest gateways to report the names to use.

**Status**


Julusian:
This might be better to do with a setting on the studio rather than a system setting?
I can see it being possible in future for an installation to want to use different ingest methods for different studios
What do you think @nytamin?

nytamin:
Yes, ideally it should come from the Gateway (that talks to the NCS) somehow, but I could be ok with a temporary solution of having it in on a Studio.

nytamin:
Following a discussion with @jesperstarkar, I looked into an alternative approach where the NRCS name is stored on the rundown instead.

It's more or less done, although I need to confer with @jstarpl and test it before it can be considered done:
https://github.com/nrkno/tv-automation-server-core/compare/release23...olzzon:feat/nrcs-name-rundown

thomasslee97:
@nytamin you may need to apply the same fix as https://github.com/nrkno/tv-automation-server-core/pull/288/commits/11e80cd7d6bd454af8e4e6490bf0877b30b80a9a a comma was in the wrong place

jstarpl:
> 
> 
> Following a discussion with @jesperstarkar, I looked into an alternative approach where the NRCS name is stored on the rundown instead.
> 
> It's more or less done, although I need to confer with @jstarpl and test it before it can be considered done:
> [release23...olzzon:feat/nrcs-name-rundown](https://github.com/nrkno/tv-automation-server-core/compare/release23...olzzon:feat/nrcs-name-rundown)

Yeah, I think I like that approach much better. This way, it can come straight from the gateway, a single studio can (potentially) use multiple NRCS/rundown sources. I think that way makes much more sense than a system- or studio-wide config.

jstarpl:
Since @nytamin has already implemented it, I would be for forgoing this intermediate step and go straight to the solution: https://github.com/nrkno/tv-automation-server-core/pull/298 . Adding this information is not an extreme burden on the gateways, and allows much cleaner implementation.

nytamin:
closing this in favor of #298

PR #289, opened 2020-08-03
fix: Hide script line when source duration is undefined

thomasslee97:
When the duration of a script piece is set to undefined, the "end of script" line is hidden. This also adds a class ('at-end') when the end of script line is positioned at the end of a part, which is used by TV 2 to hide the line in this case through custom css not included in this PR. The use case is parts where the script is Out On Next Part and the duration of the script is either not known, or is tied directly to the duration of the part - so will be expected to be the same duration as the part.

**Status**


nytamin:

PR #290, opened 2020-08-03
Fix: measure header height

thomasslee97:
* **What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

Does not use a hard-coded value for Header height, but instead attempts to measure it from the DOM.

* **What is the current behavior? (You can also link to an open issue here)**

The header height is a hard-coded constant, thus causing some problems with Follow On Air (broadly speaking).

The height of the header is measured when it's first used, allowing customizations of the header to not affect "Follow On Air".

**Status**


thomasslee97:
Closing, need to reopen with a better base branch

PR #295, opened 2020-08-05
fix: race condition in refresh part function

thomasslee97:
Fixes a bug where `refreshParts` would cause a data race and crash.

The `_initialize` function is called once for every part that is being refreshed.

The `_initialize` should be called once per segment that is affected by the refresh. There is also a lock placed on `_initialize` so that subsequent calls wait for previous to finish executing to avoid data races.

* **Other information**:
CC @Julusian 

**Status**


Julusian:
@nytamin can you give this a quick look? This will want to be merged into nrk asap, as it is an issue that will affect them too.

Its a 2 part fix, the first that the lazy initialisation in the DatabaseCache could be run in parallel if using the collection inside a `Promise.all()` or a `waitForPromiseAll`.
The second part is more of a performance fix where we would call `updateSegmentsFromIngestData` multiple times for the same segment if multiple parts in a segment were set as dirty

PR #296, opened 2020-08-10
Chore: reformatting organize imports

nytamin:
This PR adds the `prettier-plugin-organize-imports` which sorts & removes unused imports.

I'm just putting this here, pending an upcoming discussion whether the team want to implement it or not.


nytamin:
There is a strange linting issue on this one. @gundelsby is going to look at if it's possible to do a quick fix, otherwise we'll delay this until later when everything is a bit less in flux.

gundelsby:
Pretty sure the problem is that it the plugin doesn't recognize statements inside a Markup section in React components. All the files that fails the CI checks are TSX files.
I did a quick test on Timeline.tsx, moving the calls to an imported function out of the markup, and that commit did not get any imports removed.

If we are to move forward with this plugin we'd also have to find a rule that enforces no statements within markup segments. I don't know if such a plugin exists, or if it's desirable to introduce a rule like that. Personally I think it makes for cleaner code, but I also assumes it will be quite the job to refactor statements out of markup segments.

jstarpl:
> 
> 
> Pretty sure the problem is that it the plugin doesn't recognize statements inside a Markup section in React components. All the files that fails the CI checks are TSX files.
> I did a quick test on Timeline.tsx, moving the calls to an imported function out of the markup, and that commit did not get any imports removed.
> 
> If we are to move forward with this plugin we'd also have to find a rule that enforces no statements within markup segments. I don't know if such a plugin exists, or if it's desirable to introduce a rule like that. Personally I think it makes for cleaner code, but I also assumes it will be quite the job to refactor statements out of markup segments.

I disagree with the "cleaner code" argument - I'd say it will just make the code unnecessarily verbose and thus far less clean. I think the `ClassNames` (which makes state-based classes much easier) and the `unprotectString` for `key` showcase this best. If we needed to have a variable for every React `key` and a string variable for every `className` then it's just going to be unnecessary lines and lines of trivial declarations with variables used once. I don't see any practical reason to introduce such a restriction.

Julusian:
So I guess either we open an issue upstream and put this on hold, or tslint and eslint have similar rules that we could use instead https://palantir.github.io/tslint/rules/ordered-imports/ which I would expect to be more mature and not have issues like this

nytamin:
Yeah, I'll close this for now and we can look into this at a later time, probably along when moving we'll to eslint.

PR #302, opened 2020-08-20
bugfix in removeRundownFromCache: make sure all data that belong to Rundown is removed

nytamin:
Bug fix

Some data is not removed from db when removing a rundown, due to the cache only contains pieceInstances from the current, next and previous partinstances, so we end up leaving behind any pieceinstances from any other/earlier partinstances

Also some collections that had data owned by rundowns was not handled.

All data is removed

* **Other information**:

This is targeted to R22, but we should also merge it into R23+.

**Status**


Julusian (approved):
LGTM

nytamin:
Note: I've merged this into R23 & R24 as well

PR #309, opened 2020-08-26
Feat/tally tags

thomasslee97:
Adds tags to pieces that can be used by adLibs and adLib actions to signal when they should be considered to be on air or set as next.

Currently, only the `adLibSourceId` is used to determine whether an adLib is in the next / current part and should considered to be on on-air. This does not necessarily work for adLib actions where they could be creating multiple pieces or may create "data storage" pieces that do not currently have an effect on the show, but will if another condition is met. The same is true if, for example, one adLib creates a piece that contains timeline objects that need to be enabled by another piece. This PR hopes to establish an indicator of this relationship.

The above behaviour using `adLibSourceId` will continue to work as expected, however, if a piece with an `adLibSourceId` matching an adLibs ID is not found, Sofie will then look at the properties `onAirTags` and `setNextTags` on the adLib / adLibAction.

These tags are expected to be found on pieces either in the current or next parts. For example, if an adLib has the `onAirTags` property set to `["CAM_1", "CAM_LIVE"]` it will expect to find those tags distributed across the currently active pieces. It doesn't matter whether the tags are on a single piece or on multiple, as long as they are found within the active pieces.

The `setNextTags` have similar behaviour, except that they first check the next part and then check the currently active pieces. This allows an action which creates a "data storage" piece to be highlighted as set as next, "selected", (e.g. the next DVE layout to use) even though it is in the current part.

* **Other information**:
Requires https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/74

**Status**

### DIFF
--- a/meteor/client/styles/shelf/adLibRegionPanel.scss
+++ b/meteor/client/styles/shelf/adLibRegionPanel.scss
@@ -47,12 +47,12 @@
 			}
 		}
 
+		&.on-air {
+			border: 3px solid var(--general-live-color) !important;
+		}
+
 		&.next {
 			border: 3px solid var(--general-next-color);
-		}
-		
-		&.on-air {
-			border: 3px solid var(--general-live-color);
 		}
 	}
 }
--- a/meteor/client/styles/shelf/adLibRegionPanel.scss
+++ b/meteor/client/styles/shelf/adLibRegionPanel.scss
@@ -47,12 +47,12 @@
 			}
 		}
 
-		&.on-air {
-			border: 3px solid var(--general-live-color);
-		}
-
 		&.next {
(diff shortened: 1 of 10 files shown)


jstarpl:
How about naming these `currentPieceTags` and `nextPieceTags` ? I don't think we use the term "onAir" anywhere else in the core code, apart from the GUI code which uses that term sometimes, so this new term might be confusing.

thomasslee97:
I have renamed to `currentPieceTags` and `nextPieceTags` and brought the branch up to date with `release24` as it was starting to get to the point of conflicts becoming an issue. It should be a relatively simple rebase onto `release25` if that is the intended target

jstarpl (approved):
Looks alright!

PR #311, opened 2020-08-28
Fix: APM Slowdown

thomasslee97:
This wraps any calls to the APM Agent such that we will only make calls to `Agent` if APM is currently enabled on the system. This is because we have seen evidence of these Agent calls taking a lot of time, even if APM is not connected - particularly in places like `savePreparedChangesIntoCache` where we spend time looping over a lot of object.

**Status**

### DIFF
--- a/meteor/server/coreSystem.ts
+++ b/meteor/server/coreSystem.ts
@@ -539,7 +539,7 @@ function startInstrumenting() {
 			transactionSampleRate: system.apm.transactionSampleRate,
 			disableMeteorInstrumentations: ['methods', 'http-out', 'session', 'async', 'metrics'],
 		})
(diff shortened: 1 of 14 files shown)


meteor/server/coreSystem.ts
```diff
@@ -539,7 +539,7 @@ function startInstrumenting() {
 			transactionSampleRate: system.apm.transactionSampleRate,
 			disableMeteorInstrumentations: ['methods', 'http-out', 'session', 'async', 'metrics'],
 		})
-		profiler.setActive(true)
+		profiler.setActive(system.apm.enabled || true)
```
Julusian, line 542: Shouldnt this be `|| false`?
`undefined || true` = true
`false || true` = true
`true || true` = true

PR #312, opened 2020-08-31
Feat/clean up functions

nytamin:
This PR changes how indexes are set up, and adds two new methods:

* Manual Cleanup of old indexes (from earlier versions of Core)
* Manual cleanup of old data (that might be lying around from previous versions of Core, or due to old bugs)

These can be triggered from the GUI under Settings/System:

image

**Status**


Julusian:
Could the 'cleanup old indices' be made into a migration step?
The #cleanup old data' kind of makes sense as a ui button, but when we identify a leak we should be adding a migration to do the cleanup, so will this have any use?

nytamin:
Well, yes and no. Some data could be (and has been) orphaned during down+up-grades, and that could be only be caught in this kind of general cleanup.

I agree that migrations should have caught the issues before, but when it comes to it, the user has the ability to bypass them. If nothing else, this'll help us with cleanups in the deployments currently out there.

nytamin:
After a chat with @Julusian, this is good enough for now, we'll make an effort later to make these run automatically somehow.

PR #314, opened 2020-08-31
fix: Don't register sticky piece hotkeys twice in dashboard

thomasslee97:
Sticky hotkeys are treated as a kind of virtual adLib in `fetchAndFilter`, so get evaluated when we loop over `rundownBaselineAdLibs` in the dashboard. The block of code removed by this PR caused the hotkey to be registered twice as we have already dealt with sticky pieces in `fetchAndFilter`.

* **Other information**:

**Status**


Julusian:
Does this affect r23 too?

thomasslee97:
Yes it will

jstarpl (approved):
Seems like a safe change.

PR #315, opened 2020-08-31
fix: ensure older adlibs that never started dont play again later

Julusian:
What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This ensures that when adlibs are executed in rapid succession, we ensure that any adlibs that didnt get de-nowified (due to not having a chance to be played) get disabled, so they dont try to play later

When we run an adlib if another gets run soon after, the first may not get played out. But it gets left with a `start: 'now'` so at the next updateTimeline it may start playing

When we get the timelineTriggerTime callback from playout gateway, any adlibs inserted before the one being de-nowified get disabled.

Also, the prune pieces logic was not ignoring disabled pieces

* **Other information**:

**Status**


nytamin (approved):
Looks good upon read-through :)

PR #316, opened 2020-08-31
feat: Add method to remove MediaObjects with specific collectionId

ianshade:
This PR adds a `clearMediaObjectCollection` method to the `PeripheralDeviceAPI`. This method allows removing documents from the MediaObjects collection, where the given `collectionId` and the `studioId` of the device match.

There is no way to empty the MediaObjects collection other than removing each document explicitly.

The `clearMediaObjectCollection` removes all Media Objects where the given `collectionId` and the `studioId` of the requesting device match.

* **Other information**:
This can be used e.g. by the Playout Gateway to manage MediaObjects used for reporting statuses of Viz elements.

**Status**

### DIFF
--- a/meteor/lib/api/peripheralDevice.ts
+++ b/meteor/lib/api/peripheralDevice.ts
@@ -236,6 +236,7 @@ export interface NewPeripheralDeviceAPI {
 		id: string,
 		doc: MediaObject | null
 	): Promise<void>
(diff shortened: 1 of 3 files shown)


nytamin:
We're going to start up a rather large refactoring of the media-objects in the upcoming weeks.

I see no issues with this PR, so I'll merge this in to make sure this functionality is taken into account during that work.

PR #319, opened 2020-09-03
Add names to anonymous rundown playlist sync functions

sparkpunkd:

Bug fix for logging messages.

Issues reported during the logging of queued rundown playlist sync functions report:

    "Anonymous(rundownPlaylistSyncFunction)" completed after timeout. took 10598ms

This is because they are - for convenience and code readability - defined by anonymous functions.

A call to `rundownPlaylistSyncFunction` must now include the context.

* **Other information**:

I looked into auto-magic ways of doing this - discovering your context -  but `function.caller` has been deprecated and `arguments.callee` has been removed from _strict_ mode. All for good security reasons. 

So ... a slightly clunky manual setting of the context is used. However, this could be extended in the future to add more context if useful for debugging.

**Status**


sparkpunkd:
Now logs e.g. 

    [debug] syncFunction: ingest_rundown_undefined (triggerUpdateTimelineAfterIngestData)

meteor/server/api/ingest/rundownInput.ts
```diff
@@ -132,8 +132,12 @@ export enum RundownSyncFunctionPriority {
 export function rundownPlaylistSyncFunction<T extends Function>(
 	rundownPlaylistId: RundownPlaylistId,
 	priority: RundownSyncFunctionPriority,
+	context: string,
 	fcn: T
 ): ReturnType<T> {
+	if (!fcn.name) {
+		Object.defineProperty(fcn, 'name', { value: context })
+	}
 	return syncFunction(fcn, `ingest_rundown_${rundownPlaylistId}`, undefined, priority)()
```
nytamin, line 141: If going this way, I think we should add the context argument into the syncFunction as well. Thereby removing the need to use `getFunctionName` altogether (I've seen reports on that it performs poorly, performance-wise)

sparkpunkd: I agree. I'll take a look at extending this PR to all sync functions. Should just be _turning a handle_ across the code base.

Julusian: Personally I am leaning towards the context being a required parameter in all of the sync functions, and to always use that instead of `fcn.name`.
It does make the syncFunction api a little messier/harder to use, but means we can write some proper contexts of where/what the function is, which will aid in tracing and debugging

sparkpunkd:
Now logs context and function name (where available) e.g.: 
```
I20200920-13:57:09.996(1)? [debug] syncFunction: EkWQXHXYLAHtpxRst2QamzT4hXs_ (take - take)
```

meteor/server/codeControl.ts
```diff
@@ -27,30 +27,32 @@ interface SyncFunctionFcn {
...
-			} catch (e) {
-				// Probably reached the top?
-				break
-			}
-		}
-
-		return name
+		return context
 	}
+	// 	let name = 'Anonymous'
```
Julusian, line 36: Please delete this commented code

Julusian (approved):
Only read the key parts, but Im happy

nytamin (approved):

meteor/server/codeControl.ts
sparkpunkd: Done

PR #320, opened 2020-09-08
chore: remove old properties from TimelineObjGeneric

Julusian:
With the move to store `TimelineComplete` in the database, the individual objects don't need a studioId defined, or to have the secondary _id field. _id was kept as a guaranteed system unique copy of id to avoid data collisions if multiple studios were in a single sofie instance. This means that id can be left to be for the timeline and only needs to be unique within a timeline

This removes those two properties as they have no longer have a benefit

* **Other information**:

**Status**


nytamin (approved):
Looks like you got to it before me..! I agree, the properties are no longer needed and should be removed.

PR #321, opened 2020-09-08
Feat/move playout properties

Julusian:
This is a continuation of the PartInstances work.
It moves various timing properties from Piece to PieceInstance and Part to PartInstance.
This means that during playout, no Piece or Part is ever updated except for queued-parts.

This should help make the data structures cleaner, and makes the decoupling between Parts and PartInstances almost complete.

Additionally, this will let us set the on-air part as next. This has been allowed, and after a small fix mostly-works

* **Other information**:

**Status**


Julusian:
Before merging I can revert the change in `SegmentContextMenu.tsx`, which allowed for setting the current part as next. Other than that I am happy with the changes here, and think they are good for merging before.
Nothing will be functionally different, other than a cleanup of the properties, and some changes in the set-next logic to allow for setting the current part as next.

nytamin (approved):
I've read through collections/**, and it looks good!
I have some notes on about making times more distinguished (number vs Time vs Duration) but I'll probably address that later some time.

PR #322, opened 2020-09-08
chore: remove recorded files and recording ui

Julusian:
This removes the recording api and ui.

I have discussed with @nytamin and we do not believe it to be used anywhere, and removing it helps cleanup the code and will help keep the system performant as it reduces the db queries we need to do during playout and other times.

* **Other information**:

**Status**


nytamin (approved):
LGTM

PR #323, opened 2020-09-08
chore: remove old manual-playout prototype

Julusian:
This removes the manual playout prototype that populates the rundown list page with many buttons.

It was only ever a prototype that was not expanded on, and better things can be achieved with a combination of dashboards, adlib pieces, adlib actions and an empty rundown.

* **Other information**:

**Status**


nytamin (approved):
Looks good! Just `jest -u` before merging this in :)

PR #325, opened 2020-09-10
Chore/remove bucket ingest cache

nytamin:
This PR removes the no-longer-used BucketIngestCache collection.


Julusian (approved):
I dont like that we are having to create the collection to then destroy it, but I dont think we can do better easily.
The only question is whether it is worth it to destroy these empty collections

nytamin:
I mimicked what we're doing in the migrations for 0.25.0 (https://github.com/nrkno/tv-automation-server-core/blob/master/meteor/server/migration/deprecatedDatabases/0_25_0.ts#L42).
I haven't found a way to easilly drop the collection without going through meteors mongo implementation.

Julusian:
> I mimicked what we're doing in the migrations for 0.25.0

I was not aware we did that.
I'm not very opposed to this, so go ahead and merge

PR #327, opened 2020-09-15
Gui Optimizations

thomasslee97:

1. Limits tracker re-runs for segments not affected by current/next part changing.
2. Fixes issue of segments occasionally not rewinding properly.

1. Because of changes in subscribed collections, the tracker in `SegmentTimelineContainer` are running the autorun for every visible segment all the time, including running a heavy `getResolvedSegment` function.
2. Sometimes after resetting the rundown, segments get rewound but jump back to their old positions after a while.

1. Subscriptions are limited so that only the current segment and segments changing their current/next state have their tracker re-run.
2. Segments are not jumping to the old position after rewinding.

* **Other information**:

**Status**


thomasslee97:
Needs input from @jstarpl @ianshade

meteor/client/ui/RundownView/RundownTiming.tsx
```diff
@@ -251,7 +251,7 @@ export const RundownTimingProvider = withTracker<
 			}
 			if (prevProps.parts !== this.props.parts) {
 				// empty the temporary Part Instances cache
-				this.temporaryPartInstances = {}
+				this.onRefreshTimer()
```
jstarpl, line 254: ```this.temporaryPartInstances = {}```
Should not be removed. The temporaryPartInstances is another speed-up/improvement that has been implemented at NRK and should be kept.

jstarpl (changes requested):
I think there are some clashes with some other improvements done over at NRK. These optimizations should coexist with each other, so all that's necessary is to make sure to keep the original ones where they are.

meteor/client/ui/RundownView/RundownTiming.tsx
jstarpl: Both lines are now included.

meteor/client/lib/viewPort.ts
```diff
@@ -90,7 +90,7 @@ export async function scrollToPart(partId: PartId, forceScroll?: boolean, noAnim
 
 const FALLBACK_HEADER_HEIGHT = 65
 let HEADER_HEIGHT: number | undefined = undefined
-export const HEADER_MARGIN = 25
+export const HEADER_MARGIN = 15 // NRK uses: 25
```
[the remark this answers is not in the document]
jstarpl: I restored the 25 setting.

jstarpl (approved):
Look good

PR #329, opened 2020-09-16
Fix/readonly caches

nytamin:
This is an attempt to fix an issue I found in https://github.com/nrkno/tv-automation-server-core/blob/fix/readonly-caches/meteor/server/api/asRunLog.ts#L93 where we set up a cache, but never call `cache.saveAllToDatabase()`.

The simplest solution here would be to simply call `cache.saveAllToDatabase()`, but we might not want to do that without putting it in a `syncFunction`.

This PR instead adds a ReadOnly-version of the Cache (since the AsRunLog-context only uses it for reading stuff), which doesn't need to have `cache.saveAllToDatabase()`.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/DatabaseCaches.ts
+++ b/meteor/server/DatabaseCaches.ts
@@ -45,6 +45,16 @@ import { profiler } from './api/profiler'
 
 type DeferredFunction<Cache> = (cache: Cache) => void
 
+type ReadOnlyCacheInner<T> = T extends DbCacheWriteCollection<infer A, infer B>
+	? DbCacheReadCollection<A, B>
+	: T extends DbCacheReadCollection<infer A, infer B>
+	? DbCacheReadCollection<A, B>
+	: T
+type IReadOnlyCache<T extends Cache> = Omit<
+	{ [K in keyof T]: ReadOnlyCacheInner<T[K]> },
+	'defer' | 'deferAfterSave' | 'saveAllToDatabase'
+>
+
 /** This cache contains data relevant in a studio */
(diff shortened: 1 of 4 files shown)


meteor/server/DatabaseCaches.ts
```diff
@@ -279,6 +302,24 @@ export class CacheForRundownPlaylist extends CacheForStudioBase {
 		return super.deferAfterSave(fcn)
 	}
 }
+/** A read-only copy of CacheForRundownPlaylist */
+export interface ReadOnlyCacheForRundownPlaylist extends ReadOnlyCacheForStudioBase {
```
Julusian, line 306: If you wanted to have some fun with typescript instead of hand writing the ReadOnlyCacheFor* interfaces, you could do something like (stolen from my cache improvment branch :wink: )
The benefit of using a generic adapter is it will reduce the risk of error when editing the interfaces

```
export type ReadOnlyCacheInner<T> = T extends DbCacheWriteCollection<infer A, infer B>
	? DbCacheReadCollection<A, B>
	: T extends DbCacheWriteObject<infer A, infer B>
	? DbCacheReadObject<A, B>
	: T extends DbCacheWriteOptionalObject<infer A, infer B>
	? DbCacheReadObject<A, B, true>
	: T
export type ReadOnlyCache<T extends Cache> = { [K in keyof T]: ReadOnlyCacheInner<T[K]> }
```

Julusian (changes requested):
One suggestion/discussion point

PR #334, opened 2020-09-24
fix: ensure that the piece is actually playing in sourceLayerOnPartStop

jstarpl:

Before, sourceLayerOnPartStop would stop all pieces that it could find on a given source layer. The expected behavior is to stop only the currently playing pieces.

All pieces on a given source layer are stopped. As a result, pieces with effective negative length (with enable: X, and userDuration.end: Y, X > Y) would be set.

The method filters the enable.start value to only mutate pieces that should be currently playing. The filtering is done at the `ServerPlayoutAPI.sourceLayerOnPartStop` to leave the underlying `ServerPlayoutAdLibAPI.innerStopPieces` functionality unaffected.

* **Other information**:

**Status**


Julusian (changes requested):
Good spot
I think it would be better to do the check inside of `ServerPlayoutAdLibAPI.innerStopPieces` though, otherwise adlib-actions will still make a mess

It turns out tv2 also did a similar fix that didnt made it into a pr https://github.com/olzzon/tv-automation-server-core/commit/284b59ea67657911d5322008837dc189024d86a8

I am happy with their fix, so we could cherry-pick that in instead?

thomasslee97:
If you pull in that commit you will also want https://github.com/olzzon/tv-automation-server-core/commit/da0b290434e7479645f54904a6cf7991b31a6b4b

jstarpl:
> 
> 
> Good spot
> I think it would be better to do the check inside of `ServerPlayoutAdLibAPI.innerStopPieces` though, otherwise adlib-actions will still make a mess
> 
> It turns out tv2 also did a similar fix that didnt made it into a pr [olzzon@284b59e](https://github.com/olzzon/tv-automation-server-core/commit/284b59ea67657911d5322008837dc189024d86a8)
> 
> I am happy with their fix, so we could cherry-pick that in instead?

👍  Let's make a PR out of those two commits then.

jstarpl:
Replaced with: https://github.com/nrkno/tv-automation-server-core/pull/338

PR #335, opened 2020-09-25
Feat/playlist reordering ui

gundelsby:

GUI for reordering of rundowns in playlists, including the possibility to move rundowns out of or into playlists.

Playlists are presented on a single line in the rundowns overview, not listing individual rundown in a playlist. Reordering is not possible.

* **Other information**:

**Status**


Julusian (changes requested):

meteor/client/ui/RundownList/DragAndDropTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+enum RundownPlaylistUiActionTypes {
+	HANDLE_RUNDOWN_DROP = 'HANDLE_RUNDOWN_DROP',
+	/* no-op to use when a drop is handled but no further action is necessary */
+	NOOP = 'NOOP',
+}
+
+interface IRundownPlaylistUiAction {
+	type: string
+	rundownId: RundownId
+	targetPlaylistId: RundownPlaylistId
```
jstarpl, line 33: This should probably be an optional member.

jstarpl (changes requested):
Just a small note regarding client-side code, otherwise it seems OK.

meteor/lib/api/rundown.ts
```diff
@@ -27,6 +27,12 @@ export interface NewRundownAPI {
 	resyncSegment(rundownId: RundownId, segmentId: SegmentId): Promise<TriggerReloadDataResponse>
 	unsyncRundown(rundownId: RundownId): Promise<void>
 	unsyncSegment(rundownId: RundownId, segmentId: SegmentId): Promise<void>
+	moveRundown(
```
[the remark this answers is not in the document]
nytamin: Agreed, it's something we should consider if/when we do the full refactoring of all exposed methods

meteor/client/ui/RundownList/RundownPlaylistUi.tsx
```diff
@@ -374,7 +374,7 @@ export const RundownPlaylistUi = DropTarget(
 							<span className="rundown-list-item__actions"></span>
 						</header>
 						<ol className="rundown-playlist__rundowns">{rundownComponents}</ol>
-						<footer>{playbackProgressBar ? { playbackProgressBar } : null}</footer>
+						<footer>{playbackProgressBar ? playbackProgressBar : null}</footer>
```
gundelsby, line 377: 😅

meteor/client/ui/RundownList/RundownListItem.tsx
```diff
@@ -0,0 +1,278 @@
...
+				Translated<IRundownListItemProps> & IRundownDragSourceProps & IRundownDropTargetProps
+			> {
+				studio: Studio
+				showStyle: ShowStyleBase
+
+				constructor(props: Translated<IRundownListItemProps> & IRundownDragSourceProps & IRundownDropTargetProps) {
+					super(props)
+
+					this.studio = this.props.rundown.getStudio()
+					this.showStyle = this.props.rundown.getShowStyleBase()
```
jstarpl, line 211: Sometimes, when starting up, the list view crashes with this message, related to this line:
obraz

This needs to be changed, so that the component can re-render once the ShowStyleBase streams into the browser/isn't in the local collection yet.

jstarpl: Maybe the RundownListItem could use a tracker? or we could inject these two properties in the parent-level tracker?

jstarpl (changes requested):

meteor/client/ui/RundownList/RundownListItem.tsx
gundelsby: Now that we've separated the item and the view of it into two components already I guess that adding a tracker to the data component won't make it too complicated?

PR #337, opened 2020-09-25
Feat/mono repo

Julusian:

This pulls in the various integration libraries to make this a mono repo.

With the libraries being in separate repositories, it makes versioning them more complex, and it makes it obvious what versions should be used for any version of core.

This will allow us to version and maintain core and the libraries as a single unit. They will follow the same release cycle, and have matching version numbers.

We can also be clearer about what constitutes a breaking change in the libraries, as it is only breaking if it breaks the api for consumers, not what core has to implement.

* **Other information**:

### TODO
* [x] Discuss workflow with team
* [x] typedoc - we want this for at least blueprints
* [x] fix failing core version tests
* [x] codecov - ensure is ok for core portion
* [x] add npm deploy token
* [x] skip npm publish if no token? (for prs/forks)
* [x] Add readme to packages dir
* [x] Add documentation on working to main readme
* [ ] update, rename and tidy the absorbed libs

**Status**


sparkpunkd (changes requested):
Overall, this is absolutely the way to go and a great way to version inter-dependent packages. Most cool. 

I decided to take a practical approach to this review and try and get things up-and-running. There were a few bumps in the road, so I have a few requests for changes.

### Startup

The Quick-start instructions from README.md do not seem to work. From `npm start` I get ...

```
> cd ../packages && yarn build

yarn run v1.22.5
$ yarn workspaces run build

> tv-automation-sofie-blueprints-integration
$ trash dist && yarn build:main
'trash' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
```

I think this is because `trash` is not a dev dependency in the meteor package.json, even though it is for blueprints-integration.

If I manually `cd ../packages`, then `yarn build`, then back to meteor folder and just `meteor` then I'm up and running. If I `meteor npm run dev`, I get the same error as above.
A couple of questions:

* Can we make sure that following the README.md instructions is OK with the mono-repo?
* Can this process be simulated by the CI tests so it does not come back and bite us?

### Shared API between core and integration

I note that core itself does not actually make a reference to ...server-core-integration. This could be an opportunity for some shared typings? For example, file `corePeripherals.ts` in server-core-integration states that it is a copy of `meteor/lib/api/peripheralDevice.ts` in core.

### Local linking

I did a local `yarn link` from the embedded server-core-integration into media manager ... all good 👍 

### Typedoc documentation

I know there was some discussions on typedoc and how this fits with a mono-repo approach. I see an attempt to address this in `../packages/package.json`. However, I could not work out what to run to get this going as there is no script? Running `yarn docs` inside ...server-core-integration and ...blueprint-integration fails.

Julusian:
Thanks, this is exactly what I was hoping for in a review here :)

### Startup

That is strange, I have tested that myself and it does start. Maybe some quirk of a different os? Ill do some digging.

### Shared API between core and integration

I am expecting that to become a follow up task. I know Johan had some concerns about the dependencies that using server-core-integration directly would pull into core (ddp client), so I have opted to not do anything on this for now.

### Local linking

Perfect

### Typedoc documentation

Yeah, there isnt an explicit script as I didn't think anyone would run it locally. There is a ci step to generate and publish this on the master branch. Running `yarn typedoc` inside the packages folder will produce the docs folder containing both libraries. I shall add a script in case it is wanted.

sparkpunkd (approved):

sparkpunkd:
Approved due to:
* issues with Quick Start was a local meteor install corruption, not this repo
* shared API between core and core-integration package can be a topic of further discussion
* the build of the documentation for the packages is now working and has a script

PR #340, opened 2020-09-29
fix: downgraded i18n expressions to not use any fancy typescript

nytamin:
the i18n-extract-pot script does apparently only want javascript in the i18n expressions.
I got errors like Unable to parse code "{ nrcsName: this.props.firstRundown?.externalNRCSName || 'NRCS' }"

**Status**


jstarpl (approved):
Should be fine.

PR #341, opened 2020-09-29
fix: Warning, not Error on failed NTP-time update

nytamin:
Instead of always setting systemStatus to ERROR, only show it as a warning if a previous setting of time was successful.

This means that if the time-retrieval fails on initial startup, it'll show as an ERROR.
But if it fails later, it'll show as a warning.

**Status**


jstarpl (approved):
I think this looks alright. Shouldn't cause trouble.

PR #344, opened 2020-10-01
feat: Route Set UI 2

jstarpl:

This PR introduces a new UI for the Route Sets.

The UI for the Route Sets consists of a couple of buttons in the sidebar in Rundown View and a modal confirmation box for enabling/disabling.

All of the controls are combined into a panel, next to the support panel, with buttons to toggle the routes on and off.

* **Other information**:

The Studio has a new map that allows specifying properties for the exclusivity groups. Currently, it's just user-facing labels.

**Status**


nytamin:
When there are exactly 2 RouteSets in an exclusivity group, they should turn into a single selector:
image

vs

image

nytamin:
Another note:
Could you flip the toggles, so that "on" is to the right, instead of left?
(I asked Jonas to flip the design sketches as well)

image

nytamin (changes requested):
I've read through the code and code-wise it looks all good!

I put some remarks on the GUI functionality as separate comments though.

jstarpl:
> When there are exactly 2 RouteSets in an exclusivity group, they should turn into a single selector:
> image
> image

The 2 RouteSets in a single exclusivity group combined into one kicks in if both of those route-sets are set as "activate-only".

jstarpl:
> 
> 
> Another note:
> Could you flip the toggles, so that "on" is to the right, instead of left?
> (I asked Jonas to flip the design sketches as well)
> 
> image

I've flipped the labels.

nytamin (approved):
Looks good!

PR #345, opened 2020-10-01
RouteSets: allow "no layers" be routed into "some layers"

nytamin:
This PR adds some additional functionality to the routeSets.

Mainly, it allows "no layer" to be routed into "some layer". This means that a user should be able to define ALL layer mappings in routeSets, should that be desireable.

* **Other information**:

**Status**


jstarpl (approved):
Looks OK. I've changed the input/select combo to an input/datalist setup that uses web platform functionality & I think achieves the same.

PR #346, opened 2020-10-02
fix: clear the blueprint config caches when the blueprint is updated

Julusian:

Bug fix

When a new blueprint is uploaded, the config cache is not cleared. This means that the blueprint can be given a stale cache that does not match the typings it expects.

Whenever the blueprint is modified, the relevant caches are cleared too.
Additionally some other lifecycle bits are handled, such as clearing caches for removed documents, and handling some less likely property changes

* **Other information**:

**Status**


nytamin (approved):
Looks good!

PR #351, opened 2020-10-07
feat: api for blueprints to manage ab player sessions with partInstance awareness

Julusian:

Feature

Blueprints are left to manage ab playback sessions themselves.
This is not possible to do well for partInstances, as more knowledge about the partInstances needs to be known than we are exposing to the blueprints

This adds an a couple of methods available to onTimelineGenerate to help with ensuring ab-session ids are unique. 
With this api, unique session ids will be used for each instance of a part. The sessionName is used as an identifier for sessions, and multiple pieces using the same sessionName will be treated as the same session, even across parts. This lets a part continue a clip from the previous part. The exception here is that if the playhead moves backwards, then the session is split (this allows for having the same part as current and next).

* **Other information**:

Blueprints-integration changes: https://github.com/nrkno/tv-automation-sofie-blueprints-integration/pull/82
Blueprints changes: https://github.com/nrkno/tv-automation-sofie-blueprints/pull/73

**Status**


meteor/server/api/playout/timeline.ts
```diff
@@ -274,19 +283,19 @@ function getTimelineRundown(cache: CacheForRundownPlaylist, studio: Studio): Tim
...
 						})
 					})
-					if (tlGenRes.persistentState) {
-						cache.RundownPlaylists.update(playlist._id, {
-							$set: {
-								previousPersistentState: tlGenRes.persistentState,
-							},
-						})
-					}
+					console.log(`AB Session result: ${JSON.stringify(context.knownSessions)}`)
```
sparkpunkd, line 285: Not required in production.

sparkpunkd (approved):
Moving the state information about sessions into the TimelineContext in core is a good solution ... and I'm pleased that this could be done both by extending both the existing TimelineContext and preserving the AB session info in the database. :-) 

Also pleased to see some unit test coverage, including for some bad parameters. 

I'm approving this on the assumption that:
* AB playback will work for ABC and ABCD - I see no reason why it would not;
* persisting the AB session info will mean good decisions continue to be made in the event of a core restart.

Julusian:
@sparkpunkd for your assumptions:
* I havent tested this implementation, but there is no reason it shouldnt work. This code in core doesn't have any context of time or assign anything to a player, so it is completely agnostic.
* The primary aim of persisting is to be consistent with assignments during normal playback. But yes, it will work equally well around a restart

meteor/client/ui/SegmentTimeline/SegmentContextMenu.tsx
```diff
@@ -36,8 +36,8 @@ export const SegmentContextMenu = withTranslation()(
 			const timecode = this.getTimePosition()
 			const startsAt = this.getPartStartsAt()
 
-			const isCurrentPart =
-				(part && this.props.playlist && part.instance._id === this.props.playlist.currentPartInstanceId) || undefined
+			const isCurrentPart = false
+			// (part && this.props.playlist && part.instance._id === this.props.playlist.currentPartInstanceId) || undefined
```
Julusian, line 40: This line should be reverted before merging, as non-ab setups havent been resolved yet

PR #352, opened 2020-10-07
feat: prompter title & prompter maintain focus script

jstarpl:

This is a feature enhancement

If the prompter looses focus when the browser is in full screen, it may be impossible for the user to operate the prompter.

Combined with some refactoring in the GUI, the [AutoHotKey](https://www.autohotkey.com/) script in the `resources` folder, allows to force focus on the browser window. The focus is restored (if needed) every 1s.

* **Other information**:

**Status**


meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -137,11 +138,15 @@ export class PrompterViewInner extends MeteorReactComponent<Translated<IProps &
 
 		this.triggerCheckCurrentTakeMarkers()
 		this.checkScrollToCurrent()
+
+		this.setDocumentTile()
```
nytamin, line 142: typo?

resources/prompter_keep_focus.ahk
```diff
@@ -0,0 +1,11 @@
+#WinActivateForce
```
nytamin, line 1: Could you add a header to this file? Since AutoHotkey isn't used anywhere else in the project, it might be good to describe it a little better. Something like:
```
This is an AutoHotkey (https://www.autohotkey.com) script that is intended to run on a computer that displays the Sofie prompter page.
It automatically restores the focus to the Prompter-window if the focus is lost, so that the prompter controlling device (keyboard-equivalent / mouse-equivalent) continues to work.
```

meteor/client/lib/DocumentTitleProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const documentTitle = new ReactiveVar<null | string>(null)
+
+interface IProps {}
+
+interface ITrackedProps {
+	cs: ICoreSystem | undefined
+	doc: string | null
+}
+
+export const DocumentTitleProvider = translateWithTracker((props: IProps) => {
```
nytamin, line 17: This should be documented, preferably with an example of usage (`documentTitle.set('The title of the page')`).

nytamin (approved):
Functionality-wise, this looks great! A few notes on documentation, other than that good to go!

meteor/client/ui/Prompter/PrompterView.tsx
jstarpl: Good find! Fixed.

resources/prompter_keep_focus.ahk
jstarpl: You're right. I've added the header to the script.

meteor/client/lib/DocumentTitleProvider.tsx
jstarpl: I've added some docs to the documentTitle ReactiveVar & it shows up in Intellisense.

jesperstarkar (commented):
@jstarpl I have had a look and like the idea, however, I'm in no position to _review_ this.

PR #357, opened 2020-10-14
feat: show the state of infinites in all segments depending on playback

jstarpl:

This is a new feature

State of infinites across all segments is correctly shown only on page load and in current segment

The change causes all Segments to update when infinites are cropped/inserted, but should not add additional re-renders beyond that.

* **Other information**:

**Status**

### DIFF
--- a/meteor/lib/lib.ts
+++ b/meteor/lib/lib.ts
@@ -1395,6 +1395,14 @@ export function assertNever(_never: never): void {
 	// Do nothing. This is a type guard
 }
 
+/**
+ * This is a fast, shallow compare of two Sets.
+ *
+ * **Note**: This is a shallow compare, so it will return false if the objects in the arrays are identical, but not the same.
+ *
+ * @param a
+ * @param b
+ */
 export function equalSets<T extends any>(a: Set<T>, b: Set<T>): boolean {
 	if (a === b) return true
 	if (a.size !== b.size) return false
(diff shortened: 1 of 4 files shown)


meteor/lib/lib.ts
```diff
@@ -1412,3 +1412,12 @@ export function equivalentArrays<T>(a: T[], b: T[]): boolean {
 	}
 	return true
 }
+
+export function equalArrays<T>(a: T[], b: T[]): boolean {
```
nytamin, line 1416: Minor note: Since this function doesn't do deep comparisons, `T` should perhaps be limited to primitives `<T extends string | number | undefined | null | boolean>`?

It might also be good to document this function to motivate its existence (compared to the more common `_.isEqual`)

nytamin (approved):
Just a minor note, looks good!

meteor/lib/lib.ts
jstarpl: The motivation is that for non-deep comparisons, this implementation is 2x faster than _.isEqual on a worst-case scenario:

obraz

and an order of magnitude faster in a best-case scenario (length of array is different):

obraz

I think I would like to keep it as-is, and just be a shallow equal, but allow using it on arrays of objects - that's very useful for the UI when it needs to compare incoming props that may be produced by some sort of a `.map` or `.filter` function, for optimizations - in that case, the incoming array is a new array, but the objects inside are the same.

Julusian (approved):

PR #365, opened 2020-10-22
fix: PeripheralDevice name fix

nytamin:
A small feature

Devices can't modify their names, its set on first connection then won't change.

The the user hasn't modified the name, it'll be updated to the device name whenever the device reconnects.

* **Other information**:

**Status**


sparkpunkd (approved):
Slightly tilting my head to one side working out the logic here, but I think I get it has to do with the state of the stored peripheral device. As long as the comment `Only allow name changes if the name is unmodified:` is what the author intends then this good to go. 

Approving as there is no obvious harm in this code, subject to testing.

PR #369, opened 2020-10-27
fix: Remove deleted segments from ingestDataCache

thomasslee97:
Cleans up segments from `ingestDataCache` when `dataSegmentDelete` is recieved.

The cached ingest data for a segment remains after a segment is deleted from the NRCS.

The ingest data will be cleaned up.

**Status**

### DIFF
--- a/meteor/server/api/ingest/rundownInput.ts
+++ b/meteor/server/api/ingest/rundownInput.ts
@@ -131,6 +131,7 @@ import { removeEmptyPlaylists } from '../rundownPlaylist'
 import { profiler } from '../profiler'
 import { PieceInstance } from '../../../lib/collections/PieceInstances'
 import { syncPlayheadInfinitesForNextPartInstance } from '../playout/infinites'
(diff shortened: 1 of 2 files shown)


Julusian (approved):

Julusian (changes requested):
Actually, I think the change would be better placed in handleRemovedSegment() in rundownInput.ts

MOS does this remove before this point, so we only need to change the flow for other gateways

thomasslee97:
I'm not sure that is better, in `handleRemovedSegment` it would only make sense to do this operation around this line, and it feels like it would be conditional on this statement being true? https://github.com/nrkno/tv-automation-server-core/blob/9f350055865ffd27ebcc6ecda69860cf4bc605c8/meteor/server/api/ingest/rundownInput.ts#L1048
So it would make sense to bundle removing cached data in with removing the segment itself? Potentially it would be better to remove to `afterRemoveSegments`?

Julusian (approved):

PR #373, opened 2020-10-30
fix: presenter screen sources

jstarpl:

This is a bug fix.

The ClockView (used to render the presenter's screen) sometimes had difficulty figuring out which Piece in a Part to use for displaying the Icon and label.

The logic for selecting the Piece to be displayed has been refactored. Also, the components used by the ClockView have been refactored into smaller files to be more manageable.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/ClockView/ClockView.tsx
+++ b/meteor/client/ui/ClockView/ClockView.tsx
@@ -150,23 +150,46 @@ const ClockComponent = withTranslation()(
 						}
 					})
 					this.autorun(() => {
-						let playlist = RundownPlaylists.findOne(this.props.playlistId)
+						let playlist = RundownPlaylists.findOne(this.props.playlistId, {
+							fields: {
+								_id: 1,
+							},
+						})
 						if (playlist) {
 							this.subscribe(PubSub.rundowns, {
-								playlistId: this.props.playlistId,
+								playlistId: playlist._id,
 							})
 
-							const rundownIds = playlist.getRundownIDs()
+							this.autorun(() => {
+								const rundownIds = playlist!.getRundownIDs()
 
-							this.subscribe(PubSub.segments, {
-								rundownId: { $in: this.props.rundownIds },
-							})
-							this.subscribe(PubSub.parts, {
-								rundownId: { $in: this.props.rundownIds },
-							})
-							this.subscribe(PubSub.partInstances, {
-								rundownId: { $in: this.props.rundownIds },
-								reset: { $ne: true },
+								this.subscribe(PubSub.segments, {
+									rundownId: { $in: rundownIds },
+								})
+								this.subscribe(PubSub.parts, {
+									rundownId: { $in: rundownIds },
+								})
+								this.subscribe(PubSub.partInstances, {
+									rundownId: { $in: rundownIds },
+									reset: { $ne: true },
+								})
+
+								this.autorun(() => {
+									let playlist = RundownPlaylists.findOne(this.props.playlistId, {
+										fields: {
+											_id: 1,
+											currentPartInstanceId: 1,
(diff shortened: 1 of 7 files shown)


meteor/client/ui/PieceIcons/PieceName.tsx
```diff
@@ -0,0 +1,44 @@
...
+			})
+		}
+
+		render() {
+			if (this.props.sourceLayer) {
+				switch (this.props.sourceLayer.type) {
+					case SourceLayerType.GRAPHICS:
+					case SourceLayerType.LIVE_SPEAK:
+					case SourceLayerType.VT:
+						return this.props.pieceInstance.piece.name
```
gundelsby, line 38: Can't this simply be `supportedLayers.has(this.props.sourceLayer.type)` instead of the switch construct?

meteor/client/ui/ClockView/ClockView.tsx
```diff
@@ -332,10 +25,19 @@ interface IStateHeader {}
...
-			studioId: studioId,
-		})
+		const playlist = RundownPlaylists.findOne(
+			{
+				active: true,
+				studioId: studioId,
+			},
+			{
+				fields: {
+					_id: 1,
```
gundelsby, line 35: Think I'd like a comment that explains why id === 1 is always the right choice here, it's not immediately understandable to me :)

meteor/client/ui/ClockView/ClockView.tsx
```diff
@@ -332,10 +25,19 @@ interface IStateHeader {}
...
+				},
+			}
+		)
+
+		console.log(playlist)
```
gundelsby, line 40: Remove this?

meteor/client/ui/ClockView/Timediff.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
+import ClassNames from 'classnames'
+import { RundownUtils } from '../../lib/rundown'
+
+export const Timediff = class Timediff extends React.Component<{ time: number }> {
+	render() {
+		const time = -this.props.time
+		const isNegative = Math.floor(time / 1000) > 0
+		const timeString = RundownUtils.formatDiffToTimecode(time, true, false, true, false, true, '', false, true) // @todo: something happened here with negative time
+		// RundownUtils.formatDiffToTimecode(this.props.displayTimecode || 0, true, false, true, false, true, '', false, true)
```
gundelsby, line 10: What about this TODO then? Is it something that should be done, or should the commented out section be deleted?

gundelsby (changes requested):

meteor/client/ui/PieceIcons/PieceName.tsx
jstarpl: Yeah, good catch. After the refactoring this makes no sense.

meteor/client/ui/ClockView/ClockView.tsx
jstarpl: This was put there in order to limit the reactivity, but I now actually realize, that we need this to be pretty reactive, and the RundownPlaylist object isn't all that reactive.

meteor/client/ui/ClockView/ClockView.tsx
jstarpl: Good catch.

meteor/client/ui/ClockView/Timediff.tsx
jstarpl: Unfortunately, I don't know what that relates to, since it's a leftover from the code before the refactoring. As far as I know this view works the way it should, so I'm just going to remove this commented out block.

PR #381, opened 2020-11-17
Fix/unsync playlist

sparkpunkd:

Additional test

Toggling playlists as _ready on air_ in ENPS causes unpredictable behavior in R24, including playlists becoming "unsynced" and not able to properly re-sync via GUI buttons. This test walks through the steps to avoid future regressions.

Create a rundown playlist with 3 rundowns, activates the playlists, starts playing, keeps changing various data elements and checks that the rundowns stay in sync (or rather `unsync === false`).

* **Other information**:

Problem became apparent when dealing with rundown playlists for the first time in NRK national news rundowns.

**Status**

### DIFF
--- a/meteor/__mocks__/helpers/database.ts
+++ b/meteor/__mocks__/helpers/database.ts
@@ -30,6 +30,7 @@ import {
 	IBlueprintPart,
 	IBlueprintPiece,
 	TSR,
+	IBlueprintRundownDB,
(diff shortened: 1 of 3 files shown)


Julusian:
Replaced with https://github.com/nrkno/tv-automation-server-core/pull/434

Rundowns will only become unsynced when they are explicitly deleted.
Segments can also become unsynced, but also only when they are deleted.

PR #385, opened 2020-11-18
fix: possible missing properties

jstarpl:

This is a bugfix.

Some of the code in some parts of the RundownNotifier used properties which were optimized away using a `field` modifier, making them effectively dead.

The `fields` were updated. Also, there were three subscriptions to PieceInstances in RundownView that essentially did the same thing, so I combined them all into one.

* **Other information**:

**Status**


sparkpunkd (approved):
Embrace the type science! Also good to combine the subscriptions.

PR #386, opened 2020-11-18
fix: better part instance simulation

jstarpl:

This is a fix/improvement.

When a PartInstance is instantiated, it may be that it's PieceInstances take a little bit to stream into the Client. In order to avoid an empty Part, the GUI keeps a Part unupdated for a little while, until the PieceInstances stream in. There may unfortunately be a little bit of a jank at the Segment level.

The simulation of the PieceInstances is now moved to the `RundownUtils.getResolvedSegment` and it's internal functions. When using a simulation of a PartInstance's contents, a reactive dependency is registered to force the segment to be re-resolved after `X` time, though in general that's going to be probably canceled by the streamed in PieceInstances. Also, the simulation is only used if less than `X` time has passed since the partInstance has been taken/nexted, so that this doesn't go into an infinite loop.

* **Other information**:

**Status**


jstarpl:
I'm closing this PR, since it's superseded by https://github.com/nrkno/tv-automation-server-core/pull/391, which is the port of this PR into release27.

PR #388, opened 2020-11-19
feat: simplify some piece content typings (#388) [publish]

Julusian:

Refactoring

The Piece.content object was defined when sofie was young. A lot of the content types have never been used, and some of the properties on used content types have also never been used.
Additionally, the base type being an interface with `{ [key: string]: TimelineObjectCoreExt[] | number | string | boolean | object | undefined | null }` means that many usages were not being typed and so are rather brittle to changes.

This PR has done a quick review to remove the unused properties, and remove the unimplemented types.
A follow up will be needed to remove cases where the blueprints set some of these removed properties (often to `''`)

For reviewers, focus on the blueprints-integration changes, as that drove the rest of the changes.

* **Other information**:

**Status**


Julusian:
@jstarpl I have assigned you to review, as I think you will have the most to say on whether my pruning was overly agressive, or if the content changes are going to cause a lot of pain for you in the ui

packages/blueprints-integration/src/content.ts
```diff
@@ -52,67 +41,47 @@ export type SomeContent =
...
-	| LowerThirdContent
 	| LiveSpeakContent
-	| MicContent
 	| TransitionContent
 
 export interface VTContent extends BaseContent {
 	fileName: string
 	path: string
-	firstWords: string
-	lastWords: string
```
jstarpl, line 69: I think I'm against removing firstWords/lastWords in a VTContent type, and would suggest just making them optional. The fact that we don't use that at the moment doesn't, in my opinion, make it redundant. It's a useful piece of information to the user and it should be assigned by the blueprints.

packages/blueprints-integration/src/content.ts
```diff
@@ -52,67 +41,47 @@ export type SomeContent =
...
 
 export interface ScriptContent extends BaseContent {
 	firstWords: string
 	lastWords: string
 	fullScript?: string
-	sourceDuration?: number
 	lastModified?: Time | null
 }
 
-export interface GraphicsContent extends BaseContent {
```
jstarpl, line 105: I think it would be useful to maintain a generic graphics content type for graphics. Going "Nora or Bust" isn't very flexible and may hurt us in the long run. Again, the fact that we don't actually implement that interface in a great way, doesn't mean that we should remove it. Adding rudimentary support for GraphicsContent in hoverscrub would be very simple for anyone wanting to use, for example, Viz integration. I suppose, given what we know, making the `path` property optional might be a good improvement.

jstarpl:
I think it is slightly too aggressive, given that just because we don't actively use a feature doesn't mean it's not useful and can be useful later. I do agree, however, that some of these interfaces have been excessive and it makes sense to shrink them down & remove the ones that we haven't used and we don't actually know what they should be at all.

packages/blueprints-integration/src/content.ts
jesperstarkar: I second this

Julusian:
@jstarpl 
With this PR I am very much taking the attitude of that if it isnt supported by core, then it is being removed. It is misleading to a blueprint developer to see types or properties that do nothing.
Why should core be 'advertising' functionality that doesn't exist and has no plans on when it will exist?

When we want to implement one of these things that have been removed, we should be designing a new and correct interface and then adding it once core can do something with it.

packages/blueprints-integration/src/content.ts
Julusian: See the general comment below.

This one is a bit of an exception as I know there are plans to use it shortly, but would it be bad to remove the property now and then add it back when it does something?

jstarpl:
> @jstarpl
> With this PR I am very much taking the attitude of that if it isnt supported by core, then it is being removed. It is misleading to a blueprint developer to see types or properties that do nothing.
> Why should core be 'advertising' functionality that doesn't exist and has no plans on when it will exist?

Well, I see it as an API/Contract that Core just doesn't implement fully. After all, it's entirely up to Core how to handle any of these pieces of information (the API/Contract does not cover what exactly is supposed to happen when you provide that data), so if the blueprints *can* provide the information, they *should*, since they are a part of the description of the show, not a way of "triggering GUI controls in Core". If it were a method that can be called by the blueprints in core, and it was an empty function body - I'd agree that it would be a case of "advertising functionality that doesn't exist". In case of the Content/Piece/Parts/Segments, I see it more as a declarative system of describing a show. Therefore it's properties aren't tied to any specific functionality in Core: whatever is ingesting that data can handle it any way it likes. Or ignore it, if it decides to do so/doesn't handle that piece of data yet.

Julusian:
> Well, I see it as an API/Contract that Core just doesn't implement fully. After all, it's entirely up to Core how to handle any of these pieces of information (the API/Contract does not cover what exactly is supposed to happen when you provide that data), so if the blueprints _can_ provide the information, they _should_, since they are a part of the description of the show, not a way of allowing blueprints to "trigger GUI controls in Core". If it were a method that can be called by the blueprints in core, and it was an empty function body - I'd agree that it would be a case of "advertising functionality that doesn't exist". In case of the Content/Piece/Parts/Segments, I see it more as a declarative system of describing a show. Therefore it's properties aren't tied to any specific functionality in Core: whatever is ingesting that data can handle it any way it likes. Or ignore it, if it decides to do so/doesn't handle that piece of data yet.

But there is an expectation that the information should do something in some situation. Having a mix of properties that do something and do nothing makes it harder to know which ones are worth setting.
If we are keeping them because they describe the show, then we could add many more properties of useful bits of information that serve no purpose and just add bloat.

jstarpl:
> But there is an expectation that the information should do something in some situation. Having a mix of properties that do something and do nothing makes it harder to know which ones are worth setting.
> If we are keeping them because they describe the show, then we could add many more properties of useful bits of information that serve no purpose and just add bloat.

Well, I don't see it that way. Like with any metadata, the "worth" of having it set is entirely up to a particular scenario. Since I don't see the Content/Pieces/Parts/Segments as a way of having something done, I don't see the problem of having set a part of the description that is disregarded by whatever is consuming it. Is saving camera metadata on a media file "worth it" if none of the applications consuming that file read that data right now? I agree that they should be made optional, since it's not needed by the content creator (blueprint developer) nor the user agent (Sofie Core) nor the users. But I would still keep it as a part of the specification since:
* it doesn't hurt anyone if it's optional
* and any of these three actors can choose to start using it/requesting at any given point, without much hassle.

I think we've exhausted our arguments and debated it as long as useful/productive. Let's have our System Architect have a deciding say @sparkpunkd .

sparkpunkd (commented):
There is part of me reading this going _oh, that old chestnut - the backwards compatibility thing_. _Backwards compatibility_ (which always implies _forwards compatibility_) - two words that I used to toss around repeatedly/naively in meetings about interface design without realising their dire impact on implementation. You carefully design v1 of an interface, making sure that the optionality and/or n-arity of each property is tightly controlled ... plus excellent typings. Things start to fall apart in v2 and by v3 everything is optional and the typings are so wide you could drive a jumbo jet through them. All the benefits of having a well-designed interface are gone, the logic of the interface ends up in a Word document (or long comments) and any new implementation is forced to include a complex web of logic that understands v1, v2 and v3. (e.g. see 1000+ pages of MXF specification ... 4 different types of weak reference - ahhh!) There is a reason why we don't use XML schema / WS-* any more ...

Enough of the backstory .... what's the pragmatic solution? Its probably case-by-case but guided by some architectural principles:

- **loose coupling** - try not to do anything that might introduce a tight coupling in the future
- **keep it simple (stupid)** - don't use every trick in the book - keep it tight so the static typings are useful
- **expect change** - design minimal and tight and expect to refactor often - use strict writers and forgiving readers - chasing a change through the code with modern IDEs is not too bad
- **test early, test often** - use tests as specification - define what you expect to happen - then implement them in the first instance by throwing "Not Implemented".  

So, broadly, I agree with Julian on the optional properties ... don't add them unlress you need them. In my experience, optional properties are far from free - you make a future decision to use that property in a required way, does that then have to be enforced in logic rather than typings?

In terms of contracts and interfaces, don't rely on the contract to be the specification - rely on the tests - run tests to find out what actually works. (This is how AS-11 MXF files are now specified in the UK - not by file structure but by tests.)

jstarpl:
_Roma locuta, causa finita_ I raise no further objections.

PR #389, opened 2020-11-20
fix: ignore "ResizeObserver loop limit exceeded" errors

nytamin:
This PR adds an ignore for the "ResizeObserver loop limit exceeded" error, to reduce noise in the logs.

The error is emitted for me when use `hover` feature of `rc-tooltip`. No issues in GUI are observed.

According to https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded the error is benign and can be ignored.

**Status**


jstarpl (approved):
Looks good! I was thinking that maybe we could avoid it somehow within rc-tooltip, but it seems that it's not a thrown exception, but rather something printed out to console.error. I just switched it to a more generic system so that we can add more ignored patterns if we ever need to.

PR #390, opened 2020-11-22
feat(Prompter): joycon

jesperstarkar:
Adds support for Nintendo™ Joycon gamepad devices for prompter controls. 

New feature added. Using the HTML Gamepad API we allows for connection of the Left, the Right or the Left + Right Joycons. The Y-axes of the joystic acts as the prompter speed controller, with speedMap and range properties to map deadband and speed response. A spline function is also used for interpolation.

* **Other information**:

**Status**


jesperstarkar:
@jstarpl current status 2020-12-08 is that i consider the lifetime management + scrolling action to be "production ready". Need some more tests on connectiity + the combinations of L, R and L+R. I'm not going into theory about what happens if other devices are connected, and I think the string-match takes care of that for us nevertheless.

I think this part of the task is ready for a quick review.


Next step will be calculating button presses.

Improvements might be displaying connection info++ to the screens.

meteor/client/ui/Prompter/controller/joycon-device.ts
```diff
@@ -0,0 +1,228 @@
...
+				if (oldBtn === newBtn) continue
+
+				// if () { // press
+
+				// } else if () { // release 
+
+				// }
+			}
+		}
+		console.log('')
```
jstarpl, line 161: I suppose this can be cleaned-up?

meteor/client/ui/Prompter/controller/joycon-device.ts
```diff
@@ -0,0 +1,228 @@
...
+				// We tried to move, but haven't
+				this.lastSpeed = 0
+
+				// @todo: haptic feedback
+			}
+		}
+		this.currentPosition = scrollPosition
+
+		// debug
+		this.prompterView.DEBUG_controllerState({
```
jstarpl, line 216: Doing this on every frame might impose some penalty. The `DEBUG_controllerState` is supposed to be very fast, but still, I think it may be best if we somehow limit this to only when `lastSpeed` or `lastInputValue` has changed?

jstarpl (approved):
Looks good, just a couple of comments.

jesperstarkar:
@jstarpl Since last:
* Refactoring, a bit more logical flow, less spaghetti. Yes, I'm still mixing functions with returns and assignments to properties on the object some places, but for good reason.
* Minimized parsing, loops and getGamepads() to a minimum.
* Added buttons handling
* Learned more about hos the joycon mutates to/from the single/dual modes. Discovered a problem, probably in the joycon itself, that we can't work around in code as it is.

I have not added any throttling, neither reduced the calls to debug. Feel free to change if you want.

jesperstarkar:
Ducomentation draft: https://app.gitbook.com/@sofie/s/sofie-tv-automation/~/diff/drafts/-MODFYeE9MhCNDSJGluz/documentation/features-and-configuration/prompter

jstarpl (approved):
OK, let's get this merged in.

PR #393, opened 2020-11-25
Troubleshoot unsynced issue

nytamin:
This PR improves some logging to aid in troubleshooting an unsynced rundown issue.

### DIFF
--- a/meteor/server/api/ingest/rundownInput.ts
+++ b/meteor/server/api/ingest/rundownInput.ts
@@ -1743,14 +1743,6 @@ export function isUpdateAllowed(
 						allowed = false
 					}
 				})
-				if (!allowed) {
-					// For troubleshooting; also log which parts was inserted:
-					if (partChanges.inserted) {
-						_.each(partChanges.inserted, (part) => {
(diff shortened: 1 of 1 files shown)


meteor/server/api/ingest/rundownInput.ts
```diff
@@ -1743,6 +1743,14 @@ export function isUpdateAllowed(
 						allowed = false
 					}
 				})
+				if (!allowed) {
```
sparkpunkd, line 1746: Not sure about the value of this. Is repeated in the `printChanges(partChanges)` below. Although adding the external ID in print changes would be helpful?

nytamin: Ah, yeah that's true. I didn't scroll that far down.. This is unnecessary then.

sparkpunkd: Also no harm in this.

sparkpunkd (approved):
Will be extremely helpful in tying back to what happened in the NRCS system.

PR #395, opened 2020-11-27
Fix/reload rundown

sparkpunkd:

A fix for issues of unsync/resync getting in a loop.

Following a mosReplace message during an active and playing show, a rundown becomes unsynced and then will no resync when a producer repeatedly presses the resync button.

Pressing the re-sync should be able to _calm the waters_ during an active show.

* **Other information**:

Initial commit is only improved code documentation and variable names. 

**Status**

### DIFF
--- a/meteor/package-lock.json
+++ b/meteor/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "automation-core",
-	"version": "1.15.0-in-testing-R27.2",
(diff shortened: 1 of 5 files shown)


Julusian (approved):
While I am not 100% happy, this makes the treatment of the dynamicParts the same as when doing segment ingest updates. Also the problem will go away with the task to allow PartInstances to not have a backing Part, meaning these troublesome dynamic parts will no longer exist

PR #397, opened 2020-11-27
feat: show inspector on collapsed layers

jstarpl:

A change of UX

When an output group is collapsed, hoverscrub is disabled.

The hoverscrub should always be available, regardless of collapsed state. Also, when an output group is collapsed and it only has a single source layer present, the items in that layer will no longer span the entire height of the output group, but instead show up using only half of the height.

obraz

* **Other information**:

**Status**


sparkpunkd (approved):
Looks good.

PR #398, opened 2020-11-27
fix: Unsynced Segment and updating Next

ianshade:
This PR restores the correct behavior regarding the Unsynced Segment feature (enabled with the `allowUnsyncedSegments` setting), and updating the next Part under certain conditions.

1. With `allowUnsyncedSegments` enabled, updating or removing a segment when it is not allowed (currently playing etc.) leads to unsyncing the Rundown instead of just the Segment.
2. The `UpdateNext.afterInsertParts`, essential to finding the next Part if there is none while new Parts are added, is called only when ingesting through the the mosDevice. 
3. `handleRemovedSegment` doesn't lead to finding a new next Part in case current next was removed.
4. Resyncing a Segment doesn't lead to removing it in case the ingest device reports that the Segment does not exist.

1. With `allowUnsyncedSegments` enabled, all not allowed Segment operations lead to unsyncing that Segment, instead of the Rundown.
2. `UpdateNext.afterInsertParts` is called in afterIngestChangedData, in case Parts were inserted.
3. `handleRemovedSegment` calls `UpdateNext.ensureNextPartIsValid` if the Segment was removed.
4. When resyncing, a Segment is removed if the device reports that it does not exist.

* **Other information**:

**Status**


meteor/server/api/ingest/rundownInput.ts
```diff
@@ -1143,17 +1142,22 @@ export function updateSegmentsFromIngestData(
 	studio: Studio,
 	playlist: RundownPlaylist,
 	rundown: Rundown,
-	ingestSegments: IngestSegment[]
+	ingestSegments: IngestSegment[],
+	removedPreviousParts?: boolean
```
Julusian, line 1146: Could you rename this and some of the existing verisons of it to  `removedPartsBeforeInserted` or something a bit more descriptive. Even though its my code I had to follow it through to understand it..

meteor/server/api/ingest/rundownInput.ts
```diff
@@ -1170,9 +1174,8 @@ function updateSegmentFromIngestData(
 	playlist: RundownPlaylist,
 	rundown: Rundown,
 	ingestSegment: IngestSegment
-): SegmentId | null {
+): { segmentId: SegmentId | null; newPartIds: string[] } {
```
Julusian, line 1177: It would be better to call these `insertedPartExternalIds`, as they are externalIds

Julusian (changes requested):
Looks ok, but would be good ro clarify some variable/parameter naming.
I only marked one point in each, but please follow them around a bit and update the names there too.

meteor/server/api/ingest/rundownInput.ts
ianshade: Renamed as requested.

meteor/server/api/ingest/rundownInput.ts
ianshade: Renamed as requested.

PR #400, opened 2020-12-02
fix: evaluate transition properties when leaving part

baltedewit:
A transition on a part can delay pieces inside that part. When leaving the part the overlap with the next part is calculated, during said calculation we need to consider the transition properties such that the timing of the pieces will remain correct. In order to do this without looking at the previous previous part we should save these transition properties on the part instances.

This PR changes:
 * Save the transition properties on the part instance during a take
 * Consider the transition properties during transformation from the _previous_ part to timeline

### DIFF
--- a/meteor/server/api/playout/take.ts
+++ b/meteor/server/api/playout/take.ts
@@ -196,10 +196,7 @@ export function takeNextPartInnerSync(
 			'timings.take': now,
 			'timings.playOffset': timeOffset || 0,
 			// set transition properties to what will be used to generate timeline later:
-			'transProps.allowed': currentPartInstance ? !currentPartInstance.part.disableOutTransition : false,
-			'transProps.preroll': partInstance.part.prerollDuration,
-			'transProps.transitionPreroll': partInstance.part.transitionPrerollDuration,
-			'transProps.transitionKeepalive': partInstance.part.transitionKeepaliveDuration,
+			allowedToUseTransition: currentPartInstance && !currentPartInstance.part.disableOutTransition,
 		},
 		$unset: {} as { string: 0 | 1 },
 	}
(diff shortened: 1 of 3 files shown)


meteor/server/api/playout/take.ts
```diff
@@ -195,6 +195,11 @@ export function takeNextPartInnerSync(
 			isTaken: true,
 			'timings.take': now,
 			'timings.playOffset': timeOffset || 0,
+			// set transition properties to what will be used to generate timeline later:
+			'transProps.allowed': currentPartInstance ? !currentPartInstance.part.disableOutTransition : false,
```
Julusian, line 199: I think it would be good to do this set as a `transProps: literal<TransformTransitionProps>({` instead of each property individually, to help ensure any future changes to the type gets done.

It would be a good idea for the generation of this object to be pulled into a reusable function (also used in the other places we generate this type) to help ensure it is always done the same

nytamin (approved):

Julusian:
@baltedewit I went to quickly resolve my comment, and then noticed that both https://github.com/nrkno/tv-automation-server-core/pull/400/files#diff-a6ab1806565e12da53c451b7d616b9b81531826f66a93b9738fb265a55c6231bL626-L630 and https://github.com/nrkno/tv-automation-server-core/pull/400/files#diff-a6ab1806565e12da53c451b7d616b9b81531826f66a93b9738fb265a55c6231bL673-L679 were taking the numbers from themselves, not the part before.
It was just the `allowed` property that was based on the part before, so now only that one property is preserved

baltedewit:
> both [timeline.ts#L626-L630](https://github.com/nrkno/tv-automation-server-core/pull/400/files#diff-a6ab1806565e12da53c451b7d616b9b81531826f66a93b9738fb265a55c6231bL626-L630) and [timeline.ts#L673-L679](https://github.com/nrkno/tv-automation-server-core/pull/400/files#diff-a6ab1806565e12da53c451b7d616b9b81531826f66a93b9738fb265a55c6231bL673-L679) were taking the numbers from themselves, not the part before.

Ooh, yeah, of course they are. I didn't even bother to look at that.

PR #402, opened 2020-12-02
fix: return clientResponseError when normal operations instead of throwing in disableNextPiece

jstarpl:

This is a bugfix

When no Piece can be found to be disabled/undo-disabled, the method will throw a Meteor.Error

Instead of throwing an Error in a situation that can just regularly happen during normal operation, the method shall now return a `ClientResponseError` object and the GUI will check for that object and error code and show a custom message with just "Warning" level of criticality, instead of showing this as a Critical Error level of notification.

* **Other information**:

This PR also adds some tests for the ClientAPI functions to check that it doesn't mistake `Meteor.Errors` for `ClientResponseErrors`.

**Status**

### DIFF
--- a/meteor/client/ui/RundownView.tsx
+++ b/meteor/client/ui/RundownView.tsx
@@ -672,12 +672,34 @@ const RundownHeader = withTranslation()(
 			this.takeRundownSnapshot(e)
 		}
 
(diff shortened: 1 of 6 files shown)


sparkpunkd (approved):
Code changes seem fine and I'm looking forward to less errors all around!

PR #404, opened 2020-12-07
feat: AdLib hover scrub

jstarpl:

This is a new feature.

There is no preview available for AdLibs in the shelf.

There is now a hover-scrub preview available for the L3rd and Clip-type (STK/VT) in the AdLib shelf:

obraz

* **Other information**:

**Status**


meteor/client/ui/FloatingInspectors/MicFloatingInspector.tsx
```diff
@@ -0,0 +1,73 @@
...
+const SCRIPT_PART_LENGTH = 250
+
+export function MicFloatingInspector(props: IProps) {
+	const { t } = useTranslation()
+
+	let startOfScript = (props.content && props.content.fullScript) || ''
+	let cutLength = startOfScript.length
+	if (startOfScript.length > SCRIPT_PART_LENGTH) {
+		startOfScript = startOfScript.substring(0, startOfScript.substr(0, SCRIPT_PART_LENGTH).lastIndexOf(' '))
+		cutLength = startOfScript.length
```
gundelsby, line 26: This is the same assignment as the initial assigment, isn't it?

meteor/client/ui/FloatingInspectors/SplitsFloatingInspector.tsx
```diff
@@ -0,0 +1,68 @@
...
+								background: item.role === SplitRole.ART,
+								box: item.role === SplitRole.BOX,
+							},
+							{
+								second: array.length > 1 && index > 0 && item.type === array[index - 1].type,
+							}
+						)}
+						key={item._id + '-preview'}
+						style={{
+							left: ((item.content && item.content.x) * 100).toString() + '%',
```
gundelsby, line 36: What happens with these properties if item.content or item.content.x is undefined?

gundelsby (approved):
A couple of nitpicks, but overall this looks good. Well done on doing the refactor job too 👌

meteor/client/lib/StyledTimecode.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Time, formatDurationAsTimecode } from '../../lib/lib'
+
+interface IProps {
+	time: Time
+}
+
+const FRAMES_INLINE_STYLE: React.CSSProperties = {
```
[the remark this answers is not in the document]
jstarpl: Yeah, I specifically want this not to be dependent on the stylesheet, so that it's not overridden by any cascading styles. I realize it's not very pretty, but it ensures that I don't have to check if it actually works everywhere and modify the parent component's styling accordingly. I can just drop the StyledTimecode anywhere and it just works. I think it's an acceptable trade-off.

meteor/client/ui/FloatingInspectors/L3rdFloatingInspector.tsx
```diff
@@ -0,0 +1,155 @@
...
+					return {
+						key: key,
+						value: str,
+					}
+				}
+			})
+		) as Array<KeyValue>
+	}
+
+	let changed: Time | undefined = undefined
```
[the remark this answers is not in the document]
jstarpl: Yeah, that's a very nice optimization! I really need to refresh my operators.

meteor/client/ui/FloatingInspectors/MicFloatingInspector.tsx
jstarpl: Well, no, because at that stage, the `startOfScript.length` will be different, because `startOfScript` has been reassigned.

meteor/client/ui/FloatingInspectors/SplitsFloatingInspector.tsx
jstarpl: It's never going to be undefined, since if no geometry is provided, a default two-box setup will be used - but I agree that since typings allow for it to be undefined, it should be handled! And yeah, I suppose that's going to end up being "NaN%" - so it's very good feedback. Thanks again for reminding me about nullish coalescing operator! It allows for setting this up with a reasonable fallback. I was able to tighten the typings for `SplitSubItem.content` as well, so thanks for making me take a second look at this! 👍

meteor/client/ui/FloatingInspectors/MicFloatingInspector.tsx
gundelsby: Yes, of course :)

PR #405, opened 2020-12-07
feat: remove blueprints minimumCoreVersion as the version of blueprin…

Julusian:
…ts-integration provides the same constraint in a more maintainable manner

Fix

There is a check to ensure that the `minimumCoreVersion` property exposed by the bluerpints is compatible with the current version of core. This is no longer necessary, and not very useful as the nrk blueprints have forgotten to update the field (still set to 1.11.0 which is not correct)

blueprints-integration version is released in parallel to core, so the numbers used in core will line up perfectly.
There is already a check to ensure that the blueprints are using roughly that (`~` check)
Meaning we require the blueprints to use an integration from the same release branch.

With this, the minimumCoreVersion doesnt add anything, as the blueprints-integration version check provides both a min and a max.

* **Other information**:
The interesting part of the change is 
https://github.com/nrkno/tv-automation-server-core/compare/release28...chore/remove-blueprint-minimumCoreVersion?expand=1#diff-d2c08dcd0c0f9d69b6a571c6d5a9f52dc0d58f04b3ce41832b2ea1914310a52aR287-R293

**Status**


sparkpunkd (approved):
Another good reason for a mono repo.

PR #406, opened 2020-12-08
feat: display show style variant in lobby

jstarpl:

This is a new feature

The new lobby only displays the Show Style (`ShowStyleBase`) name. 

Because of how the ShowStyles are most commonly used, this does not provide enough granularity, since in most installations, there is a single ShowStyleBase, with multiple variants, for ease of configuration. This PR adds the ShowStyleVariant name before the ShowStyleBase name, if the two are different.

obraz

* **Other information**:

It may be worthwhile to consider making sure that the ShowStyleBase is configured to some reasonable name - other than "Default Show Style".

**Status**


meteor/client/ui/RundownList/RundownListItem.tsx
```diff
@@ -264,6 +273,16 @@ export const RundownListItem = translateWithTracker<IRundownListItemProps, {}, I
 					// rundown ids can start with digits, which is illegal for HTML id attributes
 					const htmlElementId = `${HTML_ID_PREFIX}${unprotectString(rundown._id)}`
 
+					const showStyleLabel =
+						this.props.showStyleVariant &&
+						this.props.showStyleBase &&
+						this.props.showStyleVariant.name !== this.props.showStyleBase.name
+							? t('{{showStyleVariant}} - {{showStyleBase}}', {
```
gundelsby, line 280: I'd argue that running this through the translator is unnecessary, given that there's no content that actually can be translated :)

gundelsby (approved):
LGTM 👍

meteor/client/ui/RundownList/RundownListItem.tsx
jstarpl: I'm running these kinds of strings through the translator system to be able to support non-European languages - in some languages the separator should be other than an en-dash - Japanese would use a "〜" and Hebrew would use a 0x05be character (an en-dash, but in a different position relative to the baseline). I'm not religious about it, but I try to use that system whenever possible - for consistency - when concatenating UI strings.

PR #407, opened 2020-12-09
Feat: Local source layer type

thomasslee97:

This adds the "LOCAL" source layer type, which can be used to represent sources that are not controlled by Sofie but are available and controlled within the gallery e.g. delayed playback (EVS in this case), telestrator etc. This creates a visual distinction for sources that are often treated like cameras / live inputs but don't fit into either of those categories and may be controlled by an operator within the gallery. Below is a screenshot of what the current colour chosen for this layer looks like.

image

* **Other information**:

**Status**


jstarpl (approved):
Looks good!

PR #410, opened 2020-12-10
feat: Add CasparCG cron job system setting

ianshade:
This PR adds the following feature: A Core System setting to disable the cron job restarting CasparCG servers.

The cron job that restarts CasparCG servers is running every night and there is no way to disable it.

A section called _Cron jobs_, with a setting called _Disable CasparCG restart job_ is added to Core System Settings. The setting is off by default. When the setting is on, CasparCG servers will not be restarted during the execution of the nightly cron jobs.

* **Other information**:

**Status**


meteor/lib/collections/CoreSystem.ts
```diff
@@ -89,6 +89,13 @@ export interface ICoreSystem {
 		transactionSampleRate?: number
 	}
 	enableMonitorBlockedThread?: boolean
+
+	/** Cron jobs running nightly */
+	cron?: {
+		casparCG?: {
```
Julusian, line 95: it would be clearer to call this `casparCGRestart`

ianshade: I renamed it to `casparCGRestart`

ianshade:
I refactored the setting to say _Enable CasparCG restart job_. It will be on by default. A migration was added.

sparkpunkd:
This is relevant to NRK RK5/6 workflows so I would like to get this included soon.

PR #411, opened 2020-12-10
fix: block the updateTimeline from triggerUpdateTimelineAfterIngestDa…

Julusian:
…ta firing before onPartPlaybackStarted has been executed for a partInstance

This is because if it executed in this window, then playout-gateway and core will be out of sync in timings and cause a weird loop

Bug fix

If the updateTimeline fires, then core and playout-gateway will be misaligned and constantly flicking between two start times for the partInstance

This 'fixes' it by ignoring the updateTimeline and assumes it will be run shortly by the onPartPlaybackStarted callback that is expected.

* **Other information**:

I would class this as a bit of a hack, as it is ignoring some edge conditions such as if onPartPlaybackStarted fails to do an updateTimeline for some reason, or if the updateTimeline is triggered in some other way (perhaps an adlib?)

But this catches the common case affecting rk7 that is most likely to happen due to their volume of mos updates.

I have not tested it at all, as I do not have a sufficient mos setup to do so.

**Status**

### DIFF
--- a/meteor/server/api/peripheralDevice.ts
+++ b/meteor/server/api/peripheralDevice.ts
@@ -268,6 +268,9 @@ export namespace ServerPeripheralDeviceAPI {
 					}
 				})
 
+				// TODO - we should do the same for the partInstance.
+				// Or we should we not update the now for them at all? as we should be getting the onPartPlaybackStarted immediately after
+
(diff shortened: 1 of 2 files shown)


nytamin:
The hack looks good as a preliminary hack. We've tested it and it doesn't seem to break anything (is har to test if it actually solves the issue, due to the hard-to-catch race condition).
Merging this into release27, will follow up with a more permanent solution later.

PR #417, opened 2021-01-04
feat: PartInstances without Parts

Julusian:

Feature

It is not possible to have a PartInstance without a matching Part. This means that when inserted a queued/dynamicallyInserted PartInstance, we insert a temporary Part too.
This makes the data model/flow a mess, as we have to do a lot of cleanup checks.

We can have PartInstances without matching Parts. They get labelled as orphaned, and get positioned by comparing and updating their rank relative to other Parts before and after the ingest update.

They can then follow the usual flow for PartInstances being reset.

* **Other information**:

The commit history should be squashed when merged.

**Status**


Julusian:
@jstarpl could you have a look over this and think about if the ui will be happy?
I did some changes so that the PartInstances were displayed, but I am unsure if the timing calculations or other bits will be working/correct

meteor/server/api/playout/adlib.ts
```diff
@@ -402,22 +409,21 @@ export namespace ServerPlayoutAdLibAPI {
...
+			true
+		)
+		newPartInstance.part._rank = getRank(
+			currentPartInstance.part,
+			followingPart?.part?.segmentId === newPartInstance.segmentId ? followingPart?.part : undefined
+		)
 
 		cache.PartInstances.insert(newPartInstance)
-		// TODO-PartInstance - pending new data flow
-		cache.Parts.insert(newPartInstance.part)
```
sparkpunkd, line 420: New data flow? Just checking this TODO does not need to be done?

Julusian: The 'new data flow' it was referring to was this PR :)

sparkpunkd (approved):
These changes, overall, seem to make the code a simpler to understand and cleaner. I note good test coverage that gives me confidence that these changes should do what is expected.

jstarpl:
> 
> 
> @jstarpl could you have a look over this and think about if the ui will be happy?
> I did some changes so that the PartInstances were displayed, but I am unsure if the timing calculations or other bits will be working/correct

Yeah, I did a quick read-through and it doesn't seem like the timing calculations will know about these orphaned Part Instances. I think the effect will be that the UI will crash or - at the very least - will be unable to count down to the end of an orphaned Part. I think the timing components need to understand the reason for being orphaned ('adlib-part' vs 'deleted', etc.) to be able to do any useful calculations, so they need to natively understand these orphaned Part Instances. Does this allow one to set the current part as next as well?

meteor/client/lib/rundown.ts
```diff
@@ -219,6 +221,68 @@ export namespace RundownUtils {
...
+				for (const part of segmentParts) partInstanceMap.set(part._id, wrapPartToTemporaryInstance(part))
+				for (const partInstance of segmentPartInstances) {
+					if (
+						partInstance._id === playlist.nextPartInstanceId ||
+						partInstance._id === playlist.previousPartInstanceId
+					) {
+						const currentValue = partInstanceMap.get(partInstance.part._id)
+						if (currentValue?._id === playlist.currentPartInstanceId) {
+							// Show the current instead of the next or previous PartInstance if they are the same
+							continue
```
jstarpl, line 269: I don't exactly understand this part? 🤔 Why would the Part Instance (of the same ID) be listed multiple times in `segmentPartInstances`? Wouldn't there be a single instance of this Part Instance? How can there be more then one? Shouldn't this check if the `partId` of these Part Instances (current/next/previous) (and not their IDs) is the same, and then _show the current instead of the next or previous PartInstance if they are the same_?

Julusian: Its not that the same PartInstanceId will be listed multiple times, but multiple PartInstances with the same PartId. This was not directly relevant for this task, but is needed to allow having the same part in both current and next.

I have simplified this block as it was being a bit overly verbose, which hopefully makes it a bit easier to follow. The map is `Map<PartId, PartInstance>()`, so it was always checking PartId, and doing some additional checks for the special case

sparkpunkd (approved):
Good catch that this calculation is needed.

Interesting switch to use a Map. I looked into _why don't we use more maps for Sofie_ - because I used them heavily in Java/Scala times - but it turns out that removing items from maps in Javascript can be much slower than deleting from an object. (Still the case?) But you are using `clear` here so this is not an issue.

Julusian:
> Good catch that this calculation is needed.
> 
> Interesting switch to use a Map. I looked into _why don't we use more maps for Sofie_ - because I used them heavily in Java/Scala times - but it turns out that removing items from maps in Javascript can be much slower than deleting from an object. (Still the case?) But you are using `clear` here so this is not an issue.

That is interesting about map performance, not something I have considered.
I'm not too sure why we haven't, but I am slowly moving bits over when refactoring. The only thing to be careful of is that they wont convert to and from json properly, so have to be avoided in types which go over ddp/mongo

jstarpl:
> 
> 
> Good catch that this calculation is needed.
> 
> Interesting switch to use a Map. I looked into _why don't we use more maps for Sofie_ - because I used them heavily in Java/Scala times - but it turns out that removing items from maps in Javascript can be much slower than deleting from an object. (Still the case?) But you are using `clear` here so this is not an issue.

I think the reason we didn't use Maps, and instead used objects was that the initial Meteor 1.6 which we used when we started had a pretty lousy support for ES6.

jstarpl:
Okay, I think/hope 🤞 that the `RundownTimingProvider` does something sensible with the timing. I also think I found a bug with how members of `displayDurationGroups` were handled, so hopefully that will work correctly/more as one would expect (?).

The task needs to include thorough testing of this behavior, but I think it's alright.

sparkpunkd (approved):
Nothing jumps out and the comments make clear what the code is intended to do. I look forward to see this come together in the release.

PR #418, opened 2021-01-06
fix: prompter generic gamepad

jesperstarkar:

Better parsing of gamepad names, to match a broader set of indicators.

Only matches a specific ID, which turns out to not be the only possible outcome when connecting.

Matches a broader and more common part of the ID, which will catch both variants that we have experienced.

* **Other information**:

**Status**


tsorbo (approved):
Has been tested and confirmed in XPRO

sparkpunkd (approved):
Good spot. Good to go!

PR #421, opened 2021-01-06
feat: special timeline class for when the first part hasn't been taken

baltedewit:
Adds a timeline class that is added to the timeline when the rundown is active but no part has been taken yet. This allows blueprints developer to add special behavior for this state.

**Status**

### DIFF
--- a/packages/blueprints-integration/src/timeline.ts
+++ b/packages/blueprints-integration/src/timeline.ts
@@ -8,7 +8,7 @@ export enum TimelineObjClassesCore {
 	RundownRehearsal = 'rundown_rehersal',
 	RundownActive = 'rundown_active',
 	BeforeFirstPart = 'before_first_part',
-	IsLastPart = 'last_part',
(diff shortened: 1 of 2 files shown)


packages/blueprints-integration/src/timeline.ts
```diff
@@ -4,6 +4,13 @@ export { TSR }
 
 export { Timeline } from 'timeline-state-resolver-types'
 
+export enum TimelineObjClassesCore {
+	RundownRehearsal = 'rundown_rehersal',
+	RundownActive = 'rundown_active',
+	BeforeFirstPart = 'before_first_part',
+	IsLastPart = 'last_part',
```
Julusian, line 11: I would go with 'NoNextPart' instead of 'IsLastPart' as it makes it clearer that there could technically be more parts, but they arent playable. In general this will mean the end of the playlist, but not always

Julusian (approved):

PR #422, opened 2021-01-08
Fix: syncfunction emit warning when waiting

nytamin:
This PR adds a check in the syncFunction execution which checks how long it has waited for other syncFunctions to complete before starting executing.
If the wait is too long, a message with the functions it waited on is logged as a warning.

* **Other information**:

**Status**


Julusian (approved):
I like it. I am curious to see how much this flags up

PR #424, opened 2021-01-12
fix: Reset partInstances for removed parts

ianshade:
This PR improves performance by preventing too many partInstances from being fetched by the GUI and into the cache

When parts that were played back are removed, their partInstances stay in the database and are still fetched into the cache and by the GUI. In a workflow where one rundown is reused, and segments are inserted and removed multiple times, it causes the number of processed partInstances to unnecessarily grow over time and slow down server and client.

1. When removing a part, mark its partInstances as reset.
2. Fetch only non-reset partInstances into the cache.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/rundown.ts
+++ b/meteor/server/api/rundown.ts
@@ -459,6 +459,8 @@ export function afterRemoveParts(cache: CacheForRundownPlaylist, rundownId: Rund
 		startPartId: { $in: removedPartIds },
 	})
 
+	cache.PartInstances.update({ 'part._id': { $in: _.map(removedParts, (p) => p._id) } }, { $set: { reset: true } })
+
 	cache.deferAfterSave(() => {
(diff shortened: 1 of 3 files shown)


meteor/server/api/rundown.ts
```diff
@@ -459,6 +459,8 @@ export function afterRemoveParts(cache: CacheForRundownPlaylist, rundownId: Rund
 		startPartId: { $in: removedPartIds },
 	})
 
+	cache.PartInstances.update({ 'part._id': { $in: _.map(removedParts, (p) => p._id) } }, { $set: { reset: true } })
```
Julusian, line 462: Please use removedPartIds instead of mapping over removedParts again.

ianshade: Done

Julusian (approved):

PR #425, opened 2021-01-12
feat: add getPartInstanceForPreviousPiece method to ActionExecutionContext

Julusian:
Allows for getting the part metadata for a piece found via findLastPieceOnLayer

feature

During an adlib-action, it is not possible to get the PartInstance for a PieceInstance returned from `findLastPieceOnLayer`. For the nrk case, this is a problem as any 'primary' piece simply states it was primary in its metadata, and the part metadata should be referenced for the properties needed to regenerate it.

This means it is not possible to regenerate the piece as a new piece for the current part (ie, a lazy generated adlib piece) without copying the same properties into the pieceMetadata causing the piece type to get even heavier

It is possible to get this metadata.

* **Other information**:

Unit tests need to be written before this can be merged

**Status**


Julusian:
@sparkpunkd I am not 100% sure about the api here. I think this makes it easy and avoids needing to expose any additional types or fields
But perhaps we should expose partInstanceId in the result of `findLastPieceOnLayer` and make that be the parameter here? My concern there is that it will allow for more arbitrary PartInstance access.

Once an api is agreed I shall do unit tests

sparkpunkd (approved):
Code looks fine ... but PR is still in draft?

PR #427, opened 2021-01-14
feat: strict mongo typings

Julusian:

code quality

The mongo typings are very loose, so allow for easy mistakes to be made. such as renamed properties not being updated, or deleted properties being referenced

This tightens the typings using new typescript 4.1 features. The interfaces are a little brittle (and it has been a fight to get them to work), but will benefit us greatly in code quality.

* **Other information**:

**Status**


Julusian:
I am closing this as the work here is likely not going to be very useful going forward
Mongo client 4.2 includes the type of strong typing I was trying to write here, so the same can be achieved by using those strong mongo typings instead

PR #431, opened 2021-01-21
fix: findLastPieceOnLayer returning stale data

Julusian:

bug fix

When using the `findLastPieceOnLayer` functionality via either adlib-actions or sticky sourcelayer shortcuts, it can find a pieceInstance from a previous activation of the rundown. This could return wildly unexpected results, especially in cases where the rundown has significantly changed since it was last activated.

`RundownPlaylist.active` has been replaced with `RundownPlaylist.activationId`, and is set to a random value each time the rundown is activated or reset.
Each PartInstance and PieceInstance generated will have a `playlistActionId` property set to match the RundownPlaylist.
The `findLastPieceOnLayer` logic now only looks for PieceInstances where `playlistActionId` the current value of `RundownPlaylist.activationId`

* **Other information**:

This has resulted in a large amount of changes to update all references of `RundownPlaylist.active`

I did originally add activationId as a second property, but it felt cleaner to have a single property describing the activation state and 'session'

**Status**


Julusian:
@sparkpunkd do you agree with my decision to replace the active property? I am a little concerned about something breaking due to the very weak mongo typings, but anything should be caught during testing.

I shall try running it through my prototype (and only semi-working) stricter mongo typings to see if that flags anything

sparkpunkd:
> @sparkpunkd do you agree with my decision to replace the active property? I am a little concerned about something breaking due to the very weak mongo typings, but anything should be caught during testing.
> 
> I shall try running it through my prototype (and only semi-working) stricter mongo typings to see if that flags anything

I think that cure of the use facing issue is paramount and worth this risk. After all, you are looking at a way of improving the typings in the future and you have updated the tests. We should flag to be careful to check for side effects in release testing. So I'm good with this.

sparkpunkd (approved):
Looks carefully worked through with test coverage.

PR #433, opened 2021-01-25
Fix: looping playlists

baltedewit:
This PR fixes the behaviour for looping rundown playlists. All that was missing was selecting the next part. Besides that it adds some markers in the UI to clarify that the current rundown playlist will loop.

**Status**


Julusian (approved):

jstarpl (approved):
I think it would be good to explain what `length` parameter does, since it's used a bit down in the code and I didn't quite understand it required a couple of passes for me to figure out why the stop condition for the loop can be injected there. Otherwise, very clean.

PR #434, opened 2021-01-25
feat: allow playing part to be removed

Julusian:

feature

the currently playing part cannot be removed from the rundown.
attempts to do so, will cause the rundown to become unsynced, blocking other ingest updates until it is resynced.

The rundown is only 'unsynced' when it is deleted, and its content will be retained.
All part and segment updates are accepted and applied. Any removed segments will be preserved if they are on air. Any removed parts will be removed, but the partinstances will remain.

A special case is being left in place to allow for deleted segments to become 'unsynced' like the rundown is. This is not being well tested, and is disabled by default

* **Other information**:

**Status**


sparkpunkd (approved):
Surprisingly simple change - and well articulated with readable code and the term "orphaned". Nothing stands out as being an issue.

Someone needs to really get busy with ENPS when doing the release test on this one, while checking both the UI and playout the the right thing! However, as this is based on the same concepts as adlibs that use part instances, I'm optomistic ...

PR #435, opened 2021-01-26
feat: Add PieceCountdownPanel

ianshade:
This PR adds new feature: a dashboard panel that counts down to the end of the content of a live VT-like piece.

The panel will display a countdown to the end of the content of a live piece. 
The piece has belongs to one of the selected sourceLayers and carry `content` with `sourceDuration`, which is used to calculate the remaining time. 
When past `sourceDuration`, timer changes to the late-color and count up.
If there are no live pieces matching criteria, the timer stays at 00:00.

* **Other information**:

**Status**


meteor/client/ui/Shelf/PieceCountdownPanel.tsx
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	updateTimecode(e: TimingEvent) {
+		let timecode = 0
+		if (this.props.livePieceInstance && this.props.livePieceInstance.startedPlayback) {
+			const vtContent = this.props.livePieceInstance.piece.content as VTContent | undefined
+			const sourceDuration = vtContent?.sourceDuration || 0
+			const startedPlayback = this.props.livePieceInstance.startedPlayback
+			if (startedPlayback && sourceDuration > 0) {
+				timecode = e.detail.currentTime - (startedPlayback + sourceDuration)
```
jstarpl, line 60: This also needs to take into account `vtContent?.seek` for seeked clips.

meteor/client/ui/Shelf/PieceCountdownPanel.tsx
```diff
@@ -0,0 +1,120 @@
...
+		)
+	}
+}
+
+export const PieceCountdownPanel = withTracker<IPieceCountdownPanelProps, IState, IPieceCountdownPanelTrackedProps>(
+	(props: IPieceCountdownPanelProps & IPieceCountdownPanelTrackedProps) => {
+		const unfinishedPieces = getUnfinishedPieceInstancesReactive(props.playlist.currentPartInstanceId)
+		const livePieceInstance: PieceInstance | undefined =
+			props.panel.sourceLayerIds && props.panel.sourceLayerIds.length
+				? _.find(_.flatten(_.values(unfinishedPieces)), (piece: PieceInstance) => {
```
jstarpl, line 108: I think we are trying to limit our reliance on underscore, this could easily be replaced with `Object.values().flat().find()` with better performance at type security.

meteor/lib/collections/RundownLayouts.ts
```diff
@@ -108,6 +114,13 @@ export interface DashboardLayoutAdLibRegion extends RundownLayoutAdLibRegion {
 	height: number
 }
 
+export interface DashboardLayoutPartCountdown extends RundownLayoutPieceCountdown {
+	x: number
+	y: number
+	width: number
+	height: number
```
jstarpl, line 121: `width` & `height` don't seem to do much. Perhaps they could be used as an input to scale the font-size of the text in the panel? Otherwise it feels like they can be removed, since the panel is transparent anyway and just contains a single piece of text?

jstarpl (changes requested):
The key aspect is to take `.seek` into account for accurate countdowns.

meteor/client/ui/Shelf/PieceCountdownPanel.tsx
ianshade: `Array.prototype.flat()` is an es2019 feature and looks like we're targeting es2018

jstarpl: Oh, it is & we are? I didn't realize that. Well, the question then is if it makes sense to just upgrade to es2019 (I think with Node 12 we can), but I guess the easiest option is to use `_.flatten`, but replace `_.values` and `_.find` with native alternatives for this PR.

Julusian: https://node.green/#ES2019 yeah we should be good to target es2019
Im pretty sure meteor is doing babel transpiling anyway, so maybe we could even switch that to target esnext

meteor/lib/collections/RundownLayouts.ts
ianshade: I changed it so that the text is centered horizontally, within panel's `width`, and I replaced `height` with `scale`, which scales the font size.

meteor/client/ui/Shelf/PieceCountdownPanel.tsx
ianshade: I replaced `values` and `find`. Shall moving to es2019 be made in another PR, perhaps with refactoring any other occurrences of `_.flatten` etc.?

meteor/client/ui/Shelf/PieceCountdownPanel.tsx
ianshade: Done

jstarpl (approved):

PR #436, opened 2021-01-26
Show restart button only for device notifications >= WARNING_MAJOR

thomasslee97:
The "Restart Device" button will only be shown where a device has set its status to WARNING_MAJOR or anything more severe.

The "Restart Device" button is shown in notifications for all status messages that do not have the status code "GOOD".

The "Restart Device" button will not be shown in a device status notification unless the status is >= WARNING_MAJOR. This allows WARNING_MINOR to be used to send device status updates to the user without making the user think that some intervention is required. An example use case is sending a notification to indicate that Viz graphics are loading; in this case there is no intervention required as it is just an indicator of the progress of the Viz device to let the user know that it may not be safe to take graphics on air yet.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/RundownView/RundownNotifier.tsx
+++ b/meteor/client/ui/RundownView/RundownNotifier.tsx
@@ -310,7 +310,7 @@ class RundownViewNotifier extends WithManagedTracker {
 							'Devices',
 							getCurrentTime(),
 							true,
-							parent && parent.connected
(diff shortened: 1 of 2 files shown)


jstarpl:
I think it would be good to document this in `packages/server-core-integration/src/lib/corePeripherals.ts`. This new behavior is generally in line with the comments for `StatusCode`, but I think it would be good to explicitly state that an option to restart the peripheral device will be given to the user for statuses >= `WARNING_MAJOR`.

thomasslee97:
I have added a comment to the documentation for `WARNING_MAJOR`

PR #437, opened 2021-01-26
feat: segmentRanksUpdate peripheral device method

thomasslee97:
Adds the peripheral device function "segmentRanksUpdate" which allows the order of segments to be modified by a peripheral device (e.g. ingest gateway) by passing a list of segment external Ids and a list of ranks.

To perform this operation now, the dataSegmentUpdate method would have to be called with all of the segment data. This sends more data over the wire and may lead to the segment being unsynced or blueprints being rerun.

The function "segmentRanksUpdate" allows for a lightweight method of updating just the ranks of segments.

* **Other information**:
This primarily provides benefits for the spreadsheet and iNews gateways, as the MOS ingest has its own way of handling updates only containing changes to a segment rank.

**Status**


meteor/lib/api/peripheralDevice.ts
```diff
@@ -122,6 +122,13 @@ export interface NewPeripheralDeviceAPI {
 		rundownExternalId: string,
 		ingestSegment: IngestSegment
 	): Promise<void>
+	dataSegmentRanksUpdate(
+		deviceId: PeripheralDeviceId,
+		deviceToken: string,
+		rundownExternalId: string,
```
Julusian, line 128: I don't like this api signature having two arrays that need to line up.
It would make more sense to have it as
```
segmentRanks: { [segmentExternalId]: number }
```

meteor/server/api/ingest/__tests__/ingest.test.ts
```diff
@@ -927,6 +927,318 @@ describe('Test ingest actions for rundowns and segments', () => {
 
 	// TODO Part tests are minimal/happy path only on the assumption the API gets little use
 
+	testInFiber('dataRundownUpdate remove dynamicInserted Part', () => {
```
Julusian, line 930: It looks like you have reverted the deletion of a couple of unit tests here

meteor/server/api/ingest/rundownInput.ts
```diff
@@ -1494,6 +1509,48 @@ function afterIngestChangedData(
...
+		playlistId,
+		RundownSyncFunctionPriority.INGEST,
+		'handleUpdatedSegmentRanks',
+		() => {
+			const rundown = getRundown(rundownId, rundownExternalId)
+			const playlist = getRundownPlaylist(rundown)
+			const cache = waitForPromise(initCacheForRundownPlaylist(playlist))
+
+			for (const [index, externalId] of segmentExternalIds.entries()) {
+				cache.Segments.update(
```
Julusian, line 1536: I wonder if there should be a check for the segments existing, and to log a warning if they do not?
I think the update call returns a changed count that can be used for this

Julusian (changes requested):

meteor/server/api/ingest/__tests__/ingest.test.ts
thomasslee97: I have remove these extra tests

meteor/lib/api/peripheralDevice.ts
thomasslee97: I've made this change

meteor/server/api/ingest/rundownInput.ts
thomasslee97: I have added a check to see if the number of changes is 0 after each update operation, if so then a warning is logged with the segmentExternalId and rundownExternalId similar to other ingest warnings

Julusian (approved):

PR #439, opened 2021-01-27
Better activation ux

baltedewit:
This PR contains the following changes:

- Fix: Clarify all uses of "activation" to contain either "rehearsal" or "on-air" such that it is clear to users what the system will do
- Feature: Add an option to directly activate to on-air mode when the planned start of the rundown is close or has passed
- Feature: When the users does a take on an inactive rundown we ask the user if they want to activate (the assumption here is that they think it is already active)

### DIFF
--- a/meteor/client/ui/RundownView.tsx
+++ b/meteor/client/ui/RundownView.tsx
@@ -188,7 +188,7 @@ const WarningDisplay = withTranslation()(
 						}>
 						<p>
 							{t(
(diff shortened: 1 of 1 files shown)


jstarpl (changes requested):
The capitalization of the "On-Air" text varies, I think if we're cleaning this up, it would be good to make it consistent. I propose "On-Air" since that attracts the eye better than "on-air" or "On-air".

jstarpl (approved):

PR #440, opened 2021-01-27
fix: resync missing rundown

jstarpl:

This PR changes the behavior of how re-syncing a missing rundown is handled and some related UI changes.

If a Rundown/Segment is unsynced in Sofie, when user resyncs it, the Unsynced flag will be taken off and a request to the Ingest device will be fired off. If the device returns a "MISSING" state, the Rundown/Segment is left in the "fake synced" state and a message is shown through the notification center for the user to make a decision what to do about it, but only when doing "re-syncing" from the Rundown View. When doing it from the lobby, no message is shown.

If the device returns a "MISSING" state, the Rundown/Segment's "unsynced" flag is turned back on. Also some of the UI handling of missing Rundowns was reworked so that these notifications are also shown when resyncing a missing Rundown in the Lobby.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/RundownList/util.ts
+++ b/meteor/client/ui/RundownList/util.ts
@@ -8,6 +8,7 @@ import { doModalDialog } from '../../lib/ModalDialog'
 import { doUserAction, UserAction } from '../../lib/userAction'
 import { MeteorCall } from '../../../lib/api/methods'
 import { TFunction } from 'i18next'
+import { handleRundownReloadResponse } from '../RundownView'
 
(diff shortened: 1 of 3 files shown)


Julusian (approved):
I didn't think about the unsync flag being left unset, so good spot.
I am changing all the stuff around unsync in my task, but the rest of this will still be good :)

jstarpl:
I would still appreciate if @nytamin took a look, since parts of the code I changed were his and I may have missed some nuances.

PR #443, opened 2021-02-01
fix: blueprintConfig dropped when restoring a snapshot

ianshade:
Fixes a bug causing the Blueprint Configuration on studios and showStyles to be lost when restoring a system snapshot.

When restoring a system snapshot, Blueprint Configuration on the studios and showStyles is dropped entirely instead of being restored from the snapshot. The cause is using `migrateConfigToBlueprintConfigOnObject`, which assumes that `blueprintConfig` is missing, when restoring snapshots.

Blueprint Configuration is restored correctly. `migrateConfigToBlueprintConfigOnObject` handles both cases - when `blueprintConfig` is present and when it's missing.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/migration/1_12_0.ts
+++ b/meteor/server/migration/1_12_0.ts
@@ -251,7 +251,7 @@ export function migrateConfigToBlueprintConfigOnObject<
 		blueprintConfig?: IBlueprintConfig
 	}
 >(document: DBInterface): DBInterface {
-	document.blueprintConfig = {}
+	document.blueprintConfig = document.blueprintConfig || {}
 	// @ts-ignore old typing
 	const oldConfig = document.config as any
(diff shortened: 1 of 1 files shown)


Julusian:
Im not keen on the change here, what happens if the doc somehow has both blueprintConfig and config defined? It will end up with a weird mash of the two, in a not very predictable manner.

I think it would be better to either ignore config if blueprintConfig is already defined, or if config is defined then replace blueprintConfig. I dont have strong feelings on which would be correct

ianshade:
I agree, and I think if both existed, it's better to ignore `config` because it seems more likely that it would be a leftover and `blueprintConfig` will have newer data.

Julusian (approved):

PR #444, opened 2021-02-01
Feat: layer mappings manifest

baltedewit:
This PR adds a manifest for layer mappings such that the definition of the UI is tied to the features available in the gateway. It currently only takes the manifest from the first playout-gateway in the studio and disables editing of mappings and routesets when no manifest is available. I understand that this is suboptimal but I believe it will cover 99% of current use cases.

**Status**


nytamin:
Just a comment from the side line; @baltedewit I think it would be good to add your explanation above as a comment close to the code picking the first peripheralDevice, for doc and posterity 🙂

meteor/client/ui/Settings/components/ConfigManifestEntryComponent.tsx
```diff
@@ -75,6 +84,7 @@ export const ConfigManifestEntryComponent = withTranslation()(
 					<label className="field">
 						{t(configField.name)}
```
jstarpl, line 85: Is there any localization effort included here, to provide translations for these (i.e. the manifest provides it's own locale tables)? If not, then - since these will now be inserted from the manifest, I don't think there's a point in passing these through the localization engine. There's no chance of any of these strings ending up inside the Core localization.

meteor/client/ui/Settings/StudioSettings.tsx
```diff
@@ -683,9 +434,7 @@ const StudioMappings = withTranslation()(
 										<span>
 											{t('Port: {{port}}, Channel: {{channel}}', { port: mapping.portId, channel: mapping.channelId })}
 										</span>
-									)) || (
-										<span>{t('Unknown device type: {{device}}', { device: TSR.DeviceType[mapping.device] })} </span>
-									)}
+									)) || <span>-</span>}
```
jstarpl, line 437: Shouldn't the summaries come from the translations as well, given that the configuration parameters for the devices are now completely abstracted away to the device to be configured within the manifest? In that case it doesn't seem to make sense to keep this tight binding here, methinks? This could simply use i18next's interpolation even, with "unescaping" turned off.

jstarpl (changes requested):
Looks good overall, but I think that if we are offloading the device definition UI description to the peripheralDevice, it would make sense to offload the "summary" definition there as well, since that summary still creates a tight binding between core and device configuration interface.

meteor/client/ui/Settings/components/ConfigManifestEntryComponent.tsx
baltedewit: Yeap, agreed and removed the calls to i18n

meteor/client/ui/Settings/StudioSettings.tsx
baltedewit: I've done a low effort implementation of summaries composed from the manifest, translation is redundant as the manifest doesn't carry any localization.

jstarpl: I think that's a reasonably good, functional effort. Anything above is sugar that probably needs some more thought anyway.

jstarpl (approved):
Yeah, looks good!

PR #445, opened 2021-02-01
feat: allow playing part to be removed

Julusian:

feature

reverted and reopening #434

the currently playing part cannot be removed from the rundown.
attempts to do so, will cause the rundown to become unsynced, blocking other ingest updates until it is resynced.

The rundown is only 'unsynced' when it is deleted, and its content will be retained.
All part and segment updates are accepted and applied. Any removed segments will be preserved if they are on air. Any removed parts will be removed, but the partinstances will remain.

A special case is being left in place to allow for deleted segments to become 'unsynced' like the rundown is. This is not being well tested, and is disabled by default

* **Other information**:

**Status**


Julusian:
I realised I had forgotten that there is some ui work needed before this can be merged which is unlikely to be done before release30 is frozen, so have reverted and reopened this

Julusian:
@jstarpl Ive requested a review from you to look over just the last commit [`e91d3b3` (#445)](https://github.com/nrkno/tv-automation-server-core/pull/445/commits/e91d3b3709283bdc754c865a24433ade79b86ece) . The rest of the ui changes are simple, but this is messing with notes and could have negative effects I overlooked on reactivity

jstarpl (approved):
I think this looks OK.

PR #447, opened 2021-02-03
fix: Restore shelf layout selection to lobby

thomasslee97:
Re-adds the "shelf layout" selection to the lobby (see https://github.com/nrkno/tv-automation-server-core/pull/256 for details).

image

Cannot select shelf layout from the lobby.

Re-adds the dropdown to select shelf layout. If no shelf layouts have been marked as selectable options (the default behaviour) then the text "Default" will be displayed in this column (screenshot below). There is a limitation to this implementation, as it will only work for playlists containing a single rundown for now.

image

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/RundownList.tsx
+++ b/meteor/client/ui/RundownList.tsx
@@ -297,6 +297,7 @@ export const RundownList = translateWithTracker(() => {
 											<span>{t('On Air Start Time')}</span>
 											<span>{t('Duration')}</span>
 											<span>{t('Last updated')}</span>
+											<span>{t('Shelf Layout')}</span>
 											<span>&nbsp;</span>
 										</header>
 										{this.renderRundownPlaylists(rundownPlaylists)}
--- a/meteor/client/ui/RundownList.tsx
+++ b/meteor/client/ui/RundownList.tsx
@@ -297,7 +297,9 @@ export const RundownList = translateWithTracker(() => {
 											<span>{t('On Air Start Time')}</span>
 											<span>{t('Duration')}</span>
 											<span>{t('Last updated')}</span>
-											<span>{t('Shelf Layout')}</span>
+											{this.props.rundownLayouts.some((l) => l.exposeAsShelf || l.exposeAsStandalone) && (
+												<span>{t('Shelf Layout')}</span>
+											)}
 											<span>&nbsp;</span>
 										</header>
 										{this.renderRundownPlaylists(rundownPlaylists)}
--- a/meteor/client/ui/RundownList/RundownListItemView.tsx
+++ b/meteor/client/ui/RundownList/RundownListItemView.tsx
@@ -11,6 +11,7 @@ import { iconDragHandle, iconRemove, iconResync } from './icons'
 import JonasFormattedTime from './JonasFormattedTime'
 import { EyeIcon } from '../../lib/ui/icons/rundownList'
 import { LoopingIcon } from '../../lib/ui/icons/looping'
+import { RundownShelfLayoutSelection } from './RundownShelfLayoutSelection'
 
 interface IRundownListItemViewProps {
 	isActive: boolean
@@ -109,6 +110,9 @@ export default withTranslation()(function RundownListItemView(props: Translated<
 			</span>
 			<span className="rundown-list-item__text">
 				<JonasFormattedTime timestamp={rundown.modified} t={t} />
+			</span>
+			<span className="rundown-list-item__text">
+				<RundownShelfLayoutSelection rundown={props.rundown} />
 			</span>
 			<span className="rundown-list-item__actions">
 				{confirmReSyncRundownHandler ? (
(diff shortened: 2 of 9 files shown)


meteor/client/ui/RundownList.tsx
```diff
@@ -297,6 +297,7 @@ export const RundownList = translateWithTracker(() => {
 											<span>{t('On Air Start Time')}</span>
 											<span>{t('Duration')}</span>
 											<span>{t('Last updated')}</span>
+											<span>{t('Shelf Layout')}</span>
```
jstarpl, line 300: Given that at this stage, the component already knows if there are possibly any Rundown Layouts exposed to be used for the button, I'd like this column to be conditional, so that it doesn't show up if `rundownLayouts.length === 0`.

meteor/client/ui/RundownList/RundownShelfLayoutSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+export const RundownShelfLayoutSelection = translateWithTracker<
+	IRundownShelfLayoutSelectionProps,
+	IRundownShelfLayoutSelectionState,
+	IRundownShelfLayoutSelectionTrackedProps
+>((props: IRundownShelfLayoutSelectionProps) => {
+	const rundownLayouts = RundownLayouts.find({ showStyleBaseId: props.rundown.showStyleBaseId }).fetch()
+
+	return {
+		...props,
+		rundownLayouts,
```
jstarpl, line 34: 1. RundownLayouts are already passed as a prop into RundownPlaylistUi (RundownList.tsx:254). I don't see a point in registering a tracker here and a tracker for the RundownLayouts there. Considering the general structure of all of the RundownList compoentents, I would suggest removing this tracker and relying on the list of RundownLayouts passed down through props from the main tracker.
2. You don't have to (and generally shouldn't) return all of the incoming props back into a `translateWithTracker`. The HOC will merge whatever you return from this tracker function with incoming props and inject that into the child component.

meteor/client/ui/RundownList/RundownShelfLayoutSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+			const shelfLayouts = this.props.rundownLayouts
+				.filter((layout) => layout.exposeAsShelf)
+				.map((layout) => {
+					return this.renderLinkItem(
+						layout,
+						getRundownWithLayoutLink(this.props.rundown.playlistId, layout._id),
+						`shelf${layout._id}`
+					)
+				})
+			const allElements = [
```
jstarpl, line 93: This doesn't really need to be an array of elements (and will actually work better if it isn't). You can just replace this with a fragment, remove the `key={'separatorX'}` bits and have `{...standaloneLayouts}` in the fragment, where actual arrays/lists are.

meteor/client/ui/RundownList/RundownListItemView.tsx
```diff
@@ -110,6 +111,9 @@ export default withTranslation()(function RundownListItemView(props: Translated<
 			<span className="rundown-list-item__text">
 				<JonasFormattedTime timestamp={rundown.modified} t={t} />
 			</span>
+			<span className="rundown-list-item__text">
```
jstarpl, line 114: I would very much like this to be conditional on if there are any RundownShelfLayoutSelections to be displayed. Some users don't use this feature at all and will just see a wall of "Default's" for no reason.

meteor/client/ui/RundownList.tsx
```diff
@@ -297,6 +297,7 @@ export const RundownList = translateWithTracker(() => {
...
 											<span>&nbsp;</span>
 										</header>
 										{this.renderRundownPlaylists(rundownPlaylists)}
```
jstarpl, line 303: I would also like to see the support for this picker in "real" playlists (so ones that have 2+ rundowns in them). I don't see a reason for having a feature that has a "hole" in it like that, given that multi-rundown playlists are a first class citizen in Sofie.

jstarpl (approved):
I think it would be good to make this feature conditional on if there actually is something to show & also make sure that it uses the data the RundownList tracker has already fetched. It would also be good to make sure that multi-rundown Playlists are supported as well.

jstarpl (changes requested):
I think it would be good to make this feature conditional on if there actually is something to show & also make sure that it uses the data the RundownList tracker has already fetched. It would also be good to make sure that multi-rundown Playlists are supported as well.

thomasslee97:
@jstarpl I have made the changes you requested including hiding the dropdown unless there are layouts available to select from, and supporting multiple-rundown playlists.

meteor/client/ui/RundownList/RundownListItemView.tsx
```diff
@@ -111,9 +114,15 @@ export default withTranslation()(function RundownListItemView(props: Translated<
 			<span className="rundown-list-item__text">
 				<JonasFormattedTime timestamp={rundown.modified} t={t} />
 			</span>
-			<span className="rundown-list-item__text">
-				<RundownShelfLayoutSelection rundown={props.rundown} />
-			</span>
+			{rundownLayouts.some((l) => l.exposeAsShelf || l.exposeAsStandalone) && (
```
jstarpl, line 117: Ooh, I've never heard of `.some()`! Thanks!

giphy

jstarpl (approved):

PR #448, opened 2021-02-03
fix: Don't use NTP servers in tests

thomasslee97:
Prevents the use of NTP servers in test environments.

A number of tests use the mock meteor module, which calls the Meteor.startup registered functions when it mocks startup. This means that a number of tests make calls to an NTP server, which can lead to the tests failing if these request get blocked due to rate limiting.

Calls to NTP servers are disabled if the environment variable JEST_WORKER_ID is set, indicating that the code is being run in the test environment. The time diff is faked in this scenario.

* **Other information**:

**Status**


Julusian (approved):

Julusian:
Looks fine to me.
I have been hitting issues with ntp timeouts locally on one particular test, which is now commented out because github actions did too.

There is a proposed task to remove this ntp syncing logic, but until then this is good

PR #449, opened 2021-02-04
feat: translate segment ids on mos rundown ingest

Julusian:

bug fix

If a mos segment name changes as part of a resync operation, this will cause the segment to be treated as a new segment.
This will cause it to lose the PartInstances (as-played state), and if it is currently playing will cause it to become unsynced

The same logic used for determining segmentId changes on smaller mos operations is applied, so that segments are preserved as much as possible.

* **Other information**:

**Status**


sparkpunkd (approved):
I read and understood the new tests and - on the assumption that these are passing OK - this is a carefully identified and executed fix.

PR #450, opened 2021-02-05
chore: declarationMap

nytamin:
This PR adds the property `declarationMap` to the build-tsconfig of the packages.
This enables VSCode to jump to the actual source files instead of the compiled declaration files. This increases productivity by at least 2.3%.

**Status**

### DIFF
--- a/packages/blueprints-integration/tsconfig.build.json
+++ b/packages/blueprints-integration/tsconfig.build.json
@@ -10,6 +10,7 @@
 			"@sofie-automation/blueprints-integration": ["./src/index.ts"]
 		},
 		"resolveJsonModule": true,
+		"declarationMap": true,
 		"types": ["node"]
 	}
 }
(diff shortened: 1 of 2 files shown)


Julusian:
I think this is a good idea.
It would be better to enable it https://github.com/nrkno/tv-automation-sofie-code-standard-preset/blob/master/ts/tsconfig.lib.json instead, so that we get this for all of our libraries.

The only negative is see here is the increase in npm package size (340k vs 252k), but I dont think that matters

nytamin:
Good point! We could also look into ignoring those files when publishing the npm-package. I suppose if the original source isn't included in the npm-package anyway, it won't make sense to include any map files either.

nytamin:
I've pushed the change to https://github.com/nrkno/tv-automation-sofie-code-standard-preset/blob/master/ts/tsconfig.lib.json

PR #451, opened 2021-02-06
Chore: linting fixes

nytamin:
This PR fixes most of the issues listed in SonarLint for Release 31 (and a few issues in the master branch)


meteor/client/lib/reactiveData/reactiveDataHelper.ts
```diff
@@ -66,12 +65,11 @@ export function memoizedIsolatedAutorun<T extends (...args: any) => any>(
...
 	let result: ReturnType<T>
 	const fId = hashFncAndParams(functionName, params)
-	const parentComputation = Tracker.currentComputation
 	if (isolatedAutorunsMem[fId] === undefined) {
 		const dep = new Tracker.Dependency()
 		dep.depend()
-		const computation = Tracker.nonreactive(() => {
-			const computation = Tracker.autorun(() => {
+		const computationNonreactive = Tracker.nonreactive(() => {
+			const computationAutorun = Tracker.autorun(() => {
```
jstarpl, line 72: So this is actually something I have a problem with. This `computationAutorun` and `computationNonreactive` are exactly the same thing. `Tracker.nonreactive` simply passes through, whatever is returned by the inner function. When written like this, it invites the thinking that these are two different computations and may cause someone to, for example, register two `onStop` or `onInvalidate` handlers. Line 94 becomes especially confusing.

jstarpl (changes requested):
I think one instance of variable name shadowing should be left intact, because it actually makes it easier to understand the code, without understanding how certain reactivity control functions work in Meteor. There are some cases where variables are now renamed to single letters, which I think we've agreed weren't a good thing - but I don't have as strong a feeling about them, and I personally feel that they are just fine in short iterators, so I don't have strong objections.

meteor/client/lib/reactiveData/reactiveDataHelper.ts
nytamin: Good point! I'll revert that change and mark it as wontfix in SonarCloud.

PR #454, opened 2021-02-09
fix: add missing contexts to blueprint api methods [publish]

Julusian:

Fix

Ensures every blueprint method has a context parameter. Some got lost during conflicts in https://github.com/nrkno/tv-automation-server-core/pull/287

* **Other information**:

**Status**


gundelsby (approved):
The code changes are fine, but the contexts that were removed were removed because they weren't used for anything (according to ESlint). So if they actually are necessary that means that our method signatures don't reflect that, which might be an issue we should look into

Julusian:
This is a 'public' api, so they should be there because they are useful, not only if they are used by our current blueprint implementation.
Additionally, the only way to log from the blueprints is via the `context.logWarning` and similar methods, so not having them impairs debugging when writing implementations of those methods

PR #459, opened 2021-02-12
fix: invalidReason translation

jstarpl:

This is a bugfix.

The `invalidReason` property can not be translated using the blueprint i18n feature.

The `invalidReason` field on a Part is redesigned to make it possible to translate the message. Also, the property is simplified, since the previous structure of `title` and `description` wasn't really used anyway.

* **Other information**:

### DIFF
--- a/meteor/lib/collections/Parts.ts
+++ b/meteor/lib/collections/Parts.ts
@@ -65,8 +65,10 @@ export class Part implements DBPart {
 	public displayDuration?: number
 	public invalid?: boolean
 	public invalidReason?: {
-		title: string
-		description?: string
+		message: {
+			key: string
+			args?: { [key: string]: any }
+		}
 		color?: string
 	}
 	public floated?: boolean
@@ -133,11 +135,7 @@ export class Part implements DBPart {
 			? [
 					{
 						type: NoteType.ERROR,
-						message: {
-							key: `${this.invalidReason.title}${
-								this.invalidReason.description ? `: ${+this.invalidReason.description}` : ''
-							}`,
-						},
+						message: this.invalidReason.message,
 						origin: {
 							name: this.title,
 						},
--- a/meteor/lib/collections/Parts.ts
+++ b/meteor/lib/collections/Parts.ts
@@ -19,6 +19,7 @@ import { createMongoCollection } from './lib'
 import { Studio } from './Studios'
 import { ShowStyleBase } from './ShowStyleBases'
 import { registerIndex } from '../database'
+import { ITranslatableMessage } from '../api/TranslatableMessage'
 
 /** A string, identifying a Part */
 export type PartId = ProtectedString<'PartId'>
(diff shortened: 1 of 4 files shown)


meteor/lib/collections/Parts.ts
```diff
@@ -133,11 +135,7 @@ export class Part implements DBPart {
 			? [
 					{
 						type: NoteType.ERROR,
-						message: {
-							key: `${this.invalidReason.title}${
-								this.invalidReason.description ? `: ${+this.invalidReason.description}` : ''
-							}`,
-						},
+						message: this.invalidReason.message,
```
Julusian, line 138: I expect this will need a namespace to be defined

packages/blueprints-integration/src/rundown.ts
```diff
@@ -137,20 +137,26 @@ export interface IBlueprintPart<TMetadata = unknown> extends IBlueprintMutatable
...
+	 * 			args?: { [key: string]: any }
+	 * 		},
 	 * 		color?: string
 	 * 	}}
 	 * @memberof IBlueprintPart
 	 */
 	invalidReason?: {
-		title: string
-		description?: string
+		message: {
```
Julusian, line 156: I would prefer the `message` type here to be made its own exported interface, so that inside the blueprints it can be referenced more cleanly. Currently there is a lot of `IBlueprintPart['invalidReason']` and I expect to want to reference the type of message here too

Julusian (changes requested):

packages/blueprints-integration/src/rundown.ts
gundelsby: Maybe we should just make a type for parametrized strings then? Basically what is here with a text property and optional args as a key/value object?

jstarpl: That's a good idea, though I think it's going to be used in invalidReason only, I think? In other places it seems to be that the translated messages are sent through functions, and there it is just more conveniant to write warn('something'), rather than wrapping it in an object. @Julusian what do you mean by "type of message"?

Julusian: It is currently only used here, but I do want to be able to easily reference that type in the blueprints. 
I meant the `IBlueprintPart['invalidReason']['message']` type.

My motivation for this is to change the blueprints to have the `t` function:
```
export function t(key: string, args: { [key: string}: any }): ITranslatableMessage {
  return { key, args }
}
```

gundelsby: If that type now happens, I'd like to propose that we don't use "key" as a property name for the message text string. It's really only a key in the context of i18n translations, in all other contexts it's the actual message text :)

jstarpl: Well, it's still technically, the `key` to the translations bundle. We're using the optional, and very conveniant way of "English text is the key", but it's not exactly neccessary to use it that way.

meteor/lib/collections/Parts.ts
jstarpl: Core should now add the namespace for this message on ingest in 30e1ff1a21114b88c2477b2872b74eee3af79256

packages/blueprints-integration/src/rundown.ts
jstarpl: Ok, how about now? The blueprint integrations now have their own translatable message interface. I still think it makes sense to use `key` here, since that's what it technically is.

Julusian (approved):

gundelsby:
We already had one, with the same name. So the naming should be changed or the interfaces merged, I guess. Note that I believe there is a principal difference between a parametrized string and a "translatable message".

https://github.com/nrkno/tv-automation-server-core/blob/8868a0c0d36df6a31346ac6d299744c888aa77d3/meteor/lib/api/TranslatableMessage.ts#L6

PR #460, opened 2021-02-13
chore: fix check licenses script

nytamin:
The checkLicenses.js script hanged/froze, due to the Legally library having changed how it was outputting the licenses.
So i fixed it.


Julusian:
Instead of trying to fix our use of legally, I think it would be better to import https://github.com/nrkno/tv-automation-sofie-code-standard-preset and use the license script from there. We will want to do that anyway sometime, so may as well be now

nytamin:
After discussion with Julian:
We're merging this for now, as it fixes the immediate issue. It'll be replaced later when we get around to start using  https://github.com/nrkno/tv-automation-sofie-code-standard-preset.

PR #461, opened 2021-02-16
Fix: Take after execute cache usage in takeNextPartInnerSync

thomasslee97:
Fixes a bug where you could not have an adlib action queue and `takeAfterExecute` a part at the end of a rundown.

`takeNextPartInnerSync` will throw an error because `dbPlaylist` has no `nextPartInstanceId`

The checks for `activationId` and `nextPartInstanceId` will happen on the playlist fetched from the cache, so in the case where we have an existing cache that we're operationg on we won't be checking this against the database copy.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/playout/take.ts
+++ b/meteor/server/api/playout/take.ts
@@ -75,14 +75,17 @@ export function takeNextPartInnerSync(
 	existingCache?: CacheForRundownPlaylist
 ) {
 	const span = profiler.startSpan('takeNextPartInner')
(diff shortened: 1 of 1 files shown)


Julusian (approved):
This is already handled by https://github.com/nrkno/tv-automation-server-core/pull/455/files#diff-076e60eade35b31f91310057027c396431a0b9ea58b278d8f4a4ad51e2c00135R40, but not 100% certain that will make it to r31 (I think it will)

sparkpunkd:
We know that the PR Julian's mentionned is now targetted to Release 32, so I will recommend this PR is accepted for R31.

PR #465, opened 2021-02-22
fix: Presenter screen not working

jstarpl:

Bug fix

Displaying the Presenter screen in CasparCG results in the following:

```
[2021-02-18 15:33:23.469] [13304] [error]   html[https://[sofie]/countdowns/studio0/presenter] Log: Uncaught TypeError: Promise.allSettled is not a function
[2021-02-18 15:33:23.470] [13304] [error]   html[https://[sofie]/countdowns/studio0/presenter] Log: Uncaught TypeError: Cannot destructure property `ServiceConfiguration` of 'undefined' or 'null'.
```

A shim is included to emulate the behavior of `Promise.allSettled` in browser that don't support it (like Chromium in CasparCG 2.1.12).

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/lib/polyfill/polyfills.ts
+++ b/meteor/client/lib/polyfill/polyfills.ts
@@ -1,2 +1,3 @@
 import './requestIdleCallback'
 import './vibrate'
+import './promise.allSettled'
--- a/meteor/client/lib/polyfill/promise.allSettled.ts
+++ b/meteor/client/lib/polyfill/promise.allSettled.ts
@@ -1,0 +1,10 @@
+/**
+ * CasparCG HTML renderer in 2.1.12 uses an old version of Chromium that doesn't support Promise.allSettled and as a
+ * result causes views rendered using it to crash. Promise.allSettled is used as a part of the i18next integration.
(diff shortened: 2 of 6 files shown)


sparkpunkd (approved):
Good spot and thanks for the good comment to to explain the reason for the polyfill in this context.

Julusian:
I feel like this polyfill should not be necessary.
Meteor produces two client bundles, one for modern browsers and one for legacy, so either casparcg chromium is choosing the wrong one, or the legacy bundling is bad

jstarpl:
@Julusian I think it's possible that Meteor considers CasparCG's browser "modern enough": https://docs.meteor.com/changelog.html#changes-23 - this seems to indicate that the switch is based on a browser version number, not a particular feature set. On the other hand, forcing CasparCG to use legacy bundle, when 99,9% of the "modern" code will work is harsh and may impact performance of the code running in the CasparCG renderer.

Julusian:
@jstarpl yeah, but I would expect meteor to be transpiling the very new features for older 'modern' browsers. They say the cut off is async/await support, but allSettled is a much more recent addition.

Either way, I don't object to this, Im just curious as to how it became a problem

PR #466, opened 2021-02-23
fix: display duration group display duration instability after a take

jstarpl:

Bug fix

During a take, the duration of the "previous" part is unstable, because `timings.duration` is set only once `stoppedPlayback` callback happens. As a result, the following "current" part, if a member of a display duration group will receive all of the "previous" part's `expectedDuration`, causing the part countdown to shift a bit.

The `timings.takeOut` timestamp, which is being set by Core on user action is temporarily used as a duration. There still is a bit of a shift from the difference between `startedPlayback - takeOut` and `duration`, but it's not noticeable in the countdown.

* **Other information**:

**Status**


Julusian:
I have cherrypicked the license checker changes to release31 to fix the build issues there

sparkpunkd (approved):
I guess we might still see a small bump but this will be much improved. Good catch 😄

PR #473, opened 2021-03-05
feat: reset segments upon leaving

Julusian:

feature

PartInstances are only reset by an explicit reset operation for the whole playlist, or when the part is set as next. (Other rules apply for orphaned instances)

When the segment has finished playback it is reset
When the next point is set earlier in the segment it will reset instances as appropriate to show what is planned

* **Other information**:

Additionally, the nextSegmentId has been tested a bit and fixed 

**Status**


sparkpunkd (approved):
This PR seems to be as described, along with a few bonus improvements to typings and better use of let vs const along the way.  Approved 😄

PR #477, opened 2021-03-15
feat: remove AsRunLog collection

Julusian:

feature

Timings are stored in the AsRunLog collection, to be read by the `onAsRunEvent` blueprint method.
This works fine for shows a few hours long, but data in this collection is being retained for 50 days, and the blueprints often need to load all the AsRunLog for the current rundown (the log has no knowledge of activations). This can result in many tens to hundreds of thousands of documents in this collection which get loaded into memory.

While the events are retained for 50 days, none of the pieceInstances/partInstances/segments/rundown referenced by the events are, meaning the events will be hard to interpret without a snapshot or reingesting the rundown and some interpretation to rematch the ids

We have always also stored the timings on the Parts & Pieces, but these could not be relied upon as they would be overwritten/deleted at various times. Since PartInstances, those timings are on the PartInstances and PieceInstances, and once set are never changed or removed until the Rundown is deleted from Sofie.
The `onAsRunEvent` blueprint method has been rethought to be able to operate directly on the PartInstance and PieceInstance collections for this data.

Additional improvements:
* Only find 'events' from the current activation
* Reduces DB operations, as the PartInstances would be loaded for some of the events
* Knowledge of segment 'playback', so that events can be found for the current play of a segment

As a consequence the 50 day retention of events has been removed. If this is desired again, then either the PartInstances and PieceInstances could be preserved, or snapshots could be auto-generated when removing a rundown.

* **Other information**:

**Status**


packages/blueprints-integration/src/rundown.ts
```diff
@@ -292,6 +299,13 @@ export interface IBlueprintPieceInstance<TMetadata = unknown> {
 
 	piece: IBlueprintPieceDB<TMetadata>
 
+	/** The time the system started playback of this part, null if not yet played back (milliseconds since epoch) */
+	startedPlayback?: Time
```
sparkpunkd, line 303: Is the comment supposed to be "... _undefined_ if not yet played back" or is the time `Time | null`?

packages/blueprints-integration/src/rundown.ts
```diff
@@ -292,6 +299,13 @@ export interface IBlueprintPieceInstance<TMetadata = unknown> {
...
+	/** Whether the piece has stopped playback (the most recent time it was played).
+	 * This is set from a callback from the playout gateway (milliseconds since epoch)
+	 */
+	stoppedPlayback?: Time
```
sparkpunkd, line 307: Similarly, what is the not yet set value?

packages/blueprints-integration/src/rundown.ts
sparkpunkd, line 303: Is the comment "... _undefined_ if not yet played back ... " or the type `Time | null` ?

meteor/server/api/blueprints/events.ts
```diff
@@ -0,0 +1,256 @@
...
+				asyncCollectionFindOne(
+					PartInstances,
+					{
+						rundownId: partInstance.rundownId,
+						playlistActivationId: partInstance.playlistActivationId,
+						takeCount: { $lt: partInstance.takeCount },
+					},
+					{
+						sort: {
+							// TODO - test
```
sparkpunkd, line 78: Test done?

meteor/server/api/blueprints/context/context.ts
```diff
@@ -594,91 +583,109 @@ export class AsRunEventContext extends RundownContext implements IAsRunEventCont
...
-				})
+	getFirstPartInstanceInRundown(): Readonly<IBlueprintPartInstance<unknown>> {
+		const partInstance = PartInstances.findOne(
+			{
+				rundownId: this._rundown._id,
+				playlistActivationId: this._part.playlistActivationId,
+			},
+			{
+				sort: {
+					// TODO - verify sort
```
sparkpunkd, line 630: Sort verified?

meteor/server/api/blueprints/context/context.ts
```diff
@@ -545,41 +553,22 @@ export class TimelineEventContext extends RundownContext implements ITimelineEve
 	}
 }
 
-export class AsRunEventContext extends RundownContext implements IAsRunEventContext {
-	public readonly asRunEvent: Readonly<IBlueprintAsRunLogEvent>
-
+export class AsRunRundownEventContext extends RundownContext implements IAsRunRundownEventContext {
```
sparkpunkd, line 556: Semantically, is this still _AsRun..._ - I'm fine with this conceptually - this is the as run timing information for this rundown or part. Its just with the collection going away ...

sparkpunkd (approved):
Approved subject to some minor issues to address. Otherwise, time to say goodbye to unecesarry AsRun logs.

packages/blueprints-integration/src/rundown.ts
Julusian: should read as undefined. Thats what I get for moving a property and comments without reading it..

meteor/server/api/blueprints/context/context.ts
Julusian: Ive renamed these to be a bit clearer now

meteor/server/migration/0_25_0.ts
```diff
@@ -98,18 +97,18 @@ export const addSteps = addMigrationSteps('0.25.0', [
...
-		'AsRunLog',
-		{
-			rundownId: 'runningOrderId',
-			// segmentId:	'segmentId',
-			partInstanceId: 'segmentLineId',
-			pieceInstanceId: 'segmentLineItemId',
-		},
-		'migrateDatabaseCollections'
-	),
+	// renamePropertiesInCollection(
```
nytamin, line 100: I'm missing a migration in x_x_x.ts which clears up the no-longer-used asRunLog collection. See how we did it in R25: https://github.com/nrkno/tv-automation-server-core/blob/master/meteor/server/migration/deprecatedDatabases/0_25_0.ts#L18

meteor/lib/lib.ts
```diff
@@ -946,6 +946,9 @@ export function unprotectString(protectedStr: ProtectedString<any> | undefined |
 export function unprotectStringArray(protectedStrs: Array<ProtectedString<any>>): string[] {
 	return (protectedStrs as any) as string[]
 }
+export function unDeepString<T extends ProtectedString<any>>(str: ReadonlyDeep<T> | PartialDeep<T>): T {
```
nytamin, line 949: should be unprotectStringDeep?

meteor/server/api/blueprints/context/context.ts
```diff
@@ -594,91 +583,109 @@ export class AsRunEventContext extends RundownContext implements IAsRunEventCont
...
-	getParts(): Array<IBlueprintPartDB> {
-		return unprotectObjectArray(this._rundown.getParts())
+}
+
+export class RundownTimingEventContext extends RundownDataChangedEventContext implements IRundownTimingEventContext {
+	readonly previousPart: Readonly<IBlueprintPartInstance<unknown>> | undefined
+	private readonly _part: PartInstance
+	readonly nextPart: Readonly<IBlueprintPartInstance<unknown>> | undefined
+
+	public get part(): Readonly<IBlueprintPartInstance<unknown>> {
```
nytamin, line 602: Question: Should this be named currentPart, to be constistent with previous & next?

meteor/server/api/blueprints/context/context.ts
```diff
@@ -594,91 +583,109 @@ export class AsRunEventContext extends RundownContext implements IAsRunEventCont
...
-		return []
+
+		return unprotectPartInstance(partInstance)
 	}
 
-	formatDateAsTimecode(time: number): string {
-		check(time, Number)
-		return formatDateAsTimecode(new Date(time))
+	getPartInstancesInSegmentPlayoutId(
+		refPartInstance: Readonly<IBlueprintPartInstance<unknown>>
```
nytamin, line 647: needs doc: what should the refPartInstance be?

nytamin: never mind, I found the answer in blueprints-integration

nytamin (changes requested):
A few comments, looks good in general.

meteor/server/api/blueprints/context/context.ts
Julusian: I did wonder that, but this is consistent with some other interfaces here.
But yeah I think I will change this to currentPart for better symmetry within this interface

meteor/lib/lib.ts
Julusian: should it? its converting `ReadonlyDeep<RundownId>` to `RundownId`, so its still a protected string

Julusian: `unDeepProtectedString`?

PR #480, opened 2021-03-18
feat: split security check out of lock functions

Julusian:

Fix

Some security warnings are thrown at startup in development. 

Fixes these warnings. This is done mostly by splitting the security check out of the lock functions, with the result of the check being passed into the lock function instead

* **Other information**:

We have discussed removing a lot of the `ServerPlayoutAPI` as the `UserActionsAPI` should always be used.
I have done that as part of this, as otherwise each method needs updating to be hooked into the security correctly.

**Status**

### DIFF
--- a/meteor/server/api/lib.ts
+++ b/meteor/server/api/lib.ts
@@ -4,6 +4,11 @@ import { RundownPlaylistId, RundownPlaylist } from '../../lib/collections/Rundow
 import { Rundown, RundownId } from '../../lib/collections/Rundowns'
 import { RundownPlaylistContentAccess, RundownPlaylistContentWriteAccess } from '../security/rundownPlaylist'
 
+/**
+ * This is returned from a check of access to a playlist, when access is granted.
+ * Fields will be populated about the user.
+ * It is identical to RundownPlaylistContentAccess, except for confirming access is allowed
+ */
 export interface VerifiedRundownPlaylistContentAccess extends RundownPlaylistContentAccess {
 	playlist: RundownPlaylist
 }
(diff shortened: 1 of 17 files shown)


meteor/server/api/lib.ts
```diff
@@ -2,7 +2,24 @@ import { Meteor } from 'meteor/meteor'
 import { MethodContext } from '../../lib/api/methods'
 import { RundownPlaylistId, RundownPlaylist } from '../../lib/collections/RundownPlaylists'
 import { Rundown, RundownId } from '../../lib/collections/Rundowns'
-import { RundownPlaylistContentWriteAccess } from '../security/rundownPlaylist'
+import { RundownPlaylistContentAccess, RundownPlaylistContentWriteAccess } from '../security/rundownPlaylist'
+
+export interface VerifiedRundownPlaylistContentAccess extends RundownPlaylistContentAccess {
```
nytamin, line 7: I think this would benefit from some doc

meteor/server/api/playout/playout.ts
```diff
@@ -985,7 +995,7 @@ export namespace ServerPlayoutAPI {
 		triggerWriteAccessBecauseNoCheckNecessary() // tmp
 
 		return runPlayoutOperationWithLock(
-			context,
+			null, // TODO: should security be done?
```
nytamin, line 998: In the end? Yes, I think all methods should ideally be access controlled properly, even the callbacks from devices.
But I suppose that it's a bigger issue, let's cross that stream later.

nytamin (approved):
A few minor comments, Looks good!

meteor/server/api/playout/playout.ts
Julusian: This comment was already on one of these methods, and I don't think it work the same way as the other methods, so thought it best to do this for now.

PR #482, opened 2021-03-23
feat: import gateways to mono-repo

Julusian:

Workflow change

The core gateways live as their own repositories. The purpose of these gateways is to bridge between sofie and other systems via an external library, and they are very tied to the sofie internal apis. This makes them not very useful outside of the sofie ecosystem.

As some of the integration libraries have been rehomed into this to form a mono-repo, it makes sense to bring in the core gateways too.
This will help keep them in sync with sofie core apis, and reduce the maintenance cost.

* **Other information**:

For reviewing, there is a commit for simply importing the two gateways. Then a commit of all the configuration and build changes. Not everything is perfect, some tidying should be done in the gateways still, but I am waiting for the final import of the gateway code before doing that.

Still to do:
- [x] Merge changes in jenkins-lib
- [ ] Verify gateways build properly once merged
- [x] Ensure gateway code is latest 
- [ ] Put a relocated warning in the old gateway repositories
- [ ] Deploy and verify gateway images run correctly. Verified locally, would be good to test the proper builds too
- [ ] Look into better startup scripts for sofie. it would be useful to run playout-gateway via `npm run dev` in the meteor folder, next to core (not interleaved for readability)

**Status**


sparkpunkd (approved):
I got this all up and running 😄. I made a minor tweak to the "packages/package.json" file so that "lerna bootstrap" will run OK from "meteor npm install" as per README.

One other thing is that the version number of playout-gateway and core don't (yet) match. But this can be fixed on release.

PR #483, opened 2021-03-24
feat: zero-based integers in settings

thomasslee97:

This adds a property to integer config manifest entries named `zeroBased`, which indicates that this is a value that should be stored in the databse as a value starting from 0, but makes more sense to show to the user as a value starting from 1. For example, an ATEM Aux makes more sense to show to the user as AUX 1->X than AUX 0-> Y.

Currently all numerical config values are displayed to the user as whatever is stored in the database, which is usually the value required by the consuming device/blueprint.

Integer config fields for both blueprints and device manifests can be marked as `zeroBased`, meaning that if the user wants to, for example, map a layer to sisyfos fader 1, they can enter 1 into the UI and the value 0 will be stored in the database. The motivation here is to make the UI match the language used by devices. ie, the ATEM control software uses AUX 1, AUX 2, ME 1, ME 2 etc whereas Sofie currently will display these to the user as AUX 0, AUX 1, ME 0, ME 1.

This is configurable on a per-config basis, so devices like Quantel which use 0-based naming in their 1st-party products/UIs can be presented using the same language in Sofie.

* **Other information**:

This PR also marks the config manifest entry type `NUMBER` as deprecated in favour of `INT` and `FLOAT`. The reason for this is that `INT` and `FLOAT` have already been introduced previously, but `NUMBER` is inconsistently interpreted. If you use a `NUMBER` in a blueprint manifest, it will be treated as an integer. However, if a `NUMBER` is used in a device manifest, it will be treated as a float.

Moving to an explicit distinction between `INT` and `FLOAT` also means that the `zeroBased` property can be scoped to just integer values in the future. This would also make merging the various definitions for config manifests easier in the future, once `NUMBER` is removed. No functionality is changed by marking `NUMBER` as deprecated.

**Status**

### DIFF
--- a/meteor/client/ui/Settings/ConfigManifestSettings.tsx
+++ b/meteor/client/ui/Settings/ConfigManifestSettings.tsx
@@ -123,6 +123,32 @@ function getEditAttribute<DBInterface extends { _id: ProtectedString<any> }, Doc
 					type="int"
 					collection={collection}
 					className="input text-input input-m"
+					mutateDisplayValue={(v) => (item.zeroBased ? v + 1 : v)}
+					mutateUpdateValue={(v) => (item.zeroBased ? v - 1 : v)}
+				/>
+			)
+		case ConfigManifestEntryType.INT:
+			return (
+				<EditAttribute
+					modifiedClassName="bghl"
+					attribute={attribute}
+					obj={object}
+					type="int"
+					collection={collection}
+					className="input text-input input-m"
+					mutateDisplayValue={(v) => (item.zeroBased ? v + 1 : v)}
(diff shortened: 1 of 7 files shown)


thomasslee97:
Rebased to include this PR against playout gateway https://github.com/nrkno/tv-automation-playout-gateway/pull/75

sparkpunkd (approved):
Approved ... although we need a plan to role this out into the manifests to deal with the deprecation of NUMBER and get the zero-based properties labelled ... such as the one for media manager. For example:
- need to get rid of deprecated NUMBER properties
- need to get Quantel properties that are zero-based labelled appropriately.

But ... nothing should break either. Good job! 😃

PR #484, opened 2021-03-26
feat: add EditAttributeArray to support inputting of basic arrays.

nytamin:
feature

no feature :)

This PR adds EditAttributeArray to support inputting of basic arrays.
Use arrayType: 'boolean' | 'int' | 'float' | 'string' to set the types of the inner values in the array.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/lib/EditAttribute.tsx
+++ b/meteor/client/lib/EditAttribute.tsx
@@ -10,7 +10,6 @@ import ClassNames from 'classnames'
 import { ColorPickerEvent, ColorPicker } from './colorPicker'
 import { IconPicker, IconPickerEvent } from './iconPicker'
 import { Random } from 'meteor/random'
-import { number } from 'prop-types'
 
 interface IEditAttribute extends IEditAttributeBaseProps {
 	type: EditAttributeType
@@ -951,14 +950,8 @@ const EditAttributeArray = wrapEditAttribute(
 					if (Number.isNaN(parsed)) return false // type check failed
 					values.push(parsed)
 				} else {
-					// if (this.props.arrayType === 'string')
-					let parsed: string
-					try {
-						parsed = JSON.parse(str)
-					} catch (_err) {
-						// ignore error
-						parsed = str + ''
-					}
+					// else this.props.arrayType is 'string'
+					const parsed = str + ''
 					if (typeof parsed !== 'string') return false // type check failed
 					values.push(parsed.trim())
 				}
@@ -1003,8 +996,6 @@ const EditAttributeArray = wrapEditAttribute(
 		getAttribute() {
 			const value = super.getAttribute()
 			if (Array.isArray(value)) {
-				// const str = JSON.stringify(value, null, 2) // "[...values]"
-				// return str.slice(1, -1) // remove [ and ]
 				return value.join(', ')
 			} else {
 				return ''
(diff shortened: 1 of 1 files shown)


meteor/client/lib/EditAttribute.tsx
```diff
@@ -10,6 +10,7 @@ import ClassNames from 'classnames'
 import { ColorPickerEvent, ColorPicker } from './colorPicker'
 import { IconPicker, IconPickerEvent } from './iconPicker'
 import { Random } from 'meteor/random'
+import { number } from 'prop-types'
```
jstarpl, line 13: This import doesn't seem to do anything.

meteor/client/lib/EditAttribute.tsx
```diff
@@ -913,6 +919,121 @@ const EditAttributeJson = wrapEditAttribute(
...
+			let e = event as KeyboardEvent
+			if (e.key === 'Escape') {
+				this.handleDiscard()
+			}
+		}
+		getAttribute() {
+			const value = super.getAttribute()
+			if (Array.isArray(value)) {
+				// const str = JSON.stringify(value, null, 2) // "[...values]"
+				// return str.slice(1, -1) // remove [ and ]
```
jstarpl, line 1007: These comments don't seem to add much and the solution of using `.join()` seems to be much more sensible.

meteor/client/lib/EditAttribute.tsx
```diff
@@ -913,6 +919,121 @@ const EditAttributeJson = wrapEditAttribute(
...
+					values.push(parsed)
+				} else if (this.props.arrayType === 'float') {
+					const parsed = parseFloat(str)
+					if (Number.isNaN(parsed)) return false // type check failed
+					values.push(parsed)
+				} else {
+					// if (this.props.arrayType === 'string')
+					let parsed: string
+					try {
+						parsed = JSON.parse(str)
```
jstarpl, line 957: What's the point of the `JSON.parse` here? Wouldn't it be enough to just cast `str` to string? Should probably be quicker than `JSON.parse`, especially if it actually is a long string and it will be trying to decide if it's an object or not.

meteor/client/lib/EditAttribute.tsx
```diff
@@ -913,6 +919,121 @@ const EditAttributeJson = wrapEditAttribute(
...
+					} catch (_err) {
+						// ignore error
+						parsed = str + ''
+					}
+					if (typeof parsed !== 'string') return false // type check failed
```
jstarpl, line 962: Also, strings that will look like a JSON object - but perhaps aren't, and need not be treated as such - will fail? Also, something like "1,2,3,4,MP1,MP2" will fail here, and I strongly think that it shouldn't ("1" is a perfectly acceptable string label).

jstarpl (changes requested):
I think the handling of string arrays should be improved since the `.isArray()` can fail on them if certain values are used, and I think it should generally accept them. Once such case is a string array of "1,2,3,MP1,MP2".

meteor/client/lib/EditAttribute.tsx
nytamin: Good point! I think I was aiming for something that handled the `"asdf","qwerty"` case, but it really doesn't made sense after you pointed it out. Casting to string is the way to go 👍

meteor/client/lib/EditAttribute.tsx
nytamin: This should now be okay after having removed the JSON.parse part.

jstarpl (approved):

PR #486, opened 2021-03-31
feat: playout-gw atem clip uploading

baltedewit:
Currently the playout-gateway can upload stills to the atem switcher, this PR adds clip uploading to that functionality.

Because the uploading functionality has matured a fair bit since the original implementation I've changed the uploading to upload all assets from a single process rather than spawn new processes for every upload. In addition this means the media is uploaded one after the other instead of all at once.

**Status**

### DIFF
--- a/packages/playout-gateway/src/atemUploader.ts
+++ b/packages/playout-gateway/src/atemUploader.ts
@@ -22,13 +22,13 @@ function consoleError(...args: any[]) {
 }
 export class AtemUploadScript {
 	private readonly connection: Atem
-	
+
 	constructor() {
 		this.connection = new Atem()
-		
+
 		this.connection.on('error', consoleError)
 	}
-	
+
 	public connect(ip: string): Promise<void> {
 		return new Promise<void>((resolve, reject) => {
 			this.connection.once('connected', () => {
@@ -39,7 +39,7 @@ export class AtemUploadScript {
 			})
 		})
 	}
-	
+
 	public loadFile(fileUrl: string): Promise<Buffer> {
 		return new Promise<Buffer>((resolve, reject) => {
 			fs.readFile(fileUrl, (e, data) => {
@@ -49,39 +49,34 @@ export class AtemUploadScript {
 			})
 		})
 	}
-	public loadFiles(folder: string): Promise<Buffer[]> {
-		return new Promise<Buffer[]>(async (resolve, reject) => {
-			try {
-				const files = await fs.promises.readdir(folder)
-				consoleLog(files)
-				const loadBuffers = files.map(file => fs.promises.readFile(path.join(folder, file)))
-				const buffers = await Promise.all(loadBuffers)
-				
-				consoleLog('got files')
-				resolve(buffers)
-			} catch (e) {
-				console.error(e)
-				reject(e)
-			}
-		})
+	public async loadFiles(folder: string): Promise<Buffer[]> {
+		const files = await fs.promises.readdir(folder)
+		consoleLog(files)
+		const loadBuffers = files.map((file) => fs.promises.readFile(path.join(folder, file)))
+		const buffers = await Promise.all(loadBuffers)
+
+		consoleLog('got files')
+		return buffers
 	}
-	
+
 	public checkIfFileOrClipExistsOnAtem(fileName: string, stillOrClipIndex: number, type: AtemMediaPoolType): boolean {
 		consoleLog('got a file')
-		
+
 		const still = this.connection.state ? this.connection.state.media.stillPool[stillOrClipIndex] : undefined
 		const clip = this.connection.state ? this.connection.state.media.clipPool[stillOrClipIndex] : undefined
 		let pool: typeof still | typeof clip
 		if (type === AtemMediaPoolType.Still) pool = still
 		if (type === AtemMediaPoolType.Clip) pool = clip
-		
+
 		if (pool) {
 			consoleLog('has ' + type)
 			if (pool.isUsed) {
 				consoleLog(type + ' is used')
-				const comparisonName = fileName.substr(type === AtemMediaPoolType.Still ? -ATEM_MAX_FILENAME_LENGTH : -ATEM_MAX_CLIPNAME_LENGTH)
+				const comparisonName = fileName.substr(
+					type === AtemMediaPoolType.Still ? -ATEM_MAX_FILENAME_LENGTH : -ATEM_MAX_CLIPNAME_LENGTH
+				)
 				const poolName = 'fileName' in pool ? pool.fileName : pool.name
-				
+
 				if (poolName === comparisonName) {
 					consoleLog('name equals')
 					return true
@@ -96,7 +91,7 @@ export class AtemUploadScript {
 			throw Error('Atem appears to be missing ' + type)
 		}
 	}
-	
+
 	public async uploadStillToAtem(fileName: string, fileData: Buffer, stillIndex: number): Promise<void> {
 		fileName = fileName.substr(-ATEM_MAX_FILENAME_LENGTH) // cannot be longer than 63 chars
 		if (!this.checkIfFileOrClipExistsOnAtem(fileName, stillIndex, AtemMediaPoolType.Still)) {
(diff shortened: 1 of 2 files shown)


packages/playout-gateway/src/atemUploader.ts
```diff
@@ -39,29 +40,44 @@ export class AtemUploadScript {
...
 				if (e) reject(e)
 				else resolve(data)
 			})
 		})
 	}
+	public async loadFiles(folder: string): Promise<Buffer[]> {
+		const files = await fs.promises.readdir(folder)
+		consoleLog(files)
+		const loadBuffers = files.map((file) => fs.promises.readFile(path.join(folder, file)))
+		const buffers = await Promise.all(loadBuffers)
```
gundelsby, line 56: This await is unnecessary, given that the function returns a `Promise<Buffer[]>` anyway

packages/playout-gateway/src/atemUploader.ts
```diff
@@ -39,29 +40,44 @@ export class AtemUploadScript {
...
-			if (still.isUsed) {
-				consoleLog('still is used')
-				if (fileName.length === AtemMaxFilenameLength) {
-					consoleLog('filename is max length, change detection might fail')
-				}
+		const still = this.connection.state ? this.connection.state.media.stillPool[stillOrClipIndex] : undefined
+		const clip = this.connection.state ? this.connection.state.media.clipPool[stillOrClipIndex] : undefined
+		let pool: typeof still | typeof clip
+		if (type === AtemMediaPoolType.Still) pool = still
+		if (type === AtemMediaPoolType.Clip) pool = clip
```
gundelsby, line 69: Should be an `else if` here, given that the second if can never be true if the first one is

packages/playout-gateway/src/atemUploader.ts
```diff
@@ -72,53 +88,67 @@ export class AtemUploadScript {
...
-			consoleLog('does not exist on ATEM')
+	public async uploadStillToAtem(fileName: string, fileData: Buffer, stillIndex: number): Promise<void> {
+		fileName = fileName.substr(-ATEM_MAX_FILENAME_LENGTH) // cannot be longer than 63 chars
+		if (!this.checkIfFileOrClipExistsOnAtem(fileName, stillIndex, AtemMediaPoolType.Still)) {
+			consoleLog(fileName + ' does not exist on ATEM')
 			await this.connection.clearMediaPoolStill(stillIndex)
 			await this.connection.uploadStill(stillIndex, fileData, fileName, '')
 		} else {
-			consoleLog('does exist on ATEM')
+			consoleLog(fileName + ' does exist on ATEM')
```
gundelsby, line 102: Shouldn't the else branch throw an exception to indicate that the still has not been uploaded?

packages/playout-gateway/src/atemUploader.ts
```diff
@@ -72,53 +88,67 @@ export class AtemUploadScript {
...
+		}
+	}
+	public async uploadClipToAtem(name: string, fileData: Buffer[], clipIndex: number): Promise<void> {
+		name = name.substr(-ATEM_MAX_CLIPNAME_LENGTH) // cannot be longer than 43 chars
+		if (!this.checkIfFileOrClipExistsOnAtem(name, clipIndex, AtemMediaPoolType.Clip)) {
+			consoleLog(name + ' does not exist on ATEM')
+			await this.connection.clearMediaPoolClip(clipIndex)
+			await this.connection.uploadClip(clipIndex, fileData, name)
+		} else {
+			consoleLog(name + ' does exist on ATEM')
```
gundelsby, line 112: Shouldn't the else branch throw an exception to indicate that the clip has not been uploaded?

packages/playout-gateway/src/atemUploader.ts
baltedewit: That's not my intention, if the function successfully executes that indicates the media exists on the atem. If the media already exists on the atem no upload has to be done but that is not an error because the media will exist after the function returns. The whole functionality here is intended to be replaced by the package manager so I think it's justified to leave it as is

packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -740,31 +725,24 @@ export class TSRHandler {
...
-						process.stderr.on('data', (data) => this.logger.info(data.toString()))
-						process.on('close', () => {
-							process.removeAllListeners()
-						})
-					} else {
-						throw Error('ATEM host option not set')
-					}
+	private uploadFilesToAtem(files: AtemMediaPoolAsset[]) {
+		this.logger.info('try to load ' + JSON.stringify(files.map((f) => f.path).join(', ')) + ' to atem')
+		this.tsr.getDevices().forEach(async (device) => {
```
Julusian, line 730: This was pre-existing but is excessive and a bit broken. `uploadFilesToAtem` is called only when a device has a status change, and we then upload to every atem the assets from the one with the status change..

`uploadFilesToAtem` should be updated to only operate on the device that had the status change (perhaps the only parameter should be `device` or `deviceId`)

packages/playout-gateway/src/atemUploader.ts
```diff
@@ -39,29 +40,44 @@ export class AtemUploadScript {
...
 				consoleLog('got file')
 				if (e) reject(e)
 				else resolve(data)
 			})
 		})
 	}
+	public async loadFiles(folder: string): Promise<Buffer[]> {
+		const files = await fs.promises.readdir(folder)
+		consoleLog(files)
+		const loadBuffers = files.map((file) => fs.promises.readFile(path.join(folder, file)))
```
Julusian, line 55: I think it is dangerous to load all of the frames at once, and is highly likely to run out of memory.

If we assume a 5s clip at 1080 (not unreasonable for a wipe with trailing audio) then we need 1gb of free ram to load in every raw frame at once.

To completely fill the 2me4k we would need ~5gb, and a constellation would take ~12gb.

That said, this might be good enough for now if we know our clips are going to be really short <500mb

Julusian (changes requested):
I like the structural changes to the uploader.
In reality, we were never doing uploads in parallel due to locking on the atem, but this way we are intentionally queuing uploads rather than queuing for the locks.

I have raised a couple of concerns, but I am open to simply dismissing them as 'fix later'

packages/playout-gateway/src/atemUploader.ts
baltedewit: That's a good point, unfortunately not something that can be fixed in this PR as it is just how the API in atem-connection was designed. I don't plan to upload audio as part of any wipes currently, just masks and overlays, so practically we should never hit this issue. A good thing to pick up when we implement this in the media-manager though.

packages/playout-gateway/src/atemUploader.ts
gundelsby: If I put my super pedantic hat on I could say that you only know that an item with the same (truncated) name exists on the atem 🙃 
But if it's short lived code and we've never had any issues with this so far I guess it's fine to leave it like that until we replace this with the media manager.

Julusian: that is true, but why would we exception there?
If we are throwing an exception that implies that something is wrong, but nothing is wrong here. We are choosing to skip the upload as we believe the correct file is already there

packages/playout-gateway/src/atemUploader.ts
Julusian: good point, I didnt think about that

I've created https://github.com/nrkno/tv-automation-atem-connection/issues/110 as a reminder to do something about this

packages/playout-gateway/src/atemUploader.ts
gundelsby: It might be that I don't know enough about the code that would call this function, but I guess my logic goes something like this:
- caller wants a file uploaded
- the function does not upload the file because there already is a file with the same name on the atem. Whether or not it is the same file is unknown
- the file that was requested uploaded is not uploaded, and that is an exception from the original expectation of the caller and should be flagged as such

But yeah, I might be missing information and/or constructing a problem where there is none :)

PR #488, opened 2021-03-31
Feat: Layer name

thomasslee97:

Adds support for configuring the `layerName` of a mapping and uses it in the layer mapping dropdown in settings.

No feature.

A setting is added to mapping settings to allow the input of a layer name, this name is then used instead of the layer Id in the mapping table.

image

Additionally, where the layer mapping dropdown type is used in settings, the layer name will be used if it's available (here GFX Ident is displayed instead of `graphic_layer_ident`)

image

* **Other information**:

See https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/178 for information on how this is used on a device level.

**Status**


baltedewit (approved):
I don't see a reason to reject this code wise but the NRK team expresses their concern regarding the clarity of non unique names. Whilst this makes them user representable it can also make it make it more error prone to someone configuring the system if they can't see what exact mapping is used.

PR #489, opened 2021-03-31
Feat: Find last scripted piece on layer

thomasslee97:

Adds a pair of functions to allow adLib actions to look backwards in the script from the current point to find the last *scripted* piece on a layer as apposed to the last played piece.

AdLib actions can only find the last played piece.

AdLib actions can call `findLastScriptedPieceOnLayer` which will look backwards from the current part to find the piece on a given sourcelayer that appears closest to the current part. The use case for this is, for example, a producer is on a VT going into an interview over a Live feed, but the interviewee is not ready yet. The producer chooses to skip over the scripted Live to the Camera after it. When the interviewee is ready they press a button which runs an adLib action to find the last scripted Live and cut to it. In this case the live never started playing, so the current implementation of `findLastPieceOnLayer` would not find it.

This PR also adds the function `getPartForPreviousPiece` as a counterpart to `getPartInstanceForPreviousPiece`, allowing the part associated with the found piece to be fetched. This allows e.g. the adLib action to create a new part with the same title.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/blueprints/context/adlibActions.ts
+++ b/meteor/server/api/blueprints/context/adlibActions.ts
@@ -217,7 +217,10 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 			throw new Error('Cannot find Part from invalid Piece')
 		}
 
-		const pieceDB = Pieces.findOne({ _id: protectString(piece._id) })
+		const pieceDB = Pieces.findOne({
+			_id: protectString(piece._id),
+			startRundownId: { $in: this._cache.Playlist.doc.getRundownIDs() },
+		})
 		if (!pieceDB) throw new Error(`Cannot find Piece ${piece._id}`)
 
 		return Parts.findOne({ _id: pieceDB.startPartId })
--- a/meteor/server/api/blueprints/context/adlibActions.ts
+++ b/meteor/server/api/blueprints/context/adlibActions.ts
@@ -155,7 +155,7 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 	}
 
 	findLastScriptedPieceOnLayer(
-		sourceLayerId: string,
+		sourceLayerId0: string | string[],
 		options?: {
 			excludeCurrentPart?: boolean
 			pieceMetaDataFilter?: any
@@ -166,19 +166,21 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 			for (const [key, value] of Object.entries(options.pieceMetaDataFilter)) {
 				// TODO do we need better validation here?
 				// It should be pretty safe as we are working with the cache version (for now)
-				query[`piece.metaData.${key}`] = value
+				query[`metaData.${key}`] = value
 			}
 		}
 
 		if (options && options.excludeCurrentPart && this._cache.Playlist.doc.currentPartInstanceId) {
 			const currentPartInstance = this._cache.PartInstances.findOne(
-				(p) => p._id === this._cache.Playlist.doc.currentPartInstanceId
+				this._cache.Playlist.doc.currentPartInstanceId
 			)
 
 			if (currentPartInstance) {
 				query['startPartId'] = { $ne: currentPartInstance.part._id }
 			}
 		}
+
+		const sourceLayerId = Array.isArray(sourceLayerId0) ? sourceLayerId0 : [sourceLayerId0]
 
 		const lastPiece = ServerPlayoutAdLibAPI.innerFindLastScriptedPieceOnLayer(this._cache, sourceLayerId, query)
 
--- a/meteor/server/api/blueprints/context/adlibActions.ts
+++ b/meteor/server/api/blueprints/context/adlibActions.ts
@@ -9,6 +9,7 @@ import {
 	getRandomId,
 	protectStringArray,
 	waitForPromise,
+	UnprotectedStringProperties,
 } from '../../../../lib/lib'
 import { Part, Parts } from '../../../../lib/collections/Parts'
 import { logger } from '../../../../lib/logging'
@@ -209,9 +210,13 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 		}
 	}
 
-	getPartForPreviousPiece(piece: IBlueprintPieceDB): IBlueprintPart | undefined {
+	getPartForPreviousPiece(piece: UnprotectedStringProperties<Pick<Piece, '_id'>>): IBlueprintPart | undefined {
+		if (!piece?._id) {
+			throw new Error('Cannot find Part from invalid Piece')
+		}
+
 		const pieceDB = Pieces.findOne({ _id: protectString(piece._id) })
-		if (!pieceDB) throw new Error(`Cannot find piece ${piece._id}`)
+		if (!pieceDB) throw new Error(`Cannot find Piece ${piece._id}`)
 
 		return Parts.findOne({ _id: pieceDB.startPartId })
 	}
--- a/meteor/server/api/blueprints/context/adlibActions.ts
+++ b/meteor/server/api/blueprints/context/adlibActions.ts
@@ -219,7 +219,7 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 
 		const pieceDB = Pieces.findOne({
 			_id: protectString(piece._id),
-			startRundownId: { $in: this._cache.Playlist.doc.getRundownIDs() },
+			startRundownId: { $in: getRundownIDsFromCache(this._cache) },
 		})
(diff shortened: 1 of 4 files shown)


meteor/server/api/blueprints/context/adlibActions.ts
```diff
@@ -150,6 +153,38 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
...
+			excludeCurrentPart?: boolean
+			pieceMetaDataFilter?: any
+		}
+	): IBlueprintPieceDB | undefined {
+		const query: MongoQuery<Piece> = {}
+		if (options && options.pieceMetaDataFilter) {
+			for (const [key, value] of Object.entries(options.pieceMetaDataFilter)) {
+				// TODO do we need better validation here?
+				// It should be pretty safe as we are working with the cache version (for now)
+				query[`piece.metaData.${key}`] = value
```
Julusian, line 169: the `piece.` prefix should be dropped here

meteor/server/api/blueprints/context/adlibActions.ts
```diff
@@ -150,6 +153,38 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
...
+			}
+		}
+
+		if (options && options.excludeCurrentPart && this._cache.Playlist.doc.currentPartInstanceId) {
+			const currentPartInstance = this._cache.PartInstances.findOne(
+				(p) => p._id === this._cache.Playlist.doc.currentPartInstanceId
```
Julusian, line 175: You can drop the `(p) => p._id === ` off this query, as the collections do allow finding by giving the id

meteor/server/api/playout/adlib.ts
```diff
@@ -411,6 +411,48 @@ export namespace ServerPlayoutAdLibAPI {
...
+		const span = profiler.startSpan('innerFindLastScriptedPieceOnLayer')
+
+		const playlist = cache.Playlist.doc
+		const rundownIds = getRundownIDsFromCache(cache)
+
+		if (!playlist.currentPartInstanceId) {
+			return
+		}
+
+		const currentPartInstance = cache.PartInstances.findOne({ _id: playlist.currentPartInstanceId })
```
Julusian, line 428: Again, query can be simplified

meteor/server/api/playout/adlib.ts
```diff
@@ -411,6 +411,48 @@ export namespace ServerPlayoutAdLibAPI {
...
+			return
+		}
+
+		const query = {
+			...customQuery,
+			startRundownId: { $in: rundownIds },
+			sourceLayerId,
+		}
+
+		if (span) span.end()
```
Julusian, line 440: This should be at the end?

packages/blueprints-integration/src/context.ts
```diff
@@ -128,6 +128,13 @@ export interface IActionExecutionContext extends IShowStyleUserContext, IEventCo
 			pieceMetaDataFilter?: any // Mongo query against properties inside of piece.metaData
 		}
 	): IBlueprintPieceInstance | undefined
+	findLastScriptedPieceOnLayer(
+		sourceLayerId: string,
```
Julusian, line 132: is there a reason that `sourceLayerId` doesnt allow an array?

A documentation comment would be good for the method too

meteor/server/api/playout/adlib.ts
```diff
@@ -411,6 +411,48 @@ export namespace ServerPlayoutAdLibAPI {
...
+		const part = Parts.findOne(
+			{ _id: { $in: pieces.map((p) => p.startPartId) }, _rank: { $lte: currentPartInstance.part._rank } },
+			{ sort: { _rank: -1 } }
+		)
+
+		if (!part) {
+			return
+		}
+
+		return pieces.find((p) => p.startPartId === part._id)
```
Julusian, line 453: Im not keen on this. It wont be deterministic (maybe not even completely consistent) if there are multiple matching pieces. If you filter for all the pieces in the part, and sort those by start time that will produce better results.

It might be appropriate to run them through `processAndPrunePieceInstanceTimings` to do the start time sorting and also collision handling to match how they would be played

Julusian (changes requested):

meteor/server/api/playout/adlib.ts
thomasslee97: Ah, yes it should

meteor/server/api/playout/adlib.ts
thomasslee97: We can't use `processAndPrunePieceInstanceTimings ` because we're working with pieces, so I have sorted the pieces acording to `enable.start`

packages/blueprints-integration/src/context.ts
thomasslee97: I've changed it to accept an array

meteor/server/api/blueprints/context/adlibActions.ts
```diff
@@ -175,6 +212,17 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 		}
 	}
 
+	getPartForPreviousPiece(piece: UnprotectedStringProperties<Pick<Piece, '_id'>>): IBlueprintPart | undefined {
+		if (!piece?._id) {
+			throw new Error('Cannot find Part from invalid Piece')
+		}
+
+		const pieceDB = Pieces.findOne({ _id: protectString(piece._id) })
```
Julusian, line 220: this should have a check on the startRundownId to ensure it belongs to a rundown in the playlist

meteor/server/api/playout/adlib.ts
```diff
@@ -411,6 +411,56 @@ export namespace ServerPlayoutAdLibAPI {
...
+			return
+		}
+
+		const query = {
+			...customQuery,
+			startRundownId: { $in: rundownIds },
+			sourceLayerId: { $in: sourceLayerId },
+		}
+
+		const pieces = Pieces.find(query).fetch()
```
Julusian, line 440: Looking at this again, I am a little concerned about the volume of data here. As you get further through the rundowns, the number of pieces this will find will keep on increasing.

The infinite logic does work similarly, but slightly more scoped (only certain infinite modes). Might be enough of a precedent to leave this be for now.
Or it could fetch only the id and startPartId fields, then do a second query to load the full copy of the pieces in the chosen part?

meteor/server/api/playout/adlib.ts
```diff
@@ -411,6 +411,56 @@ export namespace ServerPlayoutAdLibAPI {
...
+
+		const part = Parts.findOne(
```
Julusian, line 442: oh, `Parts` exists on the cache here

meteor/server/api/playout/adlib.ts
```diff
@@ -411,6 +411,56 @@ export namespace ServerPlayoutAdLibAPI {
...
+		)
+
+		if (!part) {
+			return
+		}
+
+		if (span) span.end()
+
+		return pieces
+			.sort((a, b) => {
```
Julusian, line 454: it would be better to make this a `pieces.filter().sort()[0]`, to make the sort quicker

meteor/server/api/playout/adlib.ts
thomasslee97: Yeah I was concerned about that too, maybe it should be scoped to look back `n` parts? Where `n` can be specified by the adlib action or defaults to 10 or so?

meteor/server/api/playout/adlib.ts
Julusian: True, but you could wrap the pieces to pieceInstances to run them through the function, and then unwrap the result. The ui does that if it is working on a part rather than partinstance

You would need to ensure you load in every piece in the part for this to be correct. maybe not its infinites though?

meteor/server/api/blueprints/context/adlibActions.ts
```diff
@@ -217,7 +217,10 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 			throw new Error('Cannot find Part from invalid Piece')
 		}
 
-		const pieceDB = Pieces.findOne({ _id: protectString(piece._id) })
+		const pieceDB = Pieces.findOne({
+			_id: protectString(piece._id),
+			startRundownId: { $in: this._cache.Playlist.doc.getRundownIDs() },
```
Julusian, line 222: `getRundownIdsFromCache(this._cache)` :)

meteor/server/api/playout/adlib.ts
Julusian: Maybe. That could be added later if this is found to be slow

Julusian (approved):

PR #492, opened 2021-03-31
feat: Add duplicate AdLibs filtering

thomasslee97:

This PR allows blueprints to generate an identifier for adLibs/adLib actions that can then be used to only display one "copy" of a given adLib to the user.

The use case for this is a script writing workflow where, for example, a lower third might be entered into the script for every segment in which an interviewee is expected to appear. This PR allows Sofie to just show one copy of the generated lower third adlib instead of one per segment.

No feature.

The blueprints can add a `uniquenessId` to adLibs/adLib actions, it is left to the blueprint writer to determine how this Id should be generated. There is then an option added to the rundown layout editor to allow a shelf layout to request that only one copy of each adLib with the same `uniquenessId` be returned from the function `matchFilter`.

image

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Shelf/AdLibPanel.tsx
+++ b/meteor/client/ui/Shelf/AdLibPanel.tsx
@@ -172,7 +172,6 @@ export function matchFilter(
 		// Scope of the filter is determined by the scope of the uniquenessIds set (typically rundown-wide).
 		if (filter.hideDuplicates && uniquenessIds) {
 			const uniquenessId = item.uniquenessId || unprotectString(item._id)
-			console.log(uniquenessId)
 			if (uniquenessIds.has(uniquenessId)) {
 				return false
 			} else {
--- a/meteor/client/ui/Shelf/AdLibPanel.tsx
+++ b/meteor/client/ui/Shelf/AdLibPanel.tsx
@@ -266,15 +266,17 @@ const AdLibListView = withTranslation()(
 				<tbody className="adlib-panel__list-view__list__segment adlib-panel__list-view__item__rundown-baseline">
 					{this.props.rundownAdLibs &&
 						this.props.rundownAdLibs
-							.filter((item) =>
-								matchFilter(
-									item,
-									this.props.showStyleBase,
-									this.props.uiSegments,
-									this.props.filter,
-									this.props.searchFilter,
-									uniquenessIds
-								)
+							.filter(
+								(item) =>
+									!item.isHidden &&
+									matchFilter(
+										item,
+										this.props.showStyleBase,
+										this.props.uiSegments,
+										this.props.filter,
+										this.props.searchFilter,
+										uniquenessIds
+									)
 							)
 							.map((adLibPiece: AdLibPieceUi) => (
 								<AdLibListItem
--- a/meteor/client/ui/Shelf/AdLibPanel.tsx
+++ b/meteor/client/ui/Shelf/AdLibPanel.tsx
@@ -92,6 +92,15 @@ interface IListViewStateHeader {
 	}
 }
 
+/**
+ * Applies a filter to an adLib to determine whether it matches filter criteria.
+ * @param item AdLib to test against filter.
+ * @param showStyleBase
+ * @param uiSegments All segments to search for live segment.
+ * @param filter Filter to match against.
+ * @param searchFilter Text to try to match against adLib label.
+ * @param uniquenessIds Set of uniquenessIds, for a given set only one adLib per uniquness Id will be matched by this filter.
+ */
 export function matchFilter(
 	item: AdLibPieceUi,
 	showStyleBase: ShowStyleBase,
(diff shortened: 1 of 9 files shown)


meteor/client/ui/Shelf/AdLibPanel.tsx
```diff
@@ -157,6 +158,16 @@ export function matchFilter(
 		) {
 			return false
 		}
+		// Hide duplicates
+		if (filter.hideDuplicates && uniquenessIds) {
+			const uniquenessId = item.uniquenessId || unprotectString(item._id)
+			console.log(uniquenessId)
```
jstarpl, line 164: This should be removed.

meteor/client/ui/Shelf/AdLibPanel.tsx
```diff
@@ -238,23 +249,22 @@ const AdLibListView = withTranslation()(
...
-									!item.isHidden &&
-									matchFilter(
-										item,
-										this.props.showStyleBase,
-										this.props.uiSegments,
-										this.props.filter,
-										this.props.searchFilter
-									)
+							.filter((item) =>
+								matchFilter(
```
jstarpl, line 260: Why is the `.isHidden` is condition removed?

meteor/client/ui/Shelf/TimelineDashboardPanel.tsx
```diff
@@ -88,6 +88,7 @@ export const TimelineDashboardPanel = translateWithTracker<
 		render() {
 			if (this.props.visible && this.props.showStyleBase && this.props.filter) {
 				const filter = this.props.filter as DashboardLayoutFilter
+				const uniquenessIds = new Set<string>()
```
jstarpl, line 91: There's a lot of this assigning an empty set and then passing it into a function. Seems unnecessary? Why not just pass a `new Set()` into the filtering function? Why not just have the filtering function build itself a new Set, if none is provided?

jstarpl (changes requested):
I feel like the implementation can be tightened a bit, since there seems to a bit of a repetition with the empty `uniquenessIds: Set<string>` being passed around, and only a single instance of it actually used for persistence. Also, I would like the "Hide duplicates" comment to be fleshed out a bit, since what that actually means is a bit vague.

meteor/client/ui/Shelf/AdLibPanel.tsx
```diff
@@ -157,6 +158,16 @@ export function matchFilter(
 		) {
 			return false
 		}
+		// Hide duplicates
```
jstarpl, line 161: I think this should be explained a bit more. I.e. that when this is enabled, only the very first item, rundown-order-wise, will be displayed, and other items will be hidden. That means that, not only "duplicates in a segment" will be hidden, but also "duplicates in any subsequent segment or rundown" will be hidden. Since this is counter-intuitive to how the rest of Sofie works (i.e. all content relevant to that segment is displayed), this needs to be explained a bit more, both here in the code-comments as well as in the GUI page.

meteor/client/ui/Shelf/TimelineDashboardPanel.tsx
thomasslee97: Because `matchFilter` is run in a loop where each call operates on a single item. Also, in e.g. the `AdLibPanel` we are passing the same set to the `renderRundownAdLibs` and `renderSegments` functions to handle the case where two equivalent adlibs are created in both the baseline and a segment, so we need a set which lives long enough.

meteor/client/ui/Shelf/AdLibPanel.tsx
thomasslee97: I've added this note to the settings 
image

As well as expanding that comment, and I've added doc comments to the `matchFilter` function https://github.com/olzzon/tv-automation-server-core/commit/8be6b082af0361d49f938d349d262bbf9536eae9

meteor/client/ui/Shelf/AdLibPanel.tsx
thomasslee97: Removed https://github.com/olzzon/tv-automation-server-core/commit/4e31a4cf8a5decd3a9ba27b40f549b6f6ecdda9b

meteor/client/ui/Shelf/AdLibPanel.tsx
thomasslee97: I'll put that down to merge error/bad commit. I've restored that check https://github.com/olzzon/tv-automation-server-core/commit/92b4179541d94d3f7f67ab13353ef409f57af601

meteor/client/ui/Shelf/TimelineDashboardPanel.tsx
jstarpl: Ha, yeah, you're totally correct! I completely missed the fact that this needs to be consistent across the entire filter invocation. Sorry!

jstarpl (approved):

PR #493, opened 2021-04-01
chore: lerna precommit hook

jstarpl:

This is a workflow bugfix where the husky linting precommit hook prevented commits because the linting would fail.

If making changes in `packages`, the husky precommit hook will trigger and will fail because the root `package.json` doesn't contain any `lint-fix` task.

Based on [#1](https://github.com/lerna/lerna/tree/main/core/filter-options) [#2](https://github.com/typicode/husky/issues/624) [#3](https://medium.com/rewrite-tech/how-to-create-a-monorepo-with-lerna-3ed6dfec5021) and [#4](https://dev.to/try_catch/top-5-handy-lerna-flags-for-your-monorepo-10fa), this PR moves the precommit lint-staged tasks into each of the packages and then uses lerna to run them for the packages that have been changed since HEAD.

* **Other information**:

**Status**

### DIFF
--- a/packages/blueprints-integration/package.json
+++ b/packages/blueprints-integration/package.json
@@ -18,6 +18,7 @@
 		"build": "rimraf dist && yarn build:main",
 		"build:main": "tsc -p tsconfig.build.json",
 		"lint": "eslint . --ext .ts --ext .js --ignore-pattern dist",
+		"lint-fix": "yarn lint --fix",
 		"unit": "jest",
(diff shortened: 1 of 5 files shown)


Julusian (approved):
server-core-integration and mos-gateway arent yet using prettier, so their lint-staged configs are technically wrong, but it is unlikely to cause any issues due to their slow rate of changes

PR #495, opened 2021-04-08
chore(ci): publish comment of prerelease version number

jstarpl:

This is a feature for the CI system.

To find out the version of the pre-release library published using the CI flow, one needs to rummage through GitHub actions logs

GitHub Actions will post a comment on the PR if a pre-release library version was published to NPM.

* **Other information**:

**Status**

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -441,13 +441,24 @@ jobs:
         env:
           CI: true
       - name: Publish to NPM
+        id: publish-npm
         if: ${{ steps.do-publish.outputs.publish }}
         run: |
           echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" >> ~/.npmrc
           cd packages/${{ matrix.package-name }}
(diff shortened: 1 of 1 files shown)


Julusian (approved):
Conceptually I like it, I have a couple of thoughts.

What happens if there is no PR?

This will post a comment for each lib that is published even though we always publish them all. Not a major issue, but is extra noise.
In my ideal flow, I would move the version calculation step to be its own job without the output fed into the prerelease job. Then there could be another job that runs after all of the prerelease jobs to do the comment. 
It is more work, and limited benefit so I dont mind if this isnt done now, we can revisit later

jstarpl:
> What happens if there is no PR?

If there is no PR, the action just quits: https://github.com/mshick/add-pr-comment/blob/07f690343c25a94e24a8acb70d03c86b701ae322/src/main.ts#L141

I agree that probably a single comment would be preferable, but also I'm not that versed in GitHub actions to do something more fancy, especially combining the result of a matrix of actions and then piping them into a single action.

PR #496, opened 2021-04-08
Feat: blueprints static assets

baltedewit:
This feature allows static assets to be included with the blueprint to be used by the Sofie UI. This can be used to preview show-specific content such as wipes.

* Blueprint assets are (small) PNG or SVG files
* Blueprint assets are either included in a json bundle or uploaded seperately through a new REST API point: `blueprints/assets`. There is currently no way of uploading the assets through the UI
* Blueprint assets are stored in the core storage folder, similar to snapshots
* Blueprint assets are served on a new HTTP endpoint: `/blueprints/assets/...`, the folder structure from the blueprint webpack bundler is respected

Additional thoughts:
 * There is no organisation access control (no one appears to be using that currently either)
 * There is a breaking change in the payload of the json bundles, which could be prevented by using a new (v2) API endpoint or by treating the `assets` index as a magical keyword both of which have their own downsides
 * Assets are uploaded directly into the core storage folder, perhaps putting them in a subfolder would be nicer? Doesn't seem high priority though
 * No thought has been given to the lifetime of assets, but because webpack seems to be consistent in the hashed filenames I feel like it won't be a big problem.
 * The mechanism for serving the assets feels a bit thrown together, but not unsimilar to how snapshots are served either. Perhaps adding a subfolder in the core storage means we could use a default static file serving method.
 * There's probably some limit to how much data should be uploaded through this mechanism at once but as long as it's mainly used for icons and previews I don't think that will be a problem either.

**Status**


sparkpunkd (commented):
I guess this could work? But overall I'm not convinced this is a good idea.

The meaning of POST needs to be clarified ... is it create new asset collection / replace all assets. Or create new asset collection / merge with existing asset collection? I'm be more of a fan of PUTing each file and forcing a check of the mimetype.

I've not seen BASE64 encoded strings in JSON documents used before as a binary data format for upload before ... but I guess its a modernisation of multi-part encoded MIME messages?

What happens about names clashes? It is possible that 2 sets of blueprints - e.g. sports and news - or national and regional - could end up referencing an asset with the same name? That needs to be different? (Maybe this is not a new problem.) 

Seems like a potential security risk? Especially uploading to the same folder as other assets. Plus SVG files can contain arbitrary Javascript - although the SVG file shouldn't end up running on the server. At least the Blueprints execute within a sandbox. 

Finally, if the main purpose of this is to serve specific types of hoverscrubs to the GUI, we already have a mechanism for doing this through media manager that will be enhanced by package manager. Why create the extra maintenance overhead?

sparkpunkd (approved):
Following a discussion and in the spirit of _just do it_, this is an improvement over the showstyle assets being fixed or slid in around the side. Approved. The API has shortcomings, but then so does the existing blueprints upload API (e.g. name clashes, what if 2 systems try and upload at the same time etc.)

Also, on the assumption that the only way the systems can "run" the SVG is in an image tab on the browser, this is not a security risk.

If Sofie is ever deployed outside of a firewall, e.g. on public Internet servers, this API will have to be re-examanied as part of a broader review.

PR #497, opened 2021-04-08
feat: move start scripts to a root package.json

Julusian:

Workflow update

Running `meteor npm install` on windows will often fail due to meteor manglign the PATH variable

Instead of triggering the packages commands via meteor, they should all be done via scripts in the package.json at the root level. This avoids meteor mangling the paths, while keeping the single command to prepare and start meteor

* **Other information**:

**Status**


baltedewit (approved):
More consistent to use yarn instead of npm and less going in and out of directories make this a good improvement 👍

PR #500, opened 2021-04-14
Fix/migrations

nytamin:
This PR fixes some issues with the migrations

* fix: issue with ensureCollectionProperty which didn't work when the value was set to "null"
* fix: don't set CURRENT_SYSTEM_VERSION to pre-release version

**Status**


sparkpunkd (approved):
Good spot on the null values and the comment on CURRENT_SYSTEM_VERSION ... I'll also clarify in the release procedures.

PR #503, opened 2021-04-20
feat: migrations to ensure the playout and mos gateways are matching the version of core

Julusian:

feature/workflow improvement

(For release34)

For each release a migration is added to update the expected version of the playout and mos gateways

Now that the gateways follow the same versioning scheme as core, a migration has been added to update the expected version to match the current version of core. This means that one doesnt have to be added for each release.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/migration/currentSystemVersion.ts
+++ b/meteor/server/migration/currentSystemVersion.ts
@@ -35,4 +35,4 @@
  */
 
 // Note: Only set this to release versions, (ie X.Y.Z), not pre-releases (ie X.Y.Z-0-pre-release)
-export const CURRENT_SYSTEM_VERSION = '1.32.0'
+export const CURRENT_SYSTEM_VERSION = '1.33.0'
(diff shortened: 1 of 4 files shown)


meteor/server/migration/currentSystemVersion.ts
```diff
@@ -35,4 +35,4 @@
  */
 
 // Note: Only set this to release versions, (ie X.Y.Z), not pre-releases (ie X.Y.Z-0-pre-release)
-export const CURRENT_SYSTEM_VERSION = '1.32.0'
+export const CURRENT_SYSTEM_VERSION = '1.33.0'
```
nytamin, line 38: (This change should ideally not be in this PR)

nytamin (approved):
Good idea!

meteor/server/migration/currentSystemVersion.ts
Julusian: true. I forgot to commit it separately, as the current value is wrong.
It got messed up in a previous merge so this may as well be the thing that corrects it now

PR #507, opened 2021-04-22
fix: only reset previous segment on looping rundowns

jstarpl:

This is a bugfix.

After playback of a segment ends, PartInstances inside are reset. This causes problems for timing in the GUI as reset PartInstances are not sent to the GUI and are not taken under consideration for timing purposes.

PartInstances inside segments that stop playing are only reset if the playlist is looping.

* **Other information**:

**Status**

### DIFF
--- a/meteor/package-lock.json
+++ b/meteor/package-lock.json
@@ -14830,7 +14830,8 @@
 		"y18n": {
 			"version": "4.0.1",
 			"resolved": "https://registry.npmjs.org/y18n/-/y18n-4.0.1.tgz",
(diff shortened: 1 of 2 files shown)


Julusian (approved):
I have no problem with this short-term, but we should revisit it in future release. 
I know that it has been discussed a few times to reset previous segments to minimise the amount of data the ui has to keep loaded, as well as to show the user the planned parts to aid in re-playing segments

jstarpl:
> 
> 
> I have no problem with this short-term, but we should revisit it in future release.
> I know that it has been discussed a few times to reset previous segments to minimise the amount of data the ui has to keep loaded, as well as to show the user the planned parts to aid in re-playing segments

Yes, let's revisit it fully at some point. With the current data structure, the PartInstances are the only source of timing information the GUI can use to calculate stuff. If we want to drop them, we need to replace them with something else.

PR #509, opened 2021-04-23
Chore/various tests

nytamin:
This PR contains various things that I found in a local branch from November 2020, which I've now rebased upon release33.

What it contains:
* A mock of the NTP-client, so that we can run the NTP-tests reliably
* Fixes to be able to `setupDefaultStudioEnvironment` with a organizationId, so that we can run tests in an organization-environment.
* fix: snapshots: Don't access file system when in unit tests, to avoid non-test-related issues

**Status**

### DIFF
--- a/meteor/__mocks__/helpers/database.ts
+++ b/meteor/__mocks__/helpers/database.ts
@@ -55,6 +55,7 @@ import {
 	defaultAdLibPiece,
 	defaultStudio,
 } from '../defaultCollectionObjects'
+import { OrganizationId } from '../../lib/collections/Organization'
 
 export enum LAYER_IDS {
 	SOURCE_CAM0 = 'cam0',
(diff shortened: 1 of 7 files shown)


meteor/__mocks__/helpers/database.ts
```diff
@@ -374,27 +381,33 @@ export interface DefaultEnvironment {
 
 	ingestDevice: PeripheralDevice
 }
-export function setupDefaultStudioEnvironment(): DefaultEnvironment {
+export function setupDefaultStudioEnvironment(organizationId?: OrganizationId | null): DefaultEnvironment {
```
Julusian, line 384: I think you can write these two lines as:
```
export function setupDefaultStudioEnvironment(organizationId: OrganizationId | null = null): DefaultEnvironment {
```
and it will have the same effect, and make the default more obvious

Julusian (approved):

PR #512, opened 2021-04-29
feat: untimed parts

jstarpl:

This is a new feature.

All playable parts are counted towards show time and as-played time - unless they are "invalid" or "floated", which renders them unplayable.

A Part can be marked as "untimed" which will keep them playable, but cause them to:
* not trigger blueprints "firstTake" callback
* not set `startedPlayback` on a playlist when they are taken, the `startedPlayback` will be set once a not-"untimed" part is taken
* not count towards segment time budget
* there will be a countdown to the end of an "untimed" part (unless it's `expectedDuration` is 0) and the "On Air In" counter will also be affected by the duration of an "untimed" part.

* **Other information**:

**Status**


Julusian (approved):
It looks like the events used for metadatabank will still be called (onRundownTimingEvent), but that could be appropriate as things did happen and the blueprints can easily filter out the parts that are untimed if it desired, so I am happy with this

Other than that it looks all good to me

PR #513, opened 2021-05-07
chore: add asserNever

nytamin:
This PR adds a few assertNever in various places.

### DIFF
--- a/meteor/client/lib/SettingsNavigation.tsx
+++ b/meteor/client/lib/SettingsNavigation.tsx
@@ -19,7 +19,7 @@ export class SettingsNavigation extends React.Component<ISettingsNavigation> {
 			return <ShowStyle {...this.props} />
 		} else if (this.props.type === 'newshowstyle') {
 			return <NewShowStyle {...this.props} />
-		} else{
+		} else {
 			assertNever(this.props.type)
 		}
 
--- a/meteor/client/lib/SettingsNavigation.tsx
+++ b/meteor/client/lib/SettingsNavigation.tsx
@@ -5,6 +5,7 @@ import { Mongo } from 'meteor/mongo'
 import { withTranslation } from 'react-i18next'
 import { Studios } from '../../lib/collections/Studios'
 import { MeteorCall } from '../../lib/api/methods'
+import { assertNever } from '../../lib/lib'
 
 interface ISettingsNavigation extends ISettingsNavigationBaseProps {
 	type: SettingsNavigationType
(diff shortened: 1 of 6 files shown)


meteor/client/ui/Settings/components/ConfigManifestEntryComponent.tsx
```diff
@@ -61,6 +62,14 @@ export const renderEditAttribute = (
 				}
 			/>
 		)
+	// } else if (configField.type === ConfigManifestEntryType.TABLE) {
```
nytamin, line 65: I've commented out these changes, as I wasn't sure on the best ways to address these

meteor/lib/collections/PeripheralDevices.ts
```diff
@@ -145,6 +145,15 @@ export function getExternalNRCSName(device: PeripheralDevice | undefined): strin
...
 				return 'Google Sheet'
+			} else if (
+				device.type === PeripheralDeviceAPI.DeviceType.PLAYOUT  ||
+				device.type === PeripheralDeviceAPI.DeviceType.MEDIA_MANAGER ||
+				device.type === PeripheralDeviceAPI.DeviceType.PACKAGE_MANAGER
+			) {
+				// These aren't ingest gateways
+				return 'N/A'
+			} else {
+				assertNever(device.type)
```
jstarpl, line 156: Considering that it's possible to connect to Sofie using a custom peripheralDevice, even if the typeguard is useful, it should still try and return some useful value, perhaps `Unknown: ${device.type}`?

meteor/client/lib/SettingsNavigation.tsx
```diff
@@ -18,6 +19,8 @@ export class SettingsNavigation extends React.Component<ISettingsNavigation> {
 			return <ShowStyle {...this.props} />
 		} else if (this.props.type === 'newshowstyle') {
 			return <NewShowStyle {...this.props} />
+		} else{
```
jstarpl, line 22: Linting error?

jstarpl (approved):
I really like what you did, one comment would be to think where it's actually physically possible that the type guard is hit in runtime (like when the data is provided by an external system) and either throw an error or provide a sensible fallback.

meteor/lib/collections/PeripheralDevices.ts
nytamin: Good insight! Will fix!

PR #515, opened 2021-05-07
fix: Update playlist after removing rundown

Julusian:

Fix

When a rundown is removed from a playlist, the playlist is not regenerated.
This means that the name, timing or any other properties can refer to the recently removed rundown.

Whenever a rundown is removed from a playlist (or moved out of a playlist), that playlist gets regenerated in the same way as when we add/update a rundown in the playlist.
If the ordering is calculated by the blueprints, the order of the remaining rundowns will be updated too.

* **Other information**:

I am unsure if this is targetting the correct release, it might want to be delayed to r34/r35 depending on how criticial the bug is

**Status**


jstarpl (approved):
Looks sensible 👍

PR #526, opened 2021-05-24
feat/standardization

Julusian:

Typescript is configured to be loose, and so does not report on a lot of potentially unsafe checks.
Tslint is being used instead of eslint. tslint is deprecated and is also configured to be quite loose.

Typescript has been tightened up a bit. There are still some rules which should be enabled, but doing so requires many code changes, and is being left for later to minimise the pain of this PR.
Eslint has been setup, also with some rules disabled for now. Again to be improved on later.

While this has pulled in @sofie-automation/code-standard-preset, it is not being used for much. This is because we need to extend the meteor typescript config to avoid issues there. And eslint needs to be configured very differently due to the paths differences, and presence of react. Perhaps some of this could be ported back as an alternate config, but for now it is simpler to keep it just in this repo

* **Other information**:

I tried to minimise the changes made, but a prettier update was forced upon us through @sofie-automation/code-standard-preset.
Some tsc/eslint errors were fixed, but only ones where there werent many occurances 

**Status**


sparkpunkd (commented):
Lint error fixed. Approved - obviously! Great to see this make it to core :-)

PR #527, opened 2021-05-28
Feat: Showstyle Infinites

thomasslee97:

Adds a new piece lifespan: `OnShowStyleEnd`.

New feature.

Adds `OnShowStyleEnd` infinites which exist across multiple rundowns in a playlist. A piece with this lifespan will continue to exist for as long as the next rundown shares the same showstyle base as the current rundown. Changes of showstyle variant will not stop the infinite. All other infinite behaviour in terms of stopping/clearing pieces should behave the same as other `On[...]End` infintes.

* **Other information**:

**Status**


meteor/client/ui/RundownView.tsx
```diff
@@ -2261,10 +2268,13 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
 												rundownId={rundownAndSegments.rundown._id}
 												segmentId={segment._id}
 												playlist={this.props.playlist}
+												rundown={rundownAndSegments.rundown}
 												timeScale={this.state.timeScale}
 												onContextMenu={this.onContextMenu}
 												onSegmentScroll={this.onSegmentScroll}
 												segmentsIdsBefore={rundownAndSegments.segmentIdsBeforeEachSegment[segmentIndex]}
+												rundownIdsBefore={new Set(rundownArray.slice(0, rundownIndex).map((r) => r.rundown._id))}
```
Julusian, line 2276: This wont be very performant. At the very least this set could be calculated once at the top of the loop on line 2243.
Ideally this is something that should be memoized to help avoid unnecessary re-renders, but I dont know if that will fit into the code well

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -375,13 +384,29 @@ export const SegmentTimelineContainer = translateWithTracker<IProps, IState, ITr
...
-						startSegmentId: { $in: Array.from(this.props.segmentsIdsBefore.values()) },
 						invalid: {
 							$ne: true,
 						},
-						// same rundown, and previous segment
-						lifespan: { $in: [PieceLifespan.OutOnRundownEnd, PieceLifespan.OutOnRundownChange] },
+						$or: [
+							// same rundown, and previous segment
+							{
+								startSegmentId: { $in: Array.from(this.props.segmentsIdsBefore.values()) },
```
Julusian, line 393: I think this part of the query could do with a startRundownId condition too, to help scope it down

meteor/lib/rundown/infinites.ts
```diff
@@ -50,14 +53,24 @@ export function buildPastInfinitePiecesForThisPartQuery(
 			},
 			{
 				// same rundown, and previous segment
-				lifespan: { $in: [PieceLifespan.OutOnRundownEnd, PieceLifespan.OutOnRundownChange] },
+				lifespan: {
+					$in: [
+						PieceLifespan.OutOnRundownEnd,
+						PieceLifespan.OutOnRundownChange,
+						PieceLifespan.OutOnRundownChange,
```
Julusian, line 60: duplicate line

meteor/lib/rundown/infinites.ts
```diff
@@ -153,6 +177,33 @@ export function getPlayheadTrackingInfinitesForPart(
...
+								isValid = false
+								break
+							}
+
+							if (candidatePiece.rundownId === part.rundownId) {
+								isValid = true
+								break
+							}
+
+							const rundownsBeforeThisInPlaylist = Array.from(rundownsBeforeThisInPlaylistSet.values())
```
Julusian, line 191: I believe that converting a set into an array will result in an array with an unpredictable order. Also I dont see how we know that `prevRundownId` is related to `currentPartInstance` or the new instance.
From reading this block of code, it looks like we are checking that `currentPartInstance` is somewhere before the new part, and then checking that the rundown before the new part (the two arent guaranteed to be the same) is of the correct showstyle. 

Q) What happens if we have 4 rundowns?
  Rd1 = 'currentPartInstance', showStyle1
  Rd2 = showstyle2
  Rd3 = showstyle1
  Rd4 = 'newInstance', showstyle1
To me it reads like this will continue infinites across Rd2, even though it is a different style.


I think there are two things to do here.
1) Make `rundownsBeforeThisInPlaylistSet` be an array. Looking back up the call trees, nothing seems to want it as a set, and all usages of the set convert it back to an array.
2) This logic needs some rethinking to check if it can be continued. Which should ideally be done outside of the nested for loops.

meteor/lib/rundown/infinites.ts
```diff
@@ -259,6 +313,25 @@ export function isPiecePotentiallyActiveInPart(
 					segmentsBeforeThisInRundown.has(pieceToCheck.startSegmentId)
 				)
 			}
+		case PieceLifespan.OutOnShowStyleEnd:
+			if (pieceToCheck.startRundownId === part.rundownId) {
+				return true
+			}
+
+			const rundownsBeforeThisInPlaylist = Array.from(rundownsBeforeThisInPlaylistSet.values())
```
Julusian, line 321: Same concerns with this block as above

meteor/server/api/ingest/syncChangesToPartInstance.ts
```diff
@@ -94,6 +94,7 @@ export function syncChangesToPartInstances(
 				const proposedPieceInstances = getPieceInstancesForPart(
 					cache,
 					previousPartInstance,
+					rundown as Rundown,
```
Julusian, line 97: please propogate the `ReadonlyDeep<Rundown>` through the other methods instead of casting here.
So far I've been lazily changing the typings as needed, so expect to need to change places

meteor/server/api/playout/infinites.ts
```diff
@@ -105,21 +120,42 @@ export async function fetchPiecesThatMayBeActiveForPart(
...
+	} = getIdsBeforeThisPart(cache, part)
 
 	const infinitePiecesQuery = buildPastInfinitePiecesForThisPartQuery(
 		part,
 		partsBeforeThisInSegment,
-		segmentsBeforeThisInRundown
+		segmentsBeforeThisInRundown,
+		rundownsBeforeThisInPlaylist
 	)
 	// Future scope: Once there is a longer than Rundown infinite mode, this will need to split the query to search everywhere
```
Julusian, line 135: update this comment to reflect that this is now bein done

Julusian (changes requested):
I have a few concerns about the correctness of some bits, but the bulk of it looks good

meteor/lib/rundown/infinites.ts
thomasslee97: I've made a change to look at all rundowns between the current part and the target part to check that they all match the same showstyle, as well as changing the set to an array.

meteor/lib/rundown/infinites.ts
```diff
@@ -271,6 +253,16 @@ export function isPiecePotentiallyActiveInPart(
 		return true
 	}
 
+	const canContinueShowStyleEndInfinties =
```
Julusian, line 256: as this is only read in oneplace, it would make sense to be done there

meteor/client/ui/RundownView.tsx
```diff
@@ -2239,74 +2239,78 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
...
+													followLiveSegments={this.state.followLiveSegments}
+													rundownId={rundownAndSegments.rundown._id}
+													segmentId={segment._id}
+													playlist={this.props.playlist}
+													rundown={rundownAndSegments.rundown}
+													timeScale={this.state.timeScale}
+													onContextMenu={this.onContextMenu}
+													onSegmentScroll={this.onSegmentScroll}
+													segmentsIdsBefore={rundownAndSegments.segmentIdsBeforeEachSegment[segmentIndex]}
+													rundownIdsBefore={new Set(rundownIdsBefore)}
```
Julusian, line 2279: I suspect this shouldnt be a set anymore

Julusian (approved):
I think this is looking good now, just a couple of very minor non-functional comments

PR #530, opened 2021-06-07
feat: speed and re-rendering optimisations

olzzon:
Optimizing Rendering of parts, with focus on rounding errors, and durationSettling after a take
Speed optimization - "now" time is rounded for 25fps

When take the new part stops for a second

Registering take and use "now" instantly as the new take.

* **Other information**:

**Status**


meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -121,6 +121,9 @@ export const RundownTimingProvider = withTracker<
 		private temporaryPartInstances: Map<PartId, PartInstance> = new Map<PartId, PartInstance>()
 
 		private linearParts: Array<[PartId, number | null]> = []
+		private prevPartId: string = ''
```
jstarpl, line 124: Why is this an empty string? This is only ever compared with PartInstanceIds, so I would expect this should be `private prevPartId = PartInstanceId | null = null`. Please add a comment what this is used for and why.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -412,7 +419,22 @@ export const RundownTimingProvider = withTracker<
 					this.partDisplayDurations[partInstancePartId] = partDisplayDuration
 					this.partDisplayDurationsNoPlayback[partInstancePartId] = partDisplayDurationNoPlayback
 					startsAtAccumulator += this.partDurations[partInstancePartId]
-					displayStartsAtAccumulator += this.partDisplayDurations[partInstancePartId] // || this.props.defaultDuration || 3000
+
+					if (playlist.previousPartInstanceId !== partInstance._id) {
+						displayStartsAtAccumulator += this.partDisplayDurations[partInstancePartId]
+					} else {
+						if (this.prevPartId !== unprotectString(playlist.previousPartInstanceId)) {
```
jstarpl, line 426: See the comment on line 124. There isn't any reason to unprotect these Ids and lose type safety.

meteor/client/ui/SegmentTimeline/SegmentTimeline.tsx
```diff
@@ -156,6 +156,10 @@ const SegmentTimelineZoom = class SegmentTimelineZoom extends React.Component<
 		return this.props.isLiveSegment ? this.calculateSegmentDuration() : this.state.totalSegmentDuration
 	}
 
+	calcTimeScale = (time: number) => {
```
jstarpl, line 159: Perhaps this could be renamed to `convertTimeToPixels` or something like that?

jstarpl:
Looking very interesting. Haven't had time to do a proper review, but I've made some quick code quality comments.

meteor/client/ui/SegmentTimeline/SegmentTimeline.tsx
olzzon: Yes thats a better name 👍

olzzon:
@jstarpl also look into the calmDownTiming used in onRefreshTimer, right now it´s set to 40ms (1 * 25fps) but i think the best performance contra GUI updates is around 70ms. To see what it does, try setting it at 1000ms

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
olzzon: If not i get this error: (i actually just copied this from line 414 where the same way is done on partInstance.part._id:
```
414    const partInstancePartId = unprotectString(partInstance.part._id)
```

image

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
olzzon: In the first setup i did the string I compared it with could be undefined, so i ended up giving it a value.
But as the first comparison now always will be with a string, | null = null is way better 👍

olzzon: Has been fixed :)

meteor/client/ui/SegmentTimeline/SegmentTimeline.tsx
olzzon: Has been fixed :)

olzzon:
The conflict in SegmentTimelineContainer.tsx is because i´ve removed the playbacksimulation, as it gets the new "take" imidiately.

```

<<<<<<< r35/timing-rendering-optimize
=======
				const simulationPercentage = this.playbackSimulationPercentage
>>>>>>> release36
```

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
jstarpl: Well, sure, because you're using a string and not the protectedString ID type. You should change the type of `prevPartId` to be `PartInstanceId | null` (since either it's going to be an ID or will be null) and then you can safely do the comparison without any unprotecting.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -412,7 +419,22 @@ export const RundownTimingProvider = withTracker<
...
+							this.lastTakeAt = now
+							this.prevPartId = unprotectString(playlist.previousPartInstanceId) || ''
```
jstarpl, line 428: I didn't notice this one before, but if you change `prevPartId` you can remove the `unprotectString` here as well and you'll end up with a nice `this.prevPartId = playlist.previousPartInstanceId`, since `playlist.previousPartInstanceId` is also a `PartInstanceId | null` type.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -121,6 +121,9 @@ export const RundownTimingProvider = withTracker<
 		private temporaryPartInstances: Map<PartId, PartInstance> = new Map<PartId, PartInstance>()
 
 		private linearParts: Array<[PartId, number | null]> = []
+		private prevPartId: string | null = null
```
jstarpl, line 124: Some sort of a comment about what this value will be used for would be good, since "prevPartId" could mean a lot of things.

jstarpl (changes requested):
I'm still going to insist on changing the `prevPartId` type from `string | null` to `PartInstanceId | null`, as that makes the comparisons type-safe.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -145,8 +151,12 @@ export const RundownTimingProvider = withTracker<
 			}
 		}
 
+		calmDownTiming = (time: number) => {
+			return Math.round(time / 40) * 40
```
jstarpl, line 155: Maybe the `40` could be moved to a constant somewhere around line 20, how about

```
const CURRENT_TIME_GRANULARITY = 40
```
? (or 66, 67 or 70 or whatever we decide)

jstarpl (changes requested):
> jstarpl also look into the calmDownTiming used in onRefreshTimer, right now it´s set to 40ms (1 * 25fps) but i think the best performance contra GUI updates is around 70ms. To see what it does, try setting it at 1000ms

Yeah, I suspect the best value would be something like 66/67ms, which would give us roughly 15fps. I wonder if the `CURRENT_TIME_GRANULARITY` should just be the same as `TIMING_DEFAULT_REFRESH_INTERVAL`, since there is little point in running the timing calculations if the timing source moves less frequently than the calculation interval.

Overall, a very good effort & it increases code readability by making the time-to-pixels conversion explicit.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
olzzon: Agree, I will add this 👍

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -798,27 +786,12 @@ export const SegmentTimelineContainer = translateWithTracker<IProps, IState, ITr
...
-						if (lastTake < lastStartedPlayback && simulationPercentage < 1) {
-							virtualStartedPlayback =
-								simulationPercentage * lastStartedPlayback + (1 - simulationPercentage) * lastTake
-						}
-					}
-				}
 
-				let newLivePosition =
-					isExpectedToPlay && virtualStartedPlayback
+				let newLivePosition =
```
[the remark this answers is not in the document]
olzzon: Then I i should reinsert Simulated Playback Margin and make that work together.

Regarding the jiggling. I think this comes down to different places of the codes that does the same type of calculations.
I think we could improve this a lot by moving calculations into RundownTimingProvider and use a get. Especially getScrollLeft() as this.scrollLeft seems to be quite unstable. We might even end up with some more streamlined props.
Again i'm not that deep into the code yet, so i could be wrong 🔢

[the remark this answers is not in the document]
olzzon: I have reinserted isExpectedToPlay and had to raise SIMULATED_PLAYBACK_HARD_MARGIN to 3500 to avoid an early jump. (one of the jumps we've seen)

jstarpl (approved):
Great, let's get it merged!

PR #532, opened 2021-06-11
fix(time formatting): Future dates are displayed as past in GUI

nytamin:
Bug fix

Future dates are displayed as past in GUI. ("a year ago")

Future dates are displayed as future in GUI. ("in a year")

* **Other information**:

**Status**


meteor/jest.config.js
```diff
@@ -19,7 +19,7 @@ const commonConfig = {
 			},
 		},
 	},
-	moduleFileExtensions: ['ts', 'js'],
+	moduleFileExtensions: ['ts', 'js', 'tsx', 'jsx'],
```
jstarpl, line 22: Do we need to include TSX and JSX in `moduleFileExtensions`? We currently have no way of unit testing React code, and all JSX we use is React.

jstarpl (approved):
Looks good, the only question I have is about adding TSX and JSX to module files for Jest - what's the point, given that we currently have no facilities for Unit-testing React components.

meteor/jest.config.js
nytamin: Good point! I added that while trying to get Jest to run a tsx-file, but I ended up splitting the file into a pure ts-file instead, and mistakenly left it there.
I'll revert it, since it's not used anywhere.

PR #533, opened 2021-06-11
fix: prompter doesn't use changes from part/piece instances

jstarpl:

Bug fix

Prompter ignores data coming from Part and Piece instances. As a result, the state of the Rundown as displayed in the RundownView and the state of the Rundown in Prompter may get out of sync.

The prompter will display the show Rundown exactly the same way the Rundown View does. This should maintain consistency across all outputs.

* **Other information**:

**Status**


nytamin (approved):
LGTM!

PR #536, opened 2021-06-14
feat: dismiss version warnings

Julusian:

feature

Expected versions of gateways are managed by 'expectedVersion' migrations. Since r34 these have been automated as the version numbers match core. But this has required an exact version match

Blueprints are required to match the blueprint-integration version of core, including patch number. Which results in needing to do a blueprint release after every core release.

Gateways now always expose the version of their core-integration library, and that version is used to check if they are compatible.
Both blueprints and gateways check this with a looser rule:
* if core is a prerelease build, then a prerelease version is accepted
* otherwise they must be of the same minor version such as `1.35` and not a prerelease

Additionally, for each gateway and blueprint the ui has a new 'Disable version check' field. When this is ticked, core will skip the version checks for that version.
Every time a blueprint has its code updated, and everytime a gateway updates its version numbers, the property for that blueprint/device gets reset back to false to ensure that each time the version changes it has to be re-verified

image

* **Other information**:

**Status**


sparkpunkd (approved):
A common sense approach to version checking given the new core repo structure.

PR #541, opened 2021-06-17
feat: Remove delayed initialisation from DbCacheReadObject and DbCacheReadCollection

Julusian:

code cleanup

The DbCacheReadObject and DbCacheReadCollection support being lazy initialised.
We only make use of this for a couple of collections, and supporting it makes it hard to understand whether an operation on the cache collection will yield the fiber or not.

Additionally, for the cases where we were lazy initialising, by knowing we might be doing so, we can optimise by running the loads in parallel. Currently we would be loading them one at a time, as the code was not aware they would need loading

The collections no longer support being lazy initialised. Instead they should be wrapped in the LazyInitialise wrapper. This makes the api slightly less clean to do the operations, as it requires an explicit await to get a reference to the collection, but this makes it explicit that it may be an async operation.

* **Other information**:

**Status**

### DIFF
--- a/meteor/package-lock.json
+++ b/meteor/package-lock.json
@@ -14460,6 +14460,11 @@
 			"integrity": "sha1-P7z7FbiZpEEjs0ttzBi3JDNqLK4=",
 			"dev": true
 		},
+		"p-lazy": {
+			"version": "3.1.0",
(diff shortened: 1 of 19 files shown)


Julusian:
As discussed with @nytamin I am merging this now and a proper review will be done next week

PR #543, opened 2021-06-25
feat: rewrite syncFunction to be promise and p-queue based

Julusian:

code quality improvement. 

syncFunction is used as a form of work queue. It is very centered around fibers

syncFunction has been replaced by `pushWorkToQueue` which operates in a similar fashion, but internally uses the `p-queue` library instead of a custom written queue.
Functionally it should be no different, but is written to be promise based first and uses the helpers to handle fibers when necessary instead.

As part of this, all of the various lockFunctions/syncFunctions for specific purposes have been made to be promise based too, as part of the 'reduce fibers' push

* **Other information**:

Overall, the change is mostly in the few files which define the syncFunction and lockFunctions, but many other files have to add a few async/await keywords to functions to make the flow be the same as before.
Many api methods are now promises all the way as a consequeunce.

**Status**


meteor/client/ui/Account/NotLoggedIn/LostPassword.tsx
```diff
@@ -57,7 +57,7 @@ export const LostPasswordPage = translateWithTracker((props: ILostPasswordPagePr
 			const { t } = this.props
 			return (
 				<NotLoggedInContainer>
-					<form onSubmit={(e: React.MouseEvent<HTMLFormElement>) => this.resetPassword(e)}>
+					<form onSubmit={async (e: React.MouseEvent<HTMLFormElement>) => this.resetPassword(e)}>
```
jstarpl, line 60: This one seems wrong. Browser event handlers can't be async, because Event obejcts aren't built for async. Specifically, `.resetPassword()` shouldn't be async, because it uses e.preventDefault(), which needs to run synchronously to do anything.

jstarpl (changes requested):
I really like the change, the only comment I have is that DOM event handlers need to be synchronous (even though the typings apparently allow returning a promise), since they can't function asynchroniously, at least as far as I know. The event handler needs to handle the promise, since the browser has no way of handling it.

meteor/client/ui/Account/NotLoggedIn/LostPassword.tsx
Julusian: Good spot. I honestly didnt look over the async linter rule changes very much, I wouldnt have known this to be bad.
`resetPassword()` has itself been async for well over a year, so nothing really changed from this in this particular case, but there could well be other similar ones.

I am wondering if the solution is to add more linter rules. specifically:
* https://github.com/github/eslint-plugin-github/blob/main/docs/rules/async-currenttarget.md
* https://github.com/github/eslint-plugin-github/blob/main/docs/rules/async-preventdefault.md
By the look of it, they will flag the issue with resetPassword, not with the handler explicitly. 

Do you think that will be sufficient? Otherwise I could turn off the new linter rule for the client code, at the risk of the original issue still being possible to occur there

jstarpl: Hmm... I'm not sure if these linter rules will work and highlight any errors in our case. See, `onSubmit={async () => something}`, isn't equivalent to `element.addEventListener()`. React does it's own event handling and accessing any properties of the event outside of the task that triggred the event handler will result in undefined behavior, since Event objects are recycled in React. I think the only "safe" rule for eslint would be to prevent the use of the Event object in any async handlers. Plus the esLint rule would need to understand JSX event handling syntax, which I'm not sure these two rules do.

jstarpl: I'm not sure if that's going to work. First, I expect that these rules only apply to `element.addEventHandler`, since React's `onEvent={}` isn't at all equivalent to `addEventHandler`. As a matter of fact, accessing any Event properties outside of the task that handles the event will result in undefined behavior since React's synthetic events are recycled (unless an Event object is persisted(), which probably complicates the matter more).  I haven't been able to quickly find any eslint plugins that provide a rule to disallow use of Event objects within async handlers.

Julusian: If you look at the definition of the rules, they are surprisingly short https://github.com/github/eslint-plugin-github/blob/main/lib/rules/async-currenttarget.js
It looks to me like it is looking for use of `.currentTarget` inside an async function, if an await call has been made earlier in that function. So I think it should catch our case, but I cant be sure without testing it

But I would like to get this merged soon, so I shall disable this rule and undo the changes for files in `/client/` to avoid breaking things, and we can consider enabling there again later on

Julusian: > The body of an async function can be thought of as being split by zero or more await expressions. Top-level code, up to and including the first await expression (if there is one), is run synchronously. In this way, an async function without an await expression will run synchronously.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

So using the event inside an async can be ok, but that doesnt change that the result with be a promise. So by adding async it doesnt change any functionality, but it does make it explicit that there is a promise there and so to be careful with the event lifetime.

Ive disabled the rule for the tsx files for now

jstarpl (approved):

PR #552, opened 2021-08-05
fix(Playout Gateway): add missing TSR device options and layer mapping options

jstarpl:

This is a fix

Options for OBS and Singular devices are missing and these cannot be properly configured. Also, some layer mapping configurations for some devices were missing.

The missing entries in the Peripheral Device settings manifest were added and some TypeScript sugar was added as well to at least ensure that when new devices are added to TSR, we will need to add some configuration entries for them for the build to succeed, hopefully making sure that this doesn't happen in the future.

* **Other information**:

**Status**


Julusian (approved):
Looks good
I like the extra type safety introduced to ensure this doesnt happen again :)

PR #553, opened 2021-08-06
feat: Action Triggers

jstarpl:

This is a new feature

Currently, certain actions have hard-coded hotkeys to trigger them and others can be set using the SourceLayer properties, using comma-separated lists of hotkeys activating AdLibs, clearing source layers and replaying sticky AdLibs.

The new Action Triggers mechanism replaces that with a more user-friendly system of Triggers (like hotkeys, but the system also allows for server-side type triggers like GPI, MIDI or OSC) that cause Actions to happen. These Actions include: 'Set Shelf state to X' or 'Activate Rundown Playlist' or 'Execute AdLib Action'. Some of these Actions accept "Filters" that allow the User (or the Blueprint Developer) to specify exactly what Action should be performed. The Filtering system is somewhat similar to the Shelf Filters/Panels system.

Demo

* **Other information**:

A follow-up PR will need to remove/rename/retype hotkey properties on ISourceLayer since they will no longer serve much purpose. `clearKeyboardHotkey` needs to be renamed to something like `isClearable`, and the rest can just be removed. I've decided to leave that clean-up off this PR, so that it doesn't become too messy.

**Status**


meteor/lib/collections/TriggeredActions.ts
```diff
@@ -0,0 +1,40 @@
...
+	_rundownVersionHash: string
+}
+
+export type TriggeredActionsObj = DBTriggeredActions
+export const TriggeredActions = createMongoCollection<TriggeredActionsObj, DBTriggeredActions>('triggeredActions')
+registerCollection('TriggeredActions', TriggeredActions)
+
+registerIndex(TriggeredActions, {
+	showStyleBaseId: 1,
+	_rank: 1,
```
nytamin, line 33: The `_rank` isnt really needed to be added to the index, since it isn't used anywhere server-side (client-side queries that use $sort doesn't use the index).

meteor/client/lib/triggers/TriggersHandler.tsx
```diff
@@ -0,0 +1,424 @@
...
+		props.rundownPlaylistId,
+		props.currentRundownId,
+		props.currentPartId,
+		props.currentSegmentPartIds,
+		props.nextPartId,
+		props.nextSegmentPartIds,
+	])
+
+	const triggerSubReady = useSubscription(PubSub.triggeredActions, {
+		$or: [
```
nytamin, line 278: Mongodb recommends to do `$in: [null, props.showStyleBaseId]` instead (I suspect it might slightly improve db-performance), see https://docs.mongodb.com/manual/reference/operator/query/or/#-or-versus--in

meteor/server/api/triggeredActions.ts
```diff
@@ -0,0 +1,183 @@
...
+		const triggeredActions = JSON.parse(body) as DBTriggeredActions[]
+		check(triggeredActions, Array)
+
+		// set new showStyleBaseId
+		for (let i = 0; i < triggeredActions.length; i++) {
+			check(triggeredActions[i]._id, String)
+			check(triggeredActions[i].name, Match.Optional(String))
+			check(triggeredActions[i].triggers, Array)
+			check(triggeredActions[i].actions, Array)
+			triggeredActions[i].showStyleBaseId = showStyleBaseId ?? null
```
nytamin, line 79: Just a thought, would it be good to do a specific check for empty-string?
Since it it is falsy, the same way null is, but wont be picked-up when querying mongo for null.

meteor/server/api/triggeredActions.ts
```diff
@@ -0,0 +1,183 @@
...
+			triggeredActions[i]._rundownVersionHash = ''
+		}
+
+		if (replace) {
+			TriggeredActions.remove({
+				showStyleBaseId: showStyleBaseId,
```
nytamin, line 85: should this be `showStyleBaseId ?? null` ?

nytamin (approved):
Oh boy, this is going to be a very welcome addition!
I won't be doing a proper in-detail review due to the size of the PR, I've added a few notes here & there.

Something I think would be very helpful is if we took this opportunity to add some module-documentation.
If you could add a README.md in  the module `client/lib/trigger`, summarizing how it all works, that could be useful.
(ping me if you want some help writing/rubbeducking it :) )

meteor/lib/collections/TriggeredActions.ts
jstarpl: The idea to add _rank here was that, once server-side triggers are implemented, this would be useful for the server-side to sort them, but I guess it can be removed for now.

meteor/client/lib/triggers/TriggersHandler.tsx
jstarpl: Yeah, that's a good note.

meteor/lib/collections/TriggeredActions.ts
nytamin: Ah ok! Makes sense!

meteor/server/api/triggeredActions.ts
jstarpl: I'm adding an undefined check at line in 54, so that any string (as opposed to any falsy value) is treated as a ShowStyleId, so an empty string should cause it to try and find a matching ShowStyleBase and fail with message.

meteor/server/api/triggeredActions.ts
jstarpl: Yeah, good catch, since it can actually be undefined.

packages/blueprints-integration/src/triggers.ts
```diff
@@ -0,0 +1,282 @@
...
+	| IGoToOnAirLineAction
+	| IRewindSegmentsAction
+	| IShowEntireCurrentSegmentAction
+
+export interface IBlueprintTriggeredActions {
+	_id: string
+	/** Rank number for visually ordering the hotkeys */
+	_rank: number
+	/** Optional label to specify what this triggered action is supposed to do, a comment basically */
+	name?: ITranslatableMessage | string
```
Julusian, line 277: I don't think this should be translatable, assuming these are only shown to the person configuring the triggers. It will add complexity to the ui and import/export, for very little benefit.
I am curious to know your thoughts

packages/blueprints-integration/src/migrations.ts
```diff
@@ -129,6 +151,11 @@ export interface MigrationStepCore extends MigrationStep {
 	migrate?: MigrateFunctionCore
```
Julusian, line 151: Note for myself: What are these `*Core` ones used for if they arent for the system blueprint?

Julusian:
Should the old SourceLayer hotkey properties also be removed as part of this? It sounds like they will no longer do anything

I didn't read through much of the UI code, the blueprints api and other backend changes look reasonable

packages/blueprints-integration/src/triggers.ts
jstarpl: No, in some cases (like system-wide/core action triggers), this name will be displayed to the User in the Help panel in the Shelf. So this is the name that will be displayed for triggers like "Take", etc. Most of these will be provided by Core migrations, but the Blueprints can override that.

jstarpl:
> Should the old SourceLayer hotkey properties also be removed as part of this? It sounds like they will no longer do anything

I think they should be removed, but I would like to do that in a separate PR, since that's going to be an involved process on it's own, with some of the properties removed, and some we need to create new ones with a new type (we'll need an "isClearable").
 
> I didn't read through much of the UI code, the blueprints api and other backend changes look reasonable

Great!

packages/blueprints-integration/src/triggers.ts
Julusian: Hmm ok
Will that work with the translations namespacing?

So it sounds like it is needed to be able to translate some triggers, but that is going to require all triggers to be defined in the blueprints and will make them much harder to create in the ui when only wanted in a single installation
I guess that for now we should leave it as being translatable, and we can figure out a solution when/if we start havign a problem

Julusian:
> I think they should be removed, but I would like to do that in a separate PR

Ok, that is fine with me as long as it happens as soon as this is merged

> we'll need an "isClearable"

Do we? (I havent put any thought into these) It would be nice to talk through what properties we have and what we may want to add, to make sure they all make sense and are in the right place. (A lot of them date back to the beginning)

packages/blueprints-integration/src/triggers.ts
jstarpl: I think theoretically it should, or at least, is in no way in a worse position to work than any other blueprint-provided translations for any other objects. Core-side, ITranslatableMessage has the namespace field which specifies the proper namespace (blueprint) to be used for it. Should still work, and if it doesn't, then it's just a bug in the processing done by Core when Blueprints are setting those values.

I assume that if they are only wanted in a single installation, then the solution is to just manually enter the label in a single/"universal" language. I think the need to be able to handle multiple languages for manually created Triggers is a pretty niche one.

jstarpl:
> Ok, that is fine with me as long as it happens as soon as this is merged

I thought we could lengthen this a bit, so that we have a bit more stable ground. I'm not sure if we are in a "break everything" mood rn?

> > we'll need an "isClearable"
> 
> Do we? (I havent put any thought into these) It would be nice to talk through what properties we have and what we may want to add, to make sure they all make sense and are in the right place. (A lot of them date back to the beginning)

I think we do, at least if we want to keep feature parity. The Action Triggers I have created right now for the NRK blueprints do use a fair bit of "clear source layers X, Y & Z". I suppose they could be replaced by actions, but that also means that a "most basic blueprint implementation" would then require a blueprint developer to manually create all the "clearSourceLayer" adLib actions they expect to need. Adding an `isClearable` follows the existing convention & maintains the same feature set without adding more burden on blueprint developers.

packages/blueprints-integration/src/triggers.ts
Julusian: > I think theoretically it should, or at least, is in no way in a worse position to work than any other blueprint-provided translations for any other objects

I was thinking that if core creates a action trigger, then a blueprint migration changes the text of it, then can we figure out what namespace to translate it under? It feels like we should, its an edge case to be careful of.

> I assume that if they are only wanted in a single installation, then the solution is to just manually enter the label in a single/"universal" language.

Yeah that will probably be good enough. And if not then there is a clear solution of defining it via a blueprint migration.

Julusian:
> I thought we could lengthen this a bit, so that we have a bit more stable ground. I'm not sure if we are in a "break everything" mood rn?

We arent, but I thought that hotkey system will no longer do anything so its a bunch of dead code/ui properties floating around. If they are still hooked up and working then longer is fine with me.

> > > we'll need an "isClearable"
> >
> > Do we? (I havent put any thought into these) It would be nice to talk through what properties we have and what we may want to add, to make sure they all make sense and are in the right place. (A lot of them date back to the beginning)
> 
> I think we do, at least if we want to keep feature parity. The Action Triggers I have created right now for the NRK blueprints do use a fair bit of "clear source layers X, Y & Z". I suppose they could be replaced by actions, but that also means that a "most basic blueprint implementation" would then require a blueprint developer to manually create all the "clearSourceLayer" adLib actions they expect to need. Adding an isClearable follows the existing convention & maintains the same feature set without adding more burden on blueprint developers.

Im not suggesting to remove functionality (this time), just questioning whether there needs to be a property allowing it as well as an action trigger. Currently the backend isnt verifying that layers can be cleared like that, so what purpose will it serve? Maybe the issue here is that the backend should be verifying the clear can be done, in which case the property makes sense. But even then, there will be an action trigger that will rely on another property being set to true for it to work. Its another box to click to make the action trigger work

Julusian (approved):
Im happy for this to be merged now

jstarpl:
> > I thought we could lengthen this a bit, so that we have a bit more stable ground. I'm not sure if we are in a "break everything" mood rn?
> 
> We arent, but I thought that hotkey system will no longer do anything so its a bunch of dead code/ui properties floating around. If they are still hooked up and working then longer is fine with me.

They still do stuff. Specifically, clearKeyboardHotkey is used by the GUI to figure out if a sourceLayer is clearable (if not falsy), that's why I would replace it with `isClearable`, because that's what it will now become, the migration being `isClearable = !!clearKeyboardHotkey`.
  
> > I think we do, at least if we want to keep feature parity. The Action Triggers I have created right now for the NRK blueprints do use a fair bit of "clear source layers X, Y & Z". I suppose they could be replaced by actions, but that also means that a "most basic blueprint implementation" would then require a blueprint developer to manually create all the "clearSourceLayer" adLib actions they expect to need. Adding an isClearable follows the existing convention & maintains the same feature set without adding more burden on blueprint developers.
> 
> Im not suggesting to remove functionality (this time), just questioning whether there needs to be a property allowing it as well as an action trigger. Currently the backend isnt verifying that layers can be cleared like that, so what purpose will it serve? Maybe the issue here is that the backend should be verifying the clear can be done, in which case the property makes sense. But even then, there will be an action trigger that will rely on another property being set to true for it to work. Its another box to click to make the action trigger work

Well, that's an architectural thing. The Action Triggers filters are filters on existing actions and not arguments to an Action. Therefore there needs to exist an action (essentially a "virtual AdLib", because that's what the Users see on the "User" side) that can be filtered out. The only alternative then would be to treat all source layers as clearable, which would make something like "Clear all LOWER_THIRD-type source layers" have unintended consequences. By allowing the Admin/BluePrint Developer to explicitly specify which Source Layers are clearable, a "Clear all LOWER_THIRD-type source layers" will be safe, while still easily allowing the Admin/BPD to add a new Source Layer of type LOWER_THIRD, mark it as clearable, and just have the Action Trigger filter pick it up.

PR #554, opened 2021-08-13
fix(Prompter): show changes in played-out Parts

jstarpl:

This is a bugfix

Changes in script of played-out Parts is not shown, because they are not reset. This includes having the rundown in a "parked" state after a rehearsal.

Non-current or next partInstances are treated as if they were reset, for the purpose of the prompter script.

* **Other information**:

### DIFF
--- a/meteor/lib/api/prompter.ts
+++ b/meteor/lib/api/prompter.ts
@@ -62,7 +62,13 @@ export namespace PrompterAPI {
 			playlist,
 			undefined,
 			undefined,
-			undefined,
+			// unless this is the current or next PartInstance, we actually don't want the PartInstances,
+			// we want it to behave as if all other PartInstances are reset.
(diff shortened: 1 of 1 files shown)


jesperstarkar:
@jstarpl This fixes the prompter but script-pieces in the timeline view will still be "proctedted" and not update?

jstarpl:
> @jstarpl This fixes the prompter but script-pieces in the timeline view will still be "proctedted" and not update?

Yes, this only affects the prompter.

jesperstarkar:
> > @jstarpl This fixes the prompter but script-pieces in the timeline view will still be "proctedted" and not update?
> 
> Yes, this only affects the prompter.

I guess this kinda makes sense.

nytamin (approved):

PR #556, opened 2021-08-17
feat: Budget Duration

ianshade:
A feature called Budget Duration

Parts have a new optional property called `budgetDuration`.
In a Segment that has any Part with `budgetDuration` set, an Editorial Marker will be displayed on the segment timeline, as a dashed vertical line, indicating the budget duration of the entire Segment (sum of `budgetDurations` of its Parts).
In said Segment, the SegmentDuration component will display the segment's budget duration, and count down from it when the segment becomes live. The duration will go negative when the segment is playing longer than its budget.
If a segment with budget duration is on air, Part Counter (on the next segment) is based on budget duration, instead of expected durations of the parts that are left.

* **Other information**:

**Status**


meteor/lib/collections/PartInstances.ts
```diff
@@ -146,7 +146,7 @@ export function findPartInstanceInMapOrWrapToTemporary<T extends Partial<PartIns
 }
 
 export function findPartInstanceOrWrapToTemporary<T extends Partial<PartInstance>>(
-	partInstances: { [partId: string]: T | undefined },
+	partInstances: { [partId: string]: T | undefined } | Map<PartId, T>,
```
nytamin, line 149: Wouldn't it be better to go all the way here? `partInstances: Map<PartId, T>,` ?

nytamin:
There should there be unit-tests added into `meteor/lib/rundown/__tests__/rundownTiming.test.ts`

nytamin:
It would be good if another unit test be added into `meteor/lib/rundown/__tests__/rundownTiming.test.ts`, to support this functionality.

meteor/lib/collections/PartInstances.ts
ianshade: It should stay as it was before because there's another function that handles `Map<PartId, T>`

ianshade:
I added a test case. Note: this feature currently affects only the `partCountdown`. It does not affect `remainingPlaylistDuration`, `totalPlaylistDuration`, etc. but it's subject to change in future PRs

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -615,6 +622,21 @@ export const SegmentTimelineContainer = translateWithTracker<IProps, IState, ITr
 			window.removeEventListener('resize', this.onWindowResize)
 		}
 
+		private getSegmentBudgetDuration(): number | undefined {
```
jstarpl, line 625: I'm somewhat confused. So `RundownTimingCalculator` seems to already be doing almost the same thing, and yet then the Segment component seems to be doing it again? And then it feels like it only does that to pass that information forward to components that have access to the timing context anyway, so they could just get it from the Calculator. Or am I missing something?

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -794,11 +816,14 @@ export const SegmentTimelineContainer = translateWithTracker<IProps, IState, ITr
 						? partOffset + e.detail.currentTime - virtualStartedPlayback + lastTakeOffset
 						: partOffset + lastTakeOffset
 
+				const budgetDuration = this.getSegmentBudgetDuration()
```
jstarpl, line 819: Why do this in the `onAirLineRefresh()`? This is going to happen on every tick, yet `getSegmentBudgetDuration()` only depends on semi-constant `budgetDuration` on the PartInstance. Wouldn't it make more sense to calculate that re actively in the tracker, which already depends on those PartInstances, so it's going to refresh if/when they change?

jstarpl (changes requested):
I'm really confused by some of the implementation & some parts can be definitely implemented in a more straightforward fashion & reducing the amount of inter-dependency between the components.

jstarpl:
@ianshade Can you look into the PR review? I guess you would like this merged into Release 38.

ianshade:
We found an issue causing this to break the fit-to-viewport zoom feature. We're still working on a fix

PR #558, opened 2021-09-02
feat: out transitions

baltedewit:
This PR adds out transitions to Sofie Parts. The part gets an `outTransitionDuration` property and the piece gets an `isOutTransition` property. When the next part is taken, the current part is lengthened by the `outTransitionDuration` and any pieces with `isOutTransition` set get added to the timeline. Pieces starting at `t = 0` in the next part are delayed by the `outTransitionDuration`. An out transition and an in transition may overlap if the in transition has a keepalive duration defined.

Left todo:

- [x] Render `isOutTransition` pieces correctly in the UI
- [x] ~~Separate the first piece delay from `outTransitionDuration`~~
- [x] Disable the transition for edge cases (like hold)

**Status**


Julusian:
From what I can see this is another layer to the transitions, to allow a part to define how to transition out of it. And it will only be used if the next part doesnt have an 'in' transition?

I think we should be careful what we add here, as the transition logic is already fiddly and confusing number calculations, and @nytamin will agree with me that this is something we should rethink from the ground up.

jstarpl (approved):
Looks good to me. @Julusian I think the idea here is to persist the existence of the previous Part and delay the playback of the current Part after a take until the out-transition of the previous Part ends. This allows taking out graphics and stuff to be handled internally within the Part, allowing a clean cut into a different Part after all the graphics have been taken out. To me that feels like a great addition to the Sofie architecture. Even if it might be a bit fiddly, or difficult to code for, one can just not use it. I guess we should rethink transitions at some point, but whatever that will end up to be, having out-transitions out of Parts is good input into that discussion and we should handle that case from day 1 in the new system.

Julusian:
#619 will include this

PR #560, opened 2021-09-07
fix: Assign ranks according to playlist order by rundownExternalId

thomasslee97:

Bug fix

Rundowns appear out of order due to function that assigns the ranks from blueprints treating the keys in the ranks map as `_id` values, when they are `rundownExternalId` values according to the `blueprints-integration` typings. This results in every rundown getting the default rank of 0 as a result, as all the assignments fail.

Ranks are assigned using `rundownExternalId` as the typings direct, resulting in the correct ordering.

* **Other information**:

**Status**


nytamin (approved):
I found a few issues and have pushed fixes for them.

PR #563, opened 2021-09-09
Feat/evs support

gundelsby:
Feature: Adds support for EVS content type

**Other information**:

**Status**


packages/blueprints-integration/src/content.ts
```diff
@@ -18,6 +18,7 @@ export enum SourceLayerType {
 	TRANSITION = 13,
 	// LIGHTS = 14,
 	LOCAL = 15,
+	EVS = 16,
```
jstarpl, line 21: This is my only real objection: I don't think we should have a SourceLayerType called `EVS`, the same way we don't have a SourceLayerType called `LIVEU` for remote sources. The SourceLayerType should reflect the editorial purpose of this source and not it's techical implementation. In my personal opinion, a replay input (in this case, an EVS machine) falls into the "LOCAL" category, being a local presentation source that is not a camera or a full screen graphic. Then it would fall into the same category as a laptop with a PowerPoint presentation running or a Weather graphics system. I think it would be perfectly fine to just remove the special styling for `.evs`, use the new color for `$segment-layer-background-local` (since we haven't used `LOCAL` anyway) and rename `EvsLayerItemRenderer` to `LocalLayerItemRenderer`

packages/blueprints-integration/src/content.ts
```diff
@@ -78,6 +80,13 @@ export interface RemoteContent extends BaseContent {
 	switcherInput: number | string
 }
 
+export interface EvsContent extends BaseContent {
+	studioLabel: string
+	switcherInput: number | string
+	evsChannel: string
+	uiColorCode?: string
```
jstarpl, line 87: I don't think I like `uiColorCode`, because what is that code? I think it would be safer to just have this a color RGB in hash-HEX (#xxxxxx) format, same way as we do in `IBlueprintPart.invalidReason.color`: https://github.com/nrkno/tv-automation-server-core/blob/152f988e5692669516ead1ef8fb22ee395a69398/packages/blueprints-integration/src/rundown.ts#L145-L161

This way, we use the same color system for all places. I guess what you wanted was to be able to just say `red` here, but I think it's entirely possible, that some other system (not invented yet, but still) might need to either also display this, and it would not have the X11 Color Names table or would have to do some processing on it (say dim it, or add some transparency), in which case it would make it unnecessarily complicated.

jstarpl (changes requested):
I don't think we need a new `EVS` type, why not use `LOCAL` instead?

packages/blueprints-integration/src/content.ts
gundelsby: This has the side-effect that if we reuse LOCAL for anything else later, those parts will also have the purple background color that is meant to be used for EVS parts now.
The need to denote the specific EVS channel (using the color mark) is also not something that would be relevant for a laptop with a Powerpoint presentation. Since the UI color is optional, that is probably less relevant.

I don't have a particularly strong opinion on it, but it's worth considering.

packages/blueprints-integration/src/content.ts
gundelsby: Makes sense, I guess representing it on some panel with colored LED buttons is not a too far-fetched example of something we'd like to be able to do. We don't validate the input anywhere though, so if we want some kind of security that it will always be a hex color code that should probably be added?

The reason for it to be named uiColor instead of simply color is that since the EVS channel names are colors (in wording, ex "GUL" (Norwegian for yellow)) I'd like for the property name to be unambiguous. So I'd like to keep the "ui" in the name for that reason.

jstarpl (approved):

PR #564, opened 2021-09-10
Fix: media status polling

nytamin:
Bug fix + performance improvement

* The first commits adds a cache in front of the method `getMediaObjectIssues`, so that when having multiple clients open that request the same status, the result is reused.
* The second commit fixes a bug, where an interval isn't cleaned up when leaving the rundown view. (can be a problem if going into and out of rundowns many times, as the setIntervals accumulates.)

* **Other information**:

**Status**

### DIFF
--- a/meteor/lib/lib.ts
+++ b/meteor/lib/lib.ts
@@ -374,7 +374,24 @@ export function cacheResult<T>(name: string, fcn: () => T, limitTime: number = 1
 	}
 	const cache = cacheResultCache[name]
 	if (!cache || cache.ttl < Date.now()) {
-		const value = fcn()
+		const value: T = fcn()
+		cacheResultCache[name] = {
+			ttl: Date.now() + limitTime,
+			value: value,
+		}
+		return value
+	} else {
+		return cache.value
+	}
+}
+/** Cache the result of function for a limited time */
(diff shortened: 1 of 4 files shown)


jstarpl (approved):
Looks good, good addition on the cached method result. Maybe add a comment on the caching function that the name is the cache key.

meteor/lib/lib.ts
```diff
@@ -374,7 +374,24 @@ export function cacheResult<T>(name: string, fcn: () => T, limitTime: number = 1
...
+		}
+		return value
+	} else {
+		return cache.value
+	}
+}
+/** Cache the result of function for a limited time */
+export async function cacheResultAsync<T>(name: string, fcn: () => Promise<T>, limitTime: number = 1000): Promise<T> {
+	if (Math.random() < 0.01) {
+		Meteor.setTimeout(cleanOldCacheResult, 10000)
```
jstarpl, line 390: Just a side question: is this better/more performant than having a single interval just run the cache cleanup? Is the point here to have a bit of fuzzyness for the cleanup?

nytamin: Yeah, a setInterval would probably work just as fine. 
This might make `cleanOldCacheResult` to not run unnessesarily when idle, but really it doesn't matter, brcause when idle all of the cache is likely to have been emptied, so `cleanOldCacheResult` would take no time at all to run.

PR #568, opened 2021-09-12
Fix: package data in snapshots

nytamin:
This PR adds more data into snapshots, related to Packages.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/snapshot.ts
+++ b/meteor/server/api/snapshot.ts
@@ -32,6 +32,7 @@ import {
 	makePromise,
 	ProtectedString,
 	protectStringArray,
+	literal,
 } from '../../lib/lib'
 import { ShowStyleBases, ShowStyleBase, ShowStyleBaseId } from '../../lib/collections/ShowStyleBases'
 import { PeripheralDevices, PeripheralDevice, PeripheralDeviceId } from '../../lib/collections/PeripheralDevices'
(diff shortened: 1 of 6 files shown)


meteor/server/api/snapshot.ts
```diff
@@ -711,6 +756,86 @@ export async function restoreFromRundownPlaylistSnapshot(
...
+			packageContainerPackageStatus.studioId,
+			packageContainerPackageStatus.containerId,
+			packageContainerPackageStatus.packageId
+		)
+	})
+	snapshot.packageContainerPackageStatuses = snapshot.packageContainerPackageStatuses.filter(
+		(p) => p.packageId !== protectString('N/A')
+	)
+
+	_.each(snapshot.packageInfos, (packageInfo) => {
```
Julusian, line 834: This would be nicer to do as a single loop that builds a new array, instead of using a temporary magic value. Same for all the loops above.
eg
```
let restorePackageInfos = []
_.each(snapshot.packageInfos, (packageInfo) => {
  const newPackageId = expectedPackageIdMap.get(packageInfo.packageId)
  if (newPackageId) {
    packageInfo.packageId = newPackageId
    restorePackageInfos.push(packageInfo)
  }
})
```

meteor/server/api/snapshot.ts
```diff
@@ -711,6 +756,86 @@ export async function restoreFromRundownPlaylistSnapshot(
...
+				expectedPackage.rundownId = rundownIdMap[unprotectString(expectedPackage.rundownId)]
+				ownerId = expectedPackage.pieceId
+				break
+			}
+			case ExpectedPackageDBType.RUNDOWN_BASELINE_OBJECTS: {
+				expectedPackage.rundownId = rundownIdMap[unprotectString(expectedPackage.rundownId)]
+				ownerId = expectedPackage.rundownId
+				break
+			}
+			case ExpectedPackageDBType.BUCKET_ADLIB:
```
Julusian, line 795: should we try to restore these? they don't belong to the rundown so could replace any newer version currently in the db. Same for STUDIO_BASELINE_OBJECTS

Julusian (commented):
I am concerned about the implications of restoring a snapshot into a live system here.
It looks like it could replace current versions of expected packages with the versions from the snapshot, which could be bad.

nytamin:
@Julusian regarding data that doesn't explicitly belong to a rundown:
I've been struggling with how to implement the restore, and it'll be hard to ensure that we don't break anything unintended when restoring non-rundown data.

Maybe it would be better to not restore this data at all?
I still think it adds a value to have it in a snapshot, but maybe that's enough, for now?

Julusian:
yeah I think that not restoring that data is the best option for now.
Being is the snapshot but unused in restoring is fine with me

PR #574, opened 2021-09-16
feat: pieces can specify how to be direct-played

Julusian:

Feature

When a piece is double clicked, it gets 'direct played' as an adlib piece. This is only allowed for LOWER_THIRD pieces, as they are known to be safe to do this with.

This introduces a new rule that lets the blueprints say which pieces are safe to use as an adlib like this.
Additionally, it states whether it should be used as an adlib or if an adlib-action should be executed instead.

* **Other information**:

**Status**

### DIFF
--- a/packages/blueprints-integration/src/rundown.ts
+++ b/packages/blueprints-integration/src/rundown.ts
@@ -284,17 +284,24 @@ export interface PieceTransition {
 	duration: number
 }
 
-export interface IBlueprintDirectPlayAdLibPiece {
-	type: 'adlib'
+export enum IBlueprintDirectPlayType {
+	AdLibPiece = 'adlib',
+	AdLibAction = 'action',
 }
-export interface IBlueprintDirectPlayAdLibAction {
-	type: 'action'
+export interface IBlueprintDirectPlayBase {
+	type: IBlueprintDirectPlayType
+}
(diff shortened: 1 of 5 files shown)


packages/blueprints-integration/src/rundown.ts
```diff
@@ -283,6 +284,18 @@ export interface PieceTransition {
 	duration: number
 }
 
+export interface IBlueprintDirectPlayAdLibPiece {
```
nytamin, line 287: A small note, to better describe that `IBlueprintDirectPlayAdLibPiece` and `IBlueprintDirectPlayAdLibAction` are related, I usually do something like this:

```
interface myInterfaceBase {
  type: myInterfaceType
}
enum myInterfaceType {
  A: 'a',
  B: 'b'
}
interface myInterfaceA extends myInterfaceBase {
  type: myInterfaceType.A
  ...otherProperties
}
interface myInterfaceB extends myInterfaceBase {
  type: myInterfaceType.B
  ...otherProperties
}
type myInterface = myInterfaceA | myInterfaceB

```

I think that both extending a base interface and using an enum for the `type` helps with understanding, using and extending the interfaces.

nytamin (approved):
One note, otherwise looks good!

PR #577, opened 2021-09-20
Feat: Package containter statuses

nytamin:
This PR adds API methods, a new collection and GUI for tracking the status of PackageContainers.

image


meteor/client/ui/Status/StatusCodePill.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { StatusCode } from '@sofie-automation/blueprints-integration'
+import { statusCodeToString } from './SystemStatus'
+
+interface StatusCodePillProps {
+	connected: boolean
+	statusCode: StatusCode
+	messages?: string[]
+}
+export const StatusCodePill = reacti18next.withTranslation()(
+	class StatusCodePill extends React.Component<StatusCodePillProps & reacti18next.WithTranslation, {}> {
```
jstarpl, line 15: Generally, I would suggest using a React Function component here, since this is essentially a bare render function.

meteor/client/ui/Status/package-status/PackageContainerStatus.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { StatusCodePill } from '../StatusCodePill'
+import { doUserAction, UserAction } from '../../../lib/userAction'
+import { MeteorCall } from '../../../../lib/api/methods'
+
+interface IPackageContainerStatusProps {
+	packageContainerStatus: PackageContainerStatusDB
+}
+
+export const PackageContainerStatus = withTranslation()(
+	class PackageContainerStatus extends React.Component<Translated<IPackageContainerStatusProps>, {}> {
```
jstarpl, line 15: This is also a very good candidate for a React Function component.

meteor/client/ui/Status/package-status/package-status.scss
```diff
@@ -136,6 +136,59 @@
...
+				padding: 0.5em 0.5em 0.5em 0em;
+				background-color: #c6c6c6;
+				vertical-align: middle;
+				border-bottom: 1px solid #919191;
+				cursor: pointer;
+
+				font-size: 90%;
+				position: relative;
+
+				&:first-child {
```
jstarpl, line 181: It would probably be better to use a modifier class rather than a child-selector here, a child selector seems very brittle.

jstarpl (approved):
Looks good, though I would suggest using Function components for the GUI, since their operation is reasonably straightforward.

nytamin:
I've refactored to use function-components and react hooks instead, good points @jstarpl !

meteor/client/ui/Status/package-status/index.tsx
```diff
@@ -13,131 +12,174 @@ import { doUserAction, UserAction } from '../../../lib/userAction'
...
+			useSubscription(PubSub.expectedPackages, {
+				studioId: { $in: studioIds },
+			}),
+			useSubscription(PubSub.packageContainerStatuses, {
+				studioId: { $in: studioIds },
+			}),
+			studioIds && studioIds.length > 0,
+		].reduce((memo, value) => memo && value, true) || false
+
+	const expectedPackageWorkStatuses = useTracker(() => ExpectedPackageWorkStatuses.find({}).fetch(), []) || []
```
jstarpl, line 45: Actually, useTracker has a third, optional, `initial` parameter you should use here. This is then the value returned if tracker hasn't returned anything yet.

jstarpl (changes requested):

meteor/server/api/userActions.ts
```diff
@@ -1143,6 +1150,9 @@ class ServerUserActionAPI extends MethodContextAPI implements NewUserActionAPI {
 	async packageManagerAbortExpectation(_userEvent: string, deviceId: PeripheralDeviceId, workId: string) {
 		return makePromise(() => packageManagerAbortExpectation(this, deviceId, workId))
 	}
+	async packageManagerRestartPackageContainer(_userEvent: string, deviceId: PeripheralDeviceId, containerId: string) {
+		return makePromise(() => packageManagerRestartPackageContainer(this, deviceId, containerId))
```
Julusian, line 1154: For new code it would be nice for it to be primarily promise based and avoid conversions between promises and fibers when possible.
So instead of `makePromise()` packageManagerRestartPackageContainer should be async and return a Promise.

What you have done is consistent with the other packageManager methods here, so I am happy for that to be done later

meteor/server/api/packageManager.ts
```diff
@@ -45,4 +45,15 @@ export namespace PackageManagerAPI {
 			})
 		)
 	}
+	export function restartPackageContainer(
+		context: MethodContext,
+		deviceId: PeripheralDeviceId,
+		containerId: string
+	): any {
```
Julusian, line 52: this should be `void` rather than `any`

Julusian (approved):

PR #580, opened 2021-09-22
fix: getUnfinishedPieceInstancesReactive not considering interactions between infinite pieces

Julusian:

Bug fix

When a onRundownChange piece is ended by an onRundownEnd piece, the onRundownChange one is still shown as being on-air.

This is because the piece exists, but the infinite logic prunes the piece out as it is determined to be 0 length. This causes it to not get correct playout timings, and so the ui thinks it is playing.

The pieceInstances are run through the same infinite processing logic that is used elsewhere before filtering down to the active ones.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Shelf/DashboardPanel.tsx
+++ b/meteor/client/ui/Shelf/DashboardPanel.tsx
@@ -37,7 +37,7 @@ import { setShelfContextMenuContext, ContextType } from './ShelfContextMenu'
 import { RundownUtils } from '../../lib/rundown'
 import { RundownPlaylist } from '../../../lib/collections/RundownPlaylists'
 import { processAndPrunePieceInstanceTimings } from '../../../lib/rundown/infinites'
-import { Rundowns } from '../../../lib/collections/Rundowns'
+import { DBShowStyleBase } from '../../../lib/collections/ShowStyleBases'
 
 interface IState {
 	outputLayers: {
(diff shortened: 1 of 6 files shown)


meteor/client/ui/Shelf/DashboardPanel.tsx
```diff
@@ -719,6 +697,7 @@ export function isAdLibOnAir(
 	unfinishedTags: IDashboardPanelTrackedProps['unfinishedTags'],
 	adLib: AdLibPieceUi
 ) {
+	console.log('isAdLibOnAir', adLib.name, adLib._id, adLib.currentPieceTags, unfinishedTags)
```
jstarpl: This should be removed, otherwise, unless the performance is hurt, it looks good.

meteor/client/ui/Shelf/DashboardPanel.tsx
```diff
@@ -568,79 +570,53 @@ export function getUnfinishedPieceInstancesReactive(playlist: RundownPlaylist) {
...
-					],
-				},
-			],
-		}).fetch()
+		const partInstance = PartInstances.findOne(playlist.currentPartInstanceId)
+
+		if (partInstance) {
+			const rundown = Rundowns.findOne(partInstance.rundownId)
+			if (rundown) {
+				const showStyle = rundown.getShowStyleBase()
```
jstarpl, line 578: Is the performance good with this? Seems like a lot of data to be pulled in and a lot of reactivity, even though, perhaps it doesn't have to be this reactive? I think it would be good to narrow the reactivity on the `Rundowns` at least? `rundown` seems to be pulled in only to get the ShowStyleBase, which I think could be piped in some other way maybe?

meteor/client/ui/Shelf/DashboardPanel.tsx
```diff
@@ -568,79 +570,53 @@ export function getUnfinishedPieceInstancesReactive(playlist: RundownPlaylist) {
...
+
+				prospectivePieces = PieceInstances.find({
```
jstarpl, line 580: This seems like a very wide net, is the performance good with this change?

jstarpl (changes requested):
I'm somewhat worried by the performance. It does seem to cast pretty wide, reactive nets.

meteor/client/ui/Shelf/DashboardPanel.tsx
Julusian: I honestly don't know how to check, any tips? I get pretty bad stuttering and high cpu spikes locally before this change too.

Apart from omitting fields, I don't think we can safely filter out any pieces from this without risking differences in the operation

meteor/client/ui/Shelf/DashboardPanel.tsx
Julusian: pulling in the rundown did feel a bit excessive when I started this, I forgot to look at other routes for getting the showstylebase. Ill have a look into if we can pipe in a showstyle id, or if not ill trim the rundown down to just the necessary ids

meteor/client/ui/Shelf/DashboardPanel.tsx
jstarpl: Yeah, well, that's what I've always warned about when the Infinites changes were implemented: some major heavy lifting needs to be done by the UI anytime anything changes. I have not figured out a way to make it less heavy, I think, unfortunately, that's just a part of this Infinites architecture.

Okay, I started to think about doing like a ReactiveVar or something with a React Context to at least share this data across components, but I think there's actually a simpler (at least implementation-wise) way: wrap the entire `getUnfinishedPieceInstancesReactive` in a `memoizedIsolatedAutorun`. The only thing it depends on is the playlistId, so that can be easily piped in as a parameter and then at least we'll only get one of these calculations for the entire view.

jstarpl (changes requested):

meteor/client/ui/Shelf/DashboardPanel.tsx
Julusian: I think I shall need some help with configuring this properly. I tried doing:
```
return memoizedIsolatedAutorun(
 	(playlistActivationId, currentPartInstanceId, showStyleBase) => {
.....
},
	'getUnfinishedPieceInstancesReactive',
	playlist.activationId,
	playlist.currentPartInstanceId,
	showStyleBase
)

but that is not invalidating correctly when I run an adlib. Any suggestions?

meteor/client/ui/Shelf/DashboardPanel.tsx
Julusian: yep, two method calls up we always have a showStyleBase available. That is now passed in as a parameter

jstarpl (approved):
Looks like a good change. Memoized `getUnfinishedPieceInstancesReactive` should increase performance when using a lot of panels.

PR #581, opened 2021-09-23
fix: rewinding and scrolling next part

KvelaGorrrrnio:
Bug fix

1. Sometimes segments doesn't rewind, with autoRewindLeavingSegment enabled.
2. When setting a part as next, it is not scrolled into the viewport, when it should be.

1. Segments rewind with autoRewindLeavingSegment enabled.
The bug was caused from a condition dependent on a state attribute that was set asynchronous with setState. Thus the condition sometimes received the old and sometimes the new state.

2. When setting a part as next, if the segment is not live and the part is outside of the SegmentTimelineContainer viewport, the part is scrolled to the start of the SegmentTimelineContainer viewport.
The reason was that the condition triggering scrollLeft change, did not cover all use cases and some conditions were incorrect; onGoToPartInner wrote an old value of scrollLeft into state.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
+++ b/meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
@@ -523,14 +523,14 @@ export const SegmentTimelineContainer = translateWithTracker<IProps, IState, ITr
 				this.context.durations.partDisplayStartsAt[unprotectString(currentNextPart.partId)]
 			const partOffset =
 				nextPartDisplayStartsAt -
-					this.context.durations.partDisplayStartsAt[unprotectString(this.props.parts[0].instance.part._id)] || 0
+					(this.props.parts.length > 0 ? this.context.durations.partDisplayStartsAt[unprotectString(this.props.parts[0].instance.part._id)] ?? 0 : 0)
 			const nextPartIdOrOffsetHasChanged =
 				currentNextPart &&
 				this.props.playlist.nextPartInstanceId &&
 				(prevProps.playlist.nextPartInstanceId !== this.props.playlist.nextPartInstanceId ||
 					this.nextPartOffset !== partOffset)
 			const isBecomingNextSegment = this.state.isNextSegment === false && isNextSegment
-			// Setting the correct scroll position on parts when setting as next
+			// the segment isn't live, will be next, and either the nextPartId has changed or it is just becoming next
 			if (
 				!isLiveSegment &&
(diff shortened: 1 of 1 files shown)


jstarpl (approved):
Looks like a good fix!

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -508,38 +508,46 @@ export const SegmentTimelineContainer = translateWithTracker<IProps, IState, ITr
...
+				this.context.durations.partDisplayStartsAt[unprotectString(currentNextPart.partId)]
+			const partOffset =
+				nextPartDisplayStartsAt -
+					this.context.durations.partDisplayStartsAt[unprotectString(this.props.parts[0].instance.part._id)] || 0
+			const nextPartIdOrOffsetHasChanged =
+				currentNextPart &&
+				this.props.playlist.nextPartInstanceId &&
+				(prevProps.playlist.nextPartInstanceId !== this.props.playlist.nextPartInstanceId ||
+					this.nextPartOffset !== partOffset)
+			const isBecomingNextSegment = this.state.isNextSegment === false
```
jstarpl, line 533: Shouldn't this be `const isBecomingNextSegment = this.state.isNextSegment === false && isNextSegment`?
Also, please keep the comment in for the ifs, it makes it much easier to return to this code later and understand what each of it's parts is supposed to do. I know that the variable names are somewhat self-explanatory, but still.

jstarpl (changes requested):
Thanks! This is certainly a good catch.

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
KvelaGorrrrnio: I agree. The reason for not having it was that `isNextSegment` is included in the if-statement. But I will change it to include it in `const isBecomingNextSegment` for better readability.
Just to be on the same page, you want the comment, that describes the use case for the condition, to be placed as the first line inside the condition body?

jstarpl: The comment being _"the segment isn't live, is next, and the nextPartId has changed"_, which I guess should now be _"the segment isn't live, will be next, and either the nextPartId has changed or it is just becoming next"_, which I guess maybe points to the fact that some of these checks in that if statement are now redundant? Hmm... 🤔

PR #582, opened 2021-09-23
feat: add level property to Part invalidReason (#582) [publish]

Julusian:

Feature

Part invalidReason notes are forced to be of ERROR level. This is not always desired, so they are being done manually from the blueprints as other notes to be a different level.

Additionally, these notes are currently ignored, which would be better handled in the blueprints if still desired

Blueprints can specify the level to show the notes at.
If an invalidReason should not be shown, they can omit the invalidReason field completely (not new, but should be done instead of disabling the notes entirely)

* **Other information**:

**Status**


nytamin (approved):
Looks good!
I'm personally not a fan of the naming of the `NoteType` enum, I think that something more verbose would be better, like `noteSeverity` or something like that.

PR #583, opened 2021-09-24
chore: add ignore_piece_content_status url parameter to disable piece content status checks

Julusian:

developer feature

in a setup with no working media-manager/package-manager clip missing warnings are produced for every clip in the rundown, which hides any potential real warnings that are produced

adds a url 'config' parameter, to bypass the piece content checks

* **Other information**:

@jstarpl I am unsure if doing the localStorage checks will have a performance penalty and if the value should be cached?

Targetting r37 for now, that can be changed to r38 if wanted

**Status**


jstarpl (approved):
Local storage is fast, but it's still way 100x slower than variable access, so I think we should start caching them. Also, I think we need to start thinking about some system for these debug flags, because it's starting to be unruly.

Still, this can be done in some other PR.

PR #589, opened 2021-09-30
Fix/device logging

nytamin:
This PR adds a "debug" property to the playout devices.
When checked, their debug messages are logged, otherwise they are not sent from the devices.

For backwards compatibility, their messages are logged as usual when the old debug flag on the playout-gateay is set.

* **Other information**:

**Status**

### DIFF
--- a/packages/playout-gateway/src/tsrHandler.ts
+++ b/packages/playout-gateway/src/tsrHandler.ts
@@ -468,7 +468,7 @@ export class TSRHandler {
 		const peripheralDevices = this._coreHandler.core.getCollection('peripheralDevices')
 		const peripheralDevice = peripheralDevices.findOne(this._coreHandler.core.deviceId)
 
-		const ps: Promise<any>[] = []
+		let ps: Promise<any>[] = []
 		const promiseOperations: { [id: string]: true } = {}
 		const keepTrack = <T>(p: Promise<T>, name: string) => {
 			promiseOperations[name] = true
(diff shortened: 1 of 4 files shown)


packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -479,31 +477,30 @@ export class TSRHandler {
 				return result
 			})
 		}
+		const devices: {
```
Julusian, line 480: Perhaps this could be made into a `Map<string, DeviceOptionsAny>`?
It will require checking/updating all usages, so not essential to do right now, but should be done at some point to better handle `undefined` values

packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -479,31 +477,30 @@ export class TSRHandler {
...
-			} = {}
-
 			_.each(settings.devices, (device, deviceId) => {
 				if (!device.disable) {
 					devices[deviceId] = device
 				}
 			})
 
-			_.each(devices, (deviceOptions: DeviceOptionsAny, deviceId: string) => {
+			_.each(devices, (orgDeviceOptions: DeviceOptionsAny, deviceId: string) => {
```
Julusian, line 493: while you are here, it would be nice to remove some of the `_.each` calls

packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -564,9 +566,25 @@ export class TSRHandler {
 				}, INIT_TIMEOUT)
 			), // Timeout if not all are resolved within INIT_TIMEOUT
 		])
+
+		// Set logDebug on the devices:
+		_.each(this.tsr.getDevices(), async (device: DeviceContainer<DeviceOptionsAny>) => {
```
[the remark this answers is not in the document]
nytamin: Good find! This was true for the loop at :544 too. I'll fix.

Julusian (approved):

PR #591, opened 2021-10-01
Feat: rundown metadata update

thomasslee97:

Feature

Currently, the only way to update the timing of rundown playlists via the `dataRundown*` methods is to use either the `dataRundownUpdate` or `dataRundownCreate` methods. Both of these methods require passing all of the segments in the rundown to core and will re-run `getSegment` for each of these segments. In long rundowns this can have a measurable impact on performance.

This PR adds the method `dataRundownMetaDataUpdate` which allows just the `IngestRundown` without any segments to be passed to core, to be evaluated in the `getRundown` function of blueprints.

When this method is called, the segment data for the rundown is fetched from the ingest data cache so that `getRundown` still has access to all of the segment data.

If the value of the `rundown.metaData` property returned from `getRundown` differs from what is currently the value on the rundown, then the segments are considered to have been invalidated and `getSegment` will be re-run for each segment. However, if the data in `rundown.metaData` is the same as previously seen, the existing segment data is considered to still be valid.

This way, the timing of rundowns along with the name and any other rundown-level properties can be changed without reevaluating all the segments, but we can still ensure that any changes to the rundown that may affect segments will still be picked up.

To this end, `getSegment` no longer has access to the full rundown information, but only to `rundown.metaData` and `rundown.externalId`. The typings of the data passed to `getSegment` have been updated to reflect this.

* **Other information**:

This solution was sketched out with the help of @Julusian 

**Status**

### DIFF
--- a/meteor/server/api/ingest/rundownInput.ts
+++ b/meteor/server/api/ingest/rundownInput.ts
@@ -5,14 +5,15 @@ import { DBRundown, Rundowns } from '../../../lib/collections/Rundowns'
 import { getCurrentTime, literal } from '../../../lib/lib'
 import { IngestRundown, IngestSegment, IngestPart, IngestPlaylist } from '@sofie-automation/blueprints-integration'
 import { logger } from '../../../lib/logging'
-import { Studio, StudioId } from '../../../lib/collections/Studios'
-import { SegmentId, Segments } from '../../../lib/collections/Segments'
+import { Studio, StudioId, Studios } from '../../../lib/collections/Studios'
+import { DBSegment, SegmentId, Segments } from '../../../lib/collections/Segments'
 import {
 	RundownIngestDataCache,
 	LocalIngestRundown,
 	makeNewIngestSegment,
 	makeNewIngestPart,
 	makeNewIngestRundown,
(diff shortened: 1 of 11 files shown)


meteor/server/api/ingest/rundownInput.ts
```diff
@@ -426,6 +504,86 @@ export async function regenerateRundown(
...
+		showStyleBlueprint,
+		allRundownWatchedPackages
+	)
+
+	// Save rundown and baseline
+	const dbRundown = await saveChangesForRundown(cache, dbRundownData, rundownRes, showStyle)
+
+	let segmentChanges: UpdateSegmentsResult | undefined
+	let removedSegments: DBSegment[] | undefined
+	if (!_.isEqual(existingRundown.metaData, dbRundown.metaData)) {
```
Julusian, line 559: I think this should be done as `if (!_.isEqual(rundownToSegmentRundown(existingRundown), rundownToSegmentRundown(dbRundown))) {`
so that when more properties are added to `
```suggestion
	if (!_.isEqual(existingRundown.metaData, dbRundown.metaData)) {
```

PR #594, opened 2021-10-06
fix: tsr initialization loop

mint-dewit:
Separates the creation and initialization of TSR devices such that playout-gw does not recreate devices that are waiting on initialization. In addition, this PR makes it so that only 1 setTimeout for triggerUpdateDevices is active at the same time, previously the amount of setTimeouts would keep growing.

Devices stuck in initialization will be duplicated, devices error'ing will be recreated too soon

Devices stuck in initialization will be left alone, devices error'ing will be recreated after 10 seconds.

* **Other information**:
Adds some debug logging through the debug library, only during development

**Status**

### DIFF
--- a/meteor/Dockerfile
+++ b/meteor/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:experimental
 # BUILD IMAGE
-FROM node:12.22.1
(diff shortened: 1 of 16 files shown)


nytamin (approved):
LGTM!

PR #602, opened 2021-10-20
feat: timeline and mappings test tools to view routed timeline and mappings

Julusian:

Feature

The timeline testtool page is using the raw timeline, and not the routed timeline.
Additionally, there is no way to view the result of the routing of mappings.

The timeline testtool page uses the same routing of the timeline as used by playout-gateway.
An additional mappings testtool has been added to view the mappings being sent to playout-gateway, also respecting the routesets

* **Other information**:

**Status**


nytamin (approved):
Looks good!

PR #609, opened 2021-10-27
feat: Optimize timeline publication

nytamin:
Performance improvement

When a new timeline is generated:

1. The timeline is generated and is put into the in-memory Cache
2. The timeline is saved into MongoDB
3. MongoDB emits the change back to Core
4. The publication picks up the change and starts recalculating (routing) the timeline.
5. The publication sends the new timeline to Playout-gateway

In my local tests, the time from 1 to 4 is about **20-30 ms**.

When a new timeline is generated:

1. The timeline is generated and is put into the in-memory Cache
2. The timeline is also fast-tracked and sent directly to the publication
3. The publication picks up the change and starts recalculating (routing) the timeline.
4. The publication sends the new timeline to Playout-gateway
5. The timeline is saved into MongoDB (from step 1)
6. MongoDB emits the change back to Core
7. The publication picks up the change, but does nothing, since the `timelineHash` is the same.

In my local tests, the time from 1 to 3 is basically zero.

**Other information**:
This PR is targetting release38, but is based on release37

**Status**

### DIFF
--- a/meteor/server/publications/fastTrack.ts
+++ b/meteor/server/publications/fastTrack.ts
@@ -1,0 +1,52 @@
+import { Meteor } from 'meteor/meteor'
+import { logger } from '../logging'
+
+export enum FastTrackObservers {
+	TIMELINE = 'timeline',
+}
+
+const fastTrackObserver: {
+	[key: string]: {
+		useCount: number
+		onData: (data: any) => void
+	}
+} = {}
+
+function getKey(key: string, args: any[]): string {
+	return key + args.join(',')
+}
+
+export function setupFastTrackObserver<T>(
+	key: string,
+	keyArgs: any[],
+	onData: (data: T) => void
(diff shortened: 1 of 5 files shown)


meteor/server/publications/fastTrack.ts
```diff
@@ -0,0 +1,52 @@
...
+		useCount: number
+		onData: (data: any) => void
+	}
+} = {}
+
+function getKey(key: string, args: any[]): string {
+	return key + args.join(',')
+}
+
+export function setupFastTrackObserver<T>(
```
baltedewit, line 19: Missing documentation: this function can only set up 1 observer function per key + keyargs, I presume this is acceptable for the current implementation but can lead to confusion in the future.

baltedewit (approved):
Looks good in general, just small notes regarding readability.

PR #612, opened 2021-10-27
feat: Action Triggers tag filter suggestions

jstarpl:

Feature & Bug fix

When editing the tags field in the Action Triggers, the field is a free text field, with no information on the tags available

When editing the tags filter, a list of available tag suggestions will be displayed, based on the selected "sample" rundown. Since this list may not be exhaustive, the field is still non-validated (you can still input multiple comma separated tags), but it should give a bit more help during manual setup.

This PR also fixes an issue with the performance of that view, because all of the components were reactive whenever any of the Action Triggers objects changed. Only the modified Action Trigger component should rerender now.

* **Other information**:

### DIFF
--- a/meteor/client/ui/Settings/components/triggeredActions/actionEditors/filterPreviews/AdLibFilter.tsx
+++ b/meteor/client/ui/Settings/components/triggeredActions/actionEditors/filterPreviews/AdLibFilter.tsx
@@ -6,6 +6,12 @@ import { TFunction } from 'i18next'
 import { assertNever } from '../../../../../../../lib/lib'
 import { FilterEditor } from './FilterEditor'
 import { ShowStyleBase } from '../../../../../../../lib/collections/ShowStyleBases'
+import { EditAttributeType } from '../../../../../../lib/EditAttribute'
+import { useTracker } from '../../../../../../lib/ReactMeteorData/ReactMeteorData'
+import { RundownBaselineAdLibActions } from '../../../../../../../lib/collections/RundownBaselineAdLibActions'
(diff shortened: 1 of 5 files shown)


meteor/client/ui/Settings/components/triggeredActions/actionEditors/filterPreviews/AdLibFilter.tsx
```diff
@@ -287,15 +298,39 @@ export const AdLibFilter: React.FC<IProps> = function AdLibFilter({
 		'type',
 	]
 
+	const availableOptions = useTracker<Record<string, any> | string[], Record<string, any> | string[]>(
+		() => {
+			// tags are a special case because we need to search the database for available options
+			// we should have the data subscribed already
+			if (link.field === 'tag') {
+				return _.chain(
+					RundownBaselineAdLibActions.find()
```
baltedewit, line 307: It might be more readable to do
```
[
  ...RundownBaselineAdllibActions.find().map(...),
  ...AdLibActions.find().map(...),
  ...RundownBaselineAdLibPieces.find().map(...),
  ...AdLibPieces.find().map(...),
]
```

baltedewit (approved):

meteor/client/ui/Settings/components/triggeredActions/actionEditors/filterPreviews/AdLibFilter.tsx
jstarpl: Thanks, that's a good suggestion.

PR #617, opened 2021-11-02
feat: Store timeline objects as a data blob

jstarpl:

New feature

Timeline objects are stored as structured data in the database. A working theory is that storing deeply nested objects in MongoDB has penalties and we don't actually need to do any database operations on this data and we in fact treat it as a data blob.

The data is serialized to JSON before being saved into the database as a String. This should speed up how quickly the data can be stored.

* **Other information**:

Work here is based on https://github.com/nrkno/tv-automation-server-core/pull/609 and https://github.com/nrkno/tv-automation-server-core/pull/598, so those should be merged first.


nytamin (approved):
I changed the timelineBlob to be a ProtectedString to provide additional type safety, and now I am happy with this.

~That being said, I did some local tests (which is not representative of prod, but still) and I did not see any significant improvements in response times.~
Jstarpl has told me that we do have seen pretty good performance improvements in prod-like conditions.

Julusian (approved):
I am happy with this

I have pushed another commit to this, to avoid a  place where we were parsing the timeline to try to find a special timeline object that is better served as properties on the collection document

baltedewit (approved):

PR #619, opened 2021-11-08
feat: Per piece prerollDuration

Julusian:

Feature
Builds on #558

The transition and preroll properties for a part are confusing in how they behave and interact. They are based on an early idea of how transitions should work and how parts need to interact.
The preroll offsets were applied inconsistently to the pieces depending on what they are and when they start.

The behaviour of the properties is best summarised by the demo: https://codepen.io/julusian/pen/YzxYpXm?editors=1010 

The primary motivation of this change was to move the prerollDuration property of the part to the pieces that need it. This is because when non-program content needs to preroll, it was incredibly difficult to get the blueprints to calculate the right numbers.

As part of this, the numbers used by transitions needed reviewing, and so have been renamed to better clarify what they do and when. While this is a breaking change in blueprints-integration, it is a very simple task to migrate everything across.

The way preroll is applied has changed with this. Previously it would delay every other piece by a set amount, but it now moves the start of a given piece to be earlier. As part of this, it is expected that when the timeline is built a piece with `start: 0` may not start at 0 inside the partGroup. It is likely that every piece in the part will be offset to compensate for the preroll of any other piece, the transition, or even the outTransitionDuration of the previous part. (see the demo)

* **Other information**:

Renamed properties:
* `Part.disableOutTransition` -> `Part.disableNextPartInTransition`
* `Part.transitionPrerollDuration` -> `Part.inTransition.partContentDelayDuration` 
* `Part.transitionKeepaliveDuration` -> `Part.inTransition.previousPartKeepaliveDuration` 
* `Part.transitionDuration` -> `Part.inTransition.blockTakeDuration` 
* `Piece.isTransition` -> `Piece.pieceType: IBlueprintPieceType.InTransition`

Changed properties: 
* `Part.outTransition`. This is new as of #558 that this builds off
* `Piece.adlibPreroll` -> `Piece.prerollDuration`. Partially a rename, partially a new property that drives the preroll for the part
* `Part.prerollDuration` -> `Piece.prerollDuration` The big change that requires some work to update blueprints

TODO:

- [x] has timing of the rundown been messed up? some 'hacks' were removed
- [x] Unit tests to verify the timeline.ts changes produce sane looking piece/part layouts
- [x] Do adlibs get correct offsets and things
- [x] Fix blueprints unable to block take for a period after running an adlib-action
- [x] Docs

**Status**


baltedewit (approved):
Looks good, I would appreciate it if you could also update `meteor/server/api/playout/DOCS.MD`.

packages/blueprints-integration/src/rundown.ts
```diff
@@ -343,20 +361,24 @@ export interface IBlueprintPieceGeneric<TMetadata = unknown> {
...
 	 */
 	expectedPlayoutItems?: ExpectedPlayoutItemGeneric[]
-	/** When queued, should the adlib autonext */
-	adlibAutoNext?: boolean
-	/** When queued, how much overlap with the next part */
-	adlibAutoNextOverlap?: number
-	/** When queued, block transition at the end of the part */
-	adlibDisableOutTransition?: boolean
+	// /** When queued, should the adlib autonext */
+	// adlibAutoNext?: boolean
```
nytamin, line 377: Remove these?

meteor/server/api/blueprints/context/lib.ts
```diff
@@ -26,12 +26,12 @@ export const IBlueprintPieceSampleKeys = allKeysOfObject<IBlueprintPiece>({
...
 	lifespan: true,
-	adlibPreroll: true,
+	prerollDuration: true,
 	toBeQueued: true,
 	expectedPlayoutItems: true,
-	adlibAutoNext: true,
-	adlibAutoNextOverlap: true,
-	adlibDisableOutTransition: true,
+	// adlibAutoNext: true,
+	// adlibAutoNextOverlap: true,
```
nytamin, line 33: remove these?

meteor/server/api/playout/take.ts
```diff
@@ -56,15 +57,22 @@ export async function takeNextPartInnerSync(cache: CacheForPlayout, now: number)
 	const pBlueprint = pShowStyle.then(async (s) => loadShowStyleBlueprint(s))
 
 	if (currentPartInstance) {
-		const allowTransition = previousPartInstance && !previousPartInstance.part.disableOutTransition
+		const allowTransition = previousPartInstance && !previousPartInstance.part.disableNextInTransition
 		const start = currentPartInstance.timings?.startedPlayback
 
+		if (currentPartInstance.blockTakeUntil && currentPartInstance.blockTakeUntil > getCurrentTime()) {
+			// Adlib-actions can arbitrarily block takes from being done
+			const remainingMs = currentPartInstance.blockTakeUntil - getCurrentTime()
```
nytamin, line 65: Minor note: A better variable name would be `remainingTime`.

Also, store `getCurrentTime()` in a variable first, to ensure that it'll be the same value in the if-statement and here.

meteor/server/api/playout/playout.ts
```diff
@@ -1050,6 +1058,148 @@ export namespace ServerPlayoutAPI {
...
+		// ------------------------------
+		const timeline = cache.Timeline.findOne(cache.Studio.doc._id)
+		if (timeline) {
+			const timelineObjs = deserializeTimelineBlob(timeline.timelineBlob)
+			let tlChanged = false
+
+			_.each(results, (o) => {
+				check(o.id, String)
+
+				logger.info(`Timeline: Setting time: "${o.id}": ${o.time}`)
```
nytamin, line 1122: should probably be a `logger.debug` or `logger.verbose`.

The same applies to the orther `logger.into` further down

meteor/server/api/playout/timeline.ts
```diff
@@ -646,45 +658,50 @@ function generateCurrentInfinitePieceObjects(
...
 	}
+	if (pieceInstance.disabled || pieceInstance.piece.pieceType !== IBlueprintPieceType.Normal) {
+		// Can't be generated as infinites
+		return []
+	}
 
 	const infiniteGroup = createPartGroup(currentPartInfo.partInstance, {
 		start: `#${currentPartGroup.id}.start`, // This gets overriden with a concrete time if the original piece is known to have already started
-		duration: piece.piece.enable.duration || undefined,
+		// duration: piece.piece.enable.duration || undefined,
```
nytamin, line 676: remove?

nytamin (approved):
Looks good!
I added a few minor comments.

Since this is such a large change, we should be extra careful to test it in our testing-environment before deploying to prod.

packages/blueprints-integration/src/rundown.ts
Julusian: theyve already been removed in r39, will be tidied as part of merging r39 back into this branch

meteor/server/api/blueprints/context/lib.ts
Julusian: theyve already been removed in r39, will be tidied as part of merging r39 back into this branch

meteor/server/api/playout/playout.ts
Julusian: This method was moved from another file, so that isnt a new line. But Ive changed to logger.debug

meteor/server/api/playout/take.ts
Julusian: Ive made both changes
I think I didn't bother with storing it in a variable as the second use is only for a debug log so if its off by a couple ms it doesnt matter.

PR #620, opened 2021-11-09
fix: Queued segment looping until cleared manually

ianshade:
Bug fix in the Queued (Next) Segment feature

In various edge cases the queued segment starts looping infinitely (until cleared manually), or is queued an extra time.
Sample steps to reproduce the unwanted behavior:

  Case 1:
  - Take a Part in segment A
  - Queue segment B
  - Set first part of Segment B as next
  - Queue Segment A twice
  - Take - Segment A will loop until manually unqueued

  Case 2:

  - Take the first part in a segment (A) that has multiple parts
  - Queue a segment (B) that has only one part
  - Set as next segment B
  - Take - Segment A will loop until manually unqueued

  Case 3:

  - Take the first part in a segment A, that has multiple parts
  - Queue segment B, that has only one part
  - Take, Take (reaching B) - B will be set as next once more

  The causes of the issue: Wrong order of selecting next part and clearing `nextSegmentId` during a take. Not applying `consumesNextSegmentId` to reused partInstances. Not clearing `nextSegmentId` when user queues a segment while being on the last part of the current segment.

Queued segment no longer loops infinitely, nor is queued more times than expected.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/playout/lib.ts
+++ b/meteor/server/api/playout/lib.ts
@@ -245,10 +245,20 @@ export async function setNextPart(
 		let newInstanceId: PartInstanceId
 		if (newNextPartInstance) {
 			newInstanceId = newNextPartInstance._id
+			cache.PartInstances.update(newInstanceId, {
+				$set: {
+					consumesNextSegmentId: newNextPart?.consumesNextSegmentId,
+				},
+			})
 			await syncPlayheadInfinitesForNextPartInstance(cache)
 		} else if (nextPartInstance && nextPartInstance.part._id === nextPart._id) {
 			// Re-use existing
 			newInstanceId = nextPartInstance._id
+			cache.PartInstances.update(newInstanceId, {
+				$set: {
+					consumesNextSegmentId: newNextPart?.consumesNextSegmentId,
+				},
+			})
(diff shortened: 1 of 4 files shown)


nytamin:
Thanks for this, @ianshade !
I was able to reproduce Case 2 and 3 locally, and verify that this PR does indeed fix them.

fyi: For Case 1, I wasn't able to reproduce the indefinitely looping segment (on our `release38`-branch). The Segment A is looped just once (as expected), then the next segment is Nexted.

nytamin (approved):

meteor/server/api/playout/lib.ts
```diff
@@ -245,10 +245,20 @@ export async function setNextPart(
 		let newInstanceId: PartInstanceId
 		if (newNextPartInstance) {
 			newInstanceId = newNextPartInstance._id
+			cache.PartInstances.update(newInstanceId, {
+				$set: {
+					consumesNextSegmentId: newNextPart?.consumesNextSegmentId,
```
Julusian, line 250: we know in this route that `newNextPart` is always undefined, so this could be expressed as undefined/false instead

Julusian (approved):
One minor thing, but I shall deal with that as I am going to quickly improve the typings in this area

PR #622, opened 2021-11-12
Fix: protect against blueprint throws

nytamin:
Roboustness improvement

If blueprints crash, they might impact various operations in Core badly.

Crashes in Blueprints are caught and handled properly.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/ingest/generation.ts
+++ b/meteor/server/api/ingest/generation.ts
@@ -1,4 +1,4 @@
-import { BlueprintResultRundown } from '@sofie-automation/blueprints-integration'
+import { BlueprintResultRundown, BlueprintResultSegment, NoteSeverity } from '@sofie-automation/blueprints-integration'
 import { Meteor } from 'meteor/meteor'
 import { Random } from 'meteor/random'
 import { ReadonlyDeep } from 'type-fest'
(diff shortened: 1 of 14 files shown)


meteor/server/api/ingest/generation.ts
```diff
@@ -128,11 +128,45 @@ export async function calculateSegmentsFromIngestData(
...
+							message: {
+								key: 'Internal Error generating segment',
+								namespaces: [unprotectString(blueprint.blueprintId)],
+							},
+							origin: {
+								name: '', // TODO
+							},
+						},
+					],
+					name: ingestSegment.externalId,
```
Julusian, line 160: that should probably be `name: ingestSegment.name`

meteor/server/api/ingest/generation.ts
```diff
@@ -128,11 +128,45 @@ export async function calculateSegmentsFromIngestData(
...
+				})
+				res.segments.push(newSegment)
+
+				continue // Don't generate any content for the faulty segment
```
Julusian, line 164: Is the correct behaviour here to remove all the faulty parts, or should we preserve the last 'good' copy of them?

I have a feeling that if this does succeed in removing all the old parts, then the we will set the segment as hidden and so it will disappear, which is probably not desirable. I dont know if the note will still get shown.

But maybe this is fine, as the blueprints could do proper error handling instead. In our blueprints getSegment already does this for each part generation, where if any part throws an error while generating it will be replaced by an invalid part

Julusian (commented):
Looking through, there is only one case which concerns me with what happens to the rundown/playout.

Some others will cause some weird behaviour or playout glitches, but nothing too major

meteor/server/api/ingest/generation.ts
nytamin: I'm not 100%, but I'm leaning towards that this is fine.
I checked just now and the empty segment isn't hidden and the note shows so that's good.

As you say, this is protecting against an edge-case, if the blueprints handles themselves properly this'll never happen.

PR #623, opened 2021-11-16
fix: sort Adlibs on part._rank, then on piece._rank

nytamin:
Bug fix

The Adlibs appear in the wrong order in the Shelf & Dashboard

Sorting of adlibs take into account the _rank of their parts.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Shelf/AdLibPanel.tsx
+++ b/meteor/client/ui/Shelf/AdLibPanel.tsx
@@ -6,7 +6,7 @@ import { useTranslation, withTranslation } from 'react-i18next'
 import { Rundown, RundownId } from '../../../lib/collections/Rundowns'
 import { RundownPlaylist } from '../../../lib/collections/RundownPlaylists'
 import { Segment, DBSegment, SegmentId } from '../../../lib/collections/Segments'
-import { PartId } from '../../../lib/collections/Parts'
+import { DBPart, PartId } from '../../../lib/collections/Parts'
 import { AdLibPiece, AdLibPieces } from '../../../lib/collections/AdLibPieces'
 import { AdLibListItem, IAdLibListItem } from './AdLibListItem'
 import ClassNames from 'classnames'
@@ -548,16 +548,18 @@ export function fetchAndFilter(props: Translated<IAdLibPanelProps>): AdLibFetchA
 
 	const segments = props.playlist.getSegments()
 
-	const { uiSegments, liveSegment, uiPartSegmentMap } = memoizedIsolatedAutorun(
+	const { uiSegments, liveSegment, uiPartSegmentMap, uiPartMap } = memoizedIsolatedAutorun(
 		(currentPartInstanceId: PartInstanceId | null, nextPartInstanceId: PartInstanceId | null, segments: Segment[]) => {
 			// This is a map of partIds mapped onto segments they are part of
 			const uiPartSegmentMap = new Map<PartId, AdlibSegmentUi>()
+			const uiPartMap = new Map<PartId, DBPart>()
 
 			if (!segments) {
 				return {
 					uiSegments: [],
 					liveSegment: undefined,
 					uiPartSegmentMap,
+					uiPartMap,
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


jstarpl (changes requested):
I think this line: https://github.com/nrkno/tv-automation-server-core/blob/7fed0d607e00caebf90a8ca62d17fda1997d4159/meteor/lib/api/triggers/actionFilterChainCompilers.ts#L665 also needs to be updated so that these two sorting algorithms align. Otherwise, we'll get unexpected results.

nytamin:
Good catch @jstarpl ! I pushed an update that unifies the sorting in the two places, could you have another look plz?

meteor/lib/api/triggers/actionFilterChainCompilers.ts
```diff
@@ -598,7 +598,7 @@ export function compileAdLibFilter(
 		}
 
 		{
-			let skip = adLibPieceTypeFilter.skip
+			let skip = adLibPieceTypeFilter.skip // TODO: should this be adLibActionTypeFilter  ??
```
jstarpl, line 601: Yeah, that's a good catch. This should also be changed in lines 615 and 633.

meteor/lib/api/triggers/actionFilterChainCompilers.ts
```diff
@@ -651,32 +651,54 @@ export function compileAdLibFilter(
 			}
 		}
 
-		let result: IWrappedAdLib[] = []
+		const partMap = new Map<PartId, DBPart>()
+		{
+			if (partFilter?.length) {
+				Parts.find({
```
jstarpl, line 657: This will be terribly reactive and we have a lot of these filters running at the same time. For one, I think we need to use fields here to only react to the _part changing, and then, I think this needs to be wrapped into a `memoizedIsolatedAutorun` so that this is only calculated once for a given `partFilter`, which many of these filters will share.

meteor/lib/api/triggers/actionFilterChainCompilers.ts
```diff
@@ -651,32 +651,54 @@ export function compileAdLibFilter(
...
+
+			// Sort the adliba:
+			adlibs = sortAdlibs(
+				adlibs.map((adlib) => ({
+					adlib: adlib,
+					label: adlib.label,
+					adlibRank: adlib._rank,
+					adlibId: adlib._id,
+					partRank: (adlib.partId && partMap.get(adlib.partId))?._rank ?? null,
+					segmentRank: 0, // TODO: should we take the segment rank into account?
```
jstarpl, line 685: I'm afraid we do (otherwise the result will be different than the ultimate result in the AdLibPanel.tsx) and then on top of that, we also need to consider `Rundown._rank` here as well - which we didn't have to in the AdLibPanel.tsx, because that already operated on a sorted list of Rundowns and Segments.

meteor/lib/api/triggers/actionFilterChainCompilers.ts
```diff
@@ -651,32 +651,54 @@ export function compileAdLibFilter(
...
+					partMap.set(part._id, part)
+				})
+			}
+		}
+
+		let adlibs: IWrappedAdLib[] = []
 
 		{
-			result = [
+			adlibs = [
```
jstarpl, line 668: If you insist on renaming this variable, let's at least rename it to `matchedAdLibs` or `joinedAdLibs`. `adlibs` is very generic, and to me, `result` at least implied "this is where we'll keep the final result of this operation (of matching adlibs)". Now, this implies "this is where we'll keep some adlibs in this operation (of matching adlibs)", obscuring the point of this variable.

jstarpl (changes requested):

meteor/lib/api/triggers/actionFilterChainCompilers.ts
nytamin: Good point!
yeah `results` seemed to generic to me, but I get your point too! I'll rename it to `resultingAdlibs`, which should suit us both :)

meteor/lib/api/triggers/actionFilterChainCompilers.ts
jstarpl: I've fixed this issue.

jstarpl (approved):
Looks good.

PR #627, opened 2021-11-24
fix: reset instances from previous activation or not present in cache

ianshade:
Bug fix

  - Not all rehearsal pieceInstances are removed when resetting the rundown.
  - Not all pieceInstances are reset when resetting the rundown.
  - Not all pieceInstances from previous instances of the nexted part are reset when setting as next.
  - No previous partInstances or pieceInstances are reset when rundown is activated/reset after being deactivated.

  When using a rundown playlist that is never deleted, but the content is replaced on a daily basis, it leads to nonreset pieceInstances and partInstances piling up in the database, being hard to dispose of by cron jobs or filter out when storing snapshots.

Use direct database access when needed to remove or reset pieceInstances that might not be in cache.
When playlist is inactive, load nonreset partInstances to the cache, so that they can be reset.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/playout/__tests__/playout.test.ts
+++ b/meteor/server/api/playout/__tests__/playout.test.ts
@@ -69,6 +69,9 @@ describe('Playout API', () => {
 	function getAllPartInstances() {
 		return PartInstances.find({}).fetch()
 	}
+	function getAllPieceInstances() {
+		return PieceInstances.find({}).fetch()
+	}
 	function getAllPieceInstancesForPartInstance(partInstanceId: PartInstanceId) {
 		return PieceInstances.find({
 			partInstanceId: partInstanceId,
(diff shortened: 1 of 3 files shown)


Julusian (changes requested):
I noted a couple of points where you have opted to directly update the database instead of the Cache.
I know that is necessary because the PieceInstances are not in the cache, but for the ones which are in the cache I think it would be best to keep those updated. Skimming the code like this, I cant be sure that some later step isn't relying on the cache being up to date

Ideally I would like to see it split the updates across both. We don't currently track what PartInstanceIds we loaded the PieceInstances for, but perhaps we could? (keeping it updated as the cache changes might be too risky)

@nytamin what do you think?

ianshade:
@Julusian I think it is less error-prone now. Cache is updated for all the matching documents that are there. Only the matching pieceInstances not present in the cache will be updated/removed directly. Unless we add a way to dynamically pull more data into the cache after it has been initialised, I think this is solid.

Julusian (approved):

PR #628, opened 2021-11-24
chore: fix pre-commit script to work on Linux and macOS

ianshade:
Pre-commit script fix

Failing on Linux and macOS.

Use $OSTYPE to detect the OS.
Force meteor to run in x86 on ARM Macs. (to be removed when upgrading to meteor that supports arm64) 

* **Other information**:
Tested only on Windows and macOS, not tested on Linux.

**Status**

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -4,10 +4,10 @@
 # lint meteor
 cd meteor
 
-if [[ "$OSTYPE" == "msys"* ]] || [[ "$OSTYPE" == "cygwin"* ]]; then
+if echo "$OSTYPE" | grep -E -q "msys|cygwin" ; then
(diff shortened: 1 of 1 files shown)


nytamin:
To the reviewers: plz comment below with which platforms you have tested this on

- [ ]  Windows
- [ ] MacOS
- [ ] Linux

Julusian:
Not working on linux 
image

ianshade:
TIL that [[ is a bash thing, not a POSIX standard. And bash is not the default shell on Ubuntu. I chose to stick to the default shell and use a different method of comparing strings.

Julusian:
It works on linux now

nytamin (approved):
Works on Windows 👍

PR #630, opened 2021-11-26
chore: add default VS Code settings

jstarpl:

Workflow improvement

Project doesn't come with any default VS Code settings.

We provide a sample VS Code Settings JSON to make bootstrapping easier for new developers.

* **Other information**:

**Status**

### DIFF
--- a/.vscode/settings.json.default
+++ b/.vscode/settings.json.default
@@ -1,0 +1,11 @@
+{
+    "eslint.workingDirectories": [
+        "meteor",
+        "packages/blueprints-integration",
+        "packages/mos-gateway",
+        "packages/playout-gateway",
+        "packages/server-core-integration"
+    ],
+    "prettier.enable": true,
+    "sonarlint.connectedMode.project": { "projectKey": "nrkno_tv-automation-server-core" },
+}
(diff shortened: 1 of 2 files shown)


baltedewit (approved):
Should probably be seen in parallel with https://github.com/nrkno/tv-automation-server-core/pull/629

.vscode/settings.json.default
```diff
@@ -0,0 +1,11 @@
...
+    "eslint.workingDirectories": [
+        "meteor",
+        "packages/blueprints-integration",
+        "packages/mos-gateway",
+        "packages/playout-gateway",
+        "packages/server-core-integration"
+    ],
+    "prettier.enable": true,
+    "sonarlint.connectedMode.project": { "projectKey": "nrkno_tv-automation-server-core" },
+}
```
nytamin, line 11: I also use `"editor.formatOnSave": true,`, should we add that too?

nytamin (approved):

.vscode/settings.json.default
jstarpl: I think we've agreed that that's a user preference and the default config is supposed to be minimal. Git Hooks are supposed to be the guard for code style, so I think we should leave the decision on when to format to the user.

gundelsby (approved):

Julusian (approved):

PR #631, opened 2021-12-06
fix: Hash triggers for core migrations

thomasslee97:

Bug fix.

It's currently not possible for System migrations to change / remove the default keyboard shortcuts created by core migrations.

The Id values of the default migrations are now hashed when they are created, making them match up with the Ids that the migration context will be searching for. A migration has also been added to migrate the existing actions to the new Id values, so that this change will take effect when R38 is deployed.

* **Other information**:

**Status**


jstarpl (approved):
Looks good, thanks!

PR #633, opened 2021-12-09
feat: remove transformed collections

Julusian:

performance/code quality

Some collection documents are transformed into classes instead of interfaces when being loaded from the db.
This results in an iteration over every propery on the documents, and some copying, and so will be a small performance penalty.

Instead of this, we should move the methods off the classes and feed them the interface form as a parameter.
The classes can then be removed, as they serve no purpose. 

* **Other information**:

This has been split into many many commits, to aid in reviewing. It might be best to not squash when merging to keep this readability instead of a large touching everything commit

**Status**

### DIFF
--- a/meteor/__mocks__/helpers/lib.ts
+++ b/meteor/__mocks__/helpers/lib.ts
@@ -1,6 +1,6 @@
 import * as _ from 'underscore'
 import { LogLevel, ProtectedString } from '../../lib/lib'
-import { AsyncTransformedCollection } from '../../lib/collections/lib'
+import { AsyncMongoCollection } from '../../lib/collections/lib'
 import { getLogLevel, setLogLevel } from '../../server/logging'
 
 /*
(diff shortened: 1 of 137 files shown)


nytamin (changes requested):
I don't think it's a good idea with so many renames, as that would create massive merge-conflicts for for example TV2 when pulling these changes.

Would it be possible to achieve the same performance gains without so many lines changed?

Julusian:
@nytamin To see how bad it will be for TV2, I have done a crude simulation of the merge conflict they will end up with.
I took a branch that looked to be most up to date on nrk changes, and updated it to release38. (ignoring all merge conflicts)
Then I merged this branch into that, and looked at what new conflicts arose.

Quite a few files had new conflicts:
```

# Conflicts:
#	meteor/client/ui/RundownView.tsx - method renaming
#	meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx - method renaming
#	meteor/client/ui/Shelf/AdLibPanel.tsx - import
#	meteor/client/ui/Shelf/AdLibRegionPanel.tsx - method renaming & import
#	meteor/client/ui/Shelf/BucketPanel.tsx - import
#	meteor/client/ui/Shelf/TimelineDashboardPanel.tsx - import
#	meteor/lib/api/triggers/actionFilterChainCompilers.ts - import
#	meteor/lib/collections/PartInstances.ts - code change
#	meteor/lib/collections/RundownPlaylists.ts - code change
#	meteor/lib/collections/Segments.ts - code change
#	meteor/lib/rundown/__tests__/rundownTiming.test.ts - import
#	meteor/server/api/ingest/__tests__/ingest.test.ts - code change
#	meteor/server/api/ingest/generation.ts - code change
#	meteor/server/api/ingest/rundownInput.ts - import
#	meteor/server/api/playout/adlib.ts - code change
#	meteor/server/api/playout/lib.ts - import
#	meteor/server/api/rundownPlaylist.ts - code change
#	meteor/server/api/userActions.ts - code change
#	meteor/server/cache/CacheCollection.ts - code change
```
a majority of those are due to the changed imports, or the methods having moved. Only a few are related to the interface renaming. There are a few new uses of the old names that will need to be updated, but typescript will make that easy, to track down (up to 48 cases of the wrong naming being used, merge conflicts add some uncertainty to that number)

From this, I don't think that removing the bulk renames from this PR will make merging much easier. But I am open to removing those commits anyway

Julusian:
@nytamin I have removed the commits which did the bulk renames. They didnt do anything for performance, solely code readability and tidiness.

It looks like some of those lines remain, as they got pulled back in by rebase conflicts

nytamin (approved):
Looks good!

PR #634, opened 2021-12-10
Fix/user actions gateway duration

nytamin:
Bug fix

The timelineHash was looked up in the Timeline collection, but this turned out to be quite fragile as the timeline often has changed before the Playout-Gateway sends its callback (due to the "now"-feature).

The timelineHash  is now stored on the UserAction instead. Since this is persistant, it can be looked up even if the timeline has changed.

This will give more cases in the userActions log where the full times are filled out.

* **Other information**:

**Status**

### DIFF
--- a/meteor/lib/collections/UserActionsLog.ts
+++ b/meteor/lib/collections/UserActionsLog.ts
@@ -3,6 +3,7 @@ import { createMongoCollection } from './lib'
 import { UserId } from './Users'
 import { OrganizationId } from './Organization'
 import { registerIndex } from '../database'
+import { TimelineHash } from './Timeline'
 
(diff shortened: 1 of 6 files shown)


Julusian (commented):
Im not entirely happy with how and where we match up the user log entry with the timelinehash. But not enough to stop this being merged.

It wont be the prettiest, but I wonder if we should be passing in the userlog to the call to updateTimeline? 
I would actually prefer passing the timelineHash out back up to the userActionsApi level to track it, but I suspect that will become too much of a race condition.
Im partly thinking about the core-worker here, as that currently has no awareness of the user actions log, and im not sure it should

PR #638, opened 2021-12-15
Contribute: bugfixes release38

thomasslee97:

Bug fixes.

- Segment can get stuck highlighted as next (reactivity bug).
- The indicator for a queued segment does not get shown if the currently on-air segment is queued.
- The queued segment doesn't get cleared when deactivating.
- The queued segment doesn't get cleared if the queued segment is set as next.
- The on-air segment can be hidden if blueprints return an empty array of parts (e.g. the segment is "floated" in an NRCS while on-air).
- A "missing media" notification may be shown for a media item while scrolling through a rundown due to a bug where the piece has been loaded, but the package / media object has not.
- Deleting a part that is in next does not remove the part instance in next (a bug from TV 2's perspective).

- All bugs above have been fixed
- When the part in next is deleted, a blueprint function `shouldRemoveOrphanedPartInstance` will be called to check whether the part instance associated with that part should be deleted. This allows blueprints to control when an orphaned, next, part instance should be deleted. This method is optional, if not implemented in blueprints then the current behaviour will be used (the part instance will not be deleted).

**Other information**:

**Status**


jstarpl:
@thomasslee97 Before we get to do a proper review, I can already point you towards fixing this: https://github.com/nrkno/tv-automation-server-core/runs/4549682600?check_suite_focus=true

meteor/server/api/ingest/commit.ts
```diff
@@ -413,7 +486,6 @@ function canRemoveSegment(
 		(nextPartInstance?.segmentId === segmentId && isTooCloseToAutonext(currentPartInstance, false))
 	) {
 		// Don't allow removing an active rundown
-		logger.warn(`Not allowing removal of current playing segment "${segmentId}", making segment unsynced instead`)
```
jstarpl, line 416: Why is this being removed?

meteor/server/api/ingest/generation.ts
```diff
@@ -244,11 +243,6 @@ export async function calculateSegmentsFromIngestData(
 				})
 				res.parts.push(part)
 
-				// This ensures that it doesn't accidently get played while hidden
-				if (blueprintSegment.segment.isHidden) {
-					part.invalid = true
```
jstarpl, line 249: Why is this being removed? Seems dangerous.

meteor/server/api/ingest/generation.ts
```diff
@@ -274,11 +268,6 @@ export async function calculateSegmentsFromIngestData(
 					)
 				)
 			})
-
-			// If the segment has no parts, then hide it
-			if (blueprintSegment.parts.length === 0) {
-				newSegment.isHidden = true
```
jstarpl, line 280: I understand that this removal refers to "The on-air segment can be hidden if blueprints return an empty array of parts (e.g. the segment is "floated" in an NRCS while on-air).", but it seems like this should be handled some other way then? I really don't think we want to have empty Segments visible in the GUI.

jstarpl (commented):
The RundownNotifier/SegmentTimelineContainer fixes look good to me. I've got some questions regarding the ingest changes, which to me seem to not just fix bugs but rather change the intended result, but I'm not an expert there.

meteor/server/api/ingest/generation.ts
Julusian: This was moved to the commit stage, to allow for the segment to be kept visible if it is still playing
https://github.com/nrkno/tv-automation-server-core/pull/638/files#diff-432a5ee01b84991e0799a0ccb9960dbad910336d7a5f4a5494feb0fb1f971e07R276

meteor/server/api/ingest/generation.ts
Julusian: This was also moved to the commit stage https://github.com/nrkno/tv-automation-server-core/pull/638/files#diff-432a5ee01b84991e0799a0ccb9960dbad910336d7a5f4a5494feb0fb1f971e07R281

meteor/server/api/ingest/commit.ts
```diff
@@ -200,20 +204,87 @@ export async function CommitIngestOperation(
...
+
+								// Blueprints have updated the hidden segment, so we won't try to preserve the contents
+								if (newParts.length) {
+									continue
+								}
+
+								// Restore old data
+								const oldParts = Parts.find({ segmentId }).fetch()
+								const oldPartIds = oldParts.map((part) => part._id)
+								const oldPieces = Pieces.find({ startPartId: { $in: oldPartIds } }).fetch()
```
Julusian, line 248: It would be good to do these in parallel. Also using the newer promise syntax would be nice to do

meteor/server/api/ingest/commit.ts
```diff
@@ -200,20 +204,87 @@ export async function CommitIngestOperation(
...
+								for (const adLib of oldAdLibPieces) {
+									ingestCache.AdLibPieces.insert(adLib)
+								}
+								for (const action of oldAdLibActions) {
+									ingestCache.AdLibActions.insert(action)
+								}
+							}
+						}
+
+						// Protect live segment from being hidden
```
Julusian, line 267: This is just one of the routes this block can take, either move the line to that route or update it here

meteor/server/api/ingest/shouldRemoveOrphanedPartInstance.ts
```diff
@@ -0,0 +1,88 @@
...
+	cache: CacheForPlayout,
+	showStyle: ReadonlyDeep<ShowStyleCompound>,
+	blueprint: ReadonlyDeep<ShowStyleBlueprintManifest>,
+	rundown: ReadonlyDeep<Rundown>
+): Promise<void> {
+	if (!cache.Playlist.doc.activationId) return
+	if (!blueprint.shouldRemoveOrphanedPartInstance) return
+
+	const playlistPartInstances = getSelectedPartInstancesFromCache(cache)
+	if (!playlistPartInstances.nextPartInstance?.orphaned) return
```
Julusian, line 25: Are adlibbed instances taking the correct route here?
I think they ought to be caught in this return too

packages/blueprints-integration/src/api.ts
```diff
@@ -148,6 +149,12 @@ export interface ShowStyleBlueprintManifest extends BlueprintManifestBase {
 		playoutStatus: 'current' | 'next'
 	) => void
 
+	/**
+	 * Allows the blueprint to remove the next part instance if it has become orphaned.
+	 * This call will only be made if the part instance has been orphaned.
+	 */
+	shouldRemoveOrphanedPartInstance?: (context: IRemoveOrphanedPartInstanceContext, partInstance: BlueprintRemoveOrphanedPartInstance) => void
```
Julusian, line 156: I think this method would make more sense to return a boolean, rather than to call methods on the context to set the result

meteor/server/api/ingest/shouldRemoveOrphanedPartInstance.ts
```diff
@@ -0,0 +1,88 @@
...
+		blueprint.shouldRemoveOrphanedPartInstance(orphanedPartInstanceContext, clone(existingResultPartInstance))
+	} catch (e) {
+		logger.error(e)
+	}
+
+	// Save notes:
+	if (!orphanedPartInstance.part.notes) orphanedPartInstance.part.notes = []
+	const notes: PartNote[] = orphanedPartInstance.part.notes
+	let changed = false
+	for (const note of orphanedPartInstanceContext.notes) {
```
Julusian, line 57: Not something to handle with this change, but we really need to improve the note handling from many blueprint calls. We keep appending to the PartInstance notes, and never removing old ones

meteor/server/api/ingest/shouldRemoveOrphanedPartInstance.ts
```diff
@@ -0,0 +1,88 @@
...
+				type: note.type,
+				message: note.message,
+				origin: {
+					name: '',
+				},
+			})
+		)
+	}
+	if (changed) {
+		cache.PartInstances.update(orphanedPartInstance._id, {
```
Julusian, line 70: Is this the right thing to do? It is likely that the next instance will be removed, so the notes will be discarded

packages/blueprints-integration/src/context.ts
```diff
@@ -232,6 +232,10 @@ export interface ISyncIngestUpdateToPartInstanceContext extends IRundownUserCont
 	updatePartInstance(props: Partial<IBlueprintMutatablePart>): IBlueprintPartInstance
 }
 
+export interface IRemoveOrphanedPartInstanceContext extends IRundownUserContext {
```
Julusian, line 235: Should this be a UserContext?
Does it make sense to be able to show the user notes from uses of this?

meteor/server/api/playout/actions.ts
```diff
@@ -165,6 +165,7 @@ export async function deactivateRundownPlaylistInner(cache: CacheForPlayout): Pr
 			previousPartInstanceId: null,
 			currentPartInstanceId: null,
 			holdState: RundownHoldState.NONE,
+			nextSegmentId: undefined,
```
Julusian, line 168: This should probably be in the $unset block below

meteor/server/api/playout/playout.ts
```diff
@@ -364,6 +364,17 @@ export namespace ServerPlayoutAPI {
 			if (!nextPart) throw new Meteor.Error(404, `Part "${nextPartId}" not found!`)
 		}
 
+		// If we're setting the next point to somewhere other than the current segment, and in the queued segment, clear the queued segment
```
Julusian, line 367: A question, could this be simplified down to always clear if setManually is true? ie, if the user specified where the next is rather than sofie choosing

Julusian (changes requested):

meteor/server/api/ingest/commit.ts
thomasslee97: It has been moved to here, as `canRemoveSegment` is used in multiple places where that message isn't accurate https://github.com/nrkno/tv-automation-server-core/blob/4fa9cc1cebc8fe896f3b8dbcfbc27414060e779b/meteor/server/api/ingest/commit.ts#L222
e.g. it is used here https://github.com/nrkno/tv-automation-server-core/blob/4fa9cc1cebc8fe896f3b8dbcfbc27414060e779b/meteor/server/api/ingest/commit.ts#L270

meteor/server/api/ingest/commit.ts
thomasslee97: Fixed by https://github.com/tv2/tv-automation-server-core/commit/bd0d3682e1f3695caf978570594ac12699498468

meteor/server/api/ingest/commit.ts
thomasslee97: Moved into the correct block https://github.com/tv2/tv-automation-server-core/commit/dc59ba7c34ef3dbea3dd735ff73d3416cdcf912f

meteor/server/api/ingest/shouldRemoveOrphanedPartInstance.ts
thomasslee97: If `orphaned` is set then it's either deleted or adLibbed, so this should be correct

packages/blueprints-integration/src/api.ts
thomasslee97: Changed https://github.com/tv2/tv-automation-server-core/commit/584e649d856b353a1e63dd305b852bbe6b14b702

meteor/server/api/ingest/shouldRemoveOrphanedPartInstance.ts
thomasslee97: My assumption here is that the blueprints are likely to only ever push notes in the case that the isntance is not being removed. Pushing a not inside `shouldRemoveOrphanedPartInstance` and then removing the part instance sounds like contradictory behaviour. What would a note in that situation say? "Warning: I will remove this part"?

packages/blueprints-integration/src/context.ts
thomasslee97: I think it might make sense to push something like "X has been orphaned, this may cause Y effect", depending on the installation specifics. e.g. Orphaning means that some metadata about the show won't be recorded. e.g. "You may want to reload your script to ensure [Some Media Server] is happy"

meteor/server/api/playout/actions.ts
thomasslee97: Fixed by https://github.com/tv2/tv-automation-server-core/commit/49ffd5df5f235b0269bc2d960da7721aab45e699

meteor/server/api/playout/playout.ts
thomasslee97: I don't think we should clear the queued segment every time the user moves next, what if they just need to skip over the next part in the current Segment? e.g. An external feed has been lost, so they are skipping over an interview. If I was the user I'd want my queued selection to be kept in that scenario.

ianshade:
@Julusian, FYI, more fixes for the live segment hiding issue are coming soon

meteor/server/api/ingest/shouldRemoveOrphanedPartInstance.ts
Julusian: I dont know what my logic behind this comment was...

meteor/lib/rundownNotifications.ts
```diff
@@ -142,10 +142,19 @@ export function getBasicNotesForSegment(
...
+			case SegmentOrphanedReason.DELETED:
+				baseMessage = 'Segment no longer exists in {{nrcs}}'
+				break
+			case SegmentOrphanedReason.HIDDEN:
+				baseMessage = 'Segment was hidden in {{nrcs}}'
+				break
+		}
 		notes.push({
 			type: NoteSeverity.WARNING,
-			message: generateTranslation('Segment no longer exists in {{nrcs}}', {
```
Julusian, line 148: This wont generate translations anymore. The i18n parser cant follow the routes, it needs the strings to be fixed as parameters to the function. 
Make baseMessage be `ITranslatableMessage` instead

Julusian:
> @Julusian, FYI, more fixes for the live segment hiding issue are coming soon

OK, let me know when you are happy for me to review your additions

meteor/server/api/ingest/commit.ts
```diff
@@ -277,14 +283,20 @@ export async function CommitIngestOperation(
 								if (!canRemoveSegment(currentPartInstance, nextPartInstance, segmentId)) {
 									// Protect live segment from being hidden
 									logger.warn(`Cannot hide live segment ${segmentId}, it will be orphaned`)
+									const oldSegment = Segments.findOne(segmentId, {
```
Julusian, line 286: Doing a mongo query in a loop like this doesnt feel great.
I know that it will only execute up to twice, but it would still be good to do them in parallel.
Perhaps this could be all at once inside the  `if (orphanHiddenSegmentIds.size) {` block (Line ~319)?

Also, as orphanedHiddenSegmentPropertiesToPreserve is currently an empty object, it would be nice to skip the mongo query when that is the case (eg `Object.keys(orphanedHiddenSegmentPropertiesToPreserve).length > 1 ? Segments.findOne(....) : undefined`)

meteor/server/api/ingest/lib.ts
```diff
@@ -127,7 +127,7 @@ export function canSegmentBeUpdated(
 	}
 
 	if (!segment) return true
-	if (segment.orphaned && !isCreateAction) {
+	if (segment.orphaned === 'deleted' && !isCreateAction) {
```
Julusian, line 130: It would be good if this also used the enum

Julusian (changes requested):
Just a couple of small bits

ianshade:
The fixes related to queued segment introduce a regression that can be reproduced with these steps:

1. Queue a segment X
2. Jump around using Move Next triggers (F9, F10...).
If you hit segment X, it will be unqueued due to being set as next, no matter if you take it or not.

Before, you could freely move next with the shortcuts, and retain the queued segment. 
Does that break anything for NRK?

nytamin:
> Before, you could freely move next with the shortcuts, and retain the queued segment. Does that break anything for NRK?

I don't think this will be an issue for us. We rarely use that feature, and the consequences are mild since one can just re-queue the segment if needed.

Julusian:
@ianshade please open a new pr for any additional fixes. probably targetting r39, unless there are major issues found in r38

PR #642, opened 2021-12-21
Fix: expetedPackages from snapshot

nytamin:
Bug fix

When loading a RundownPlaylist snapshot, the expected packages aren't loaded.

ExpectedPackages are now loaded into the collection.

* **Other information**:
I've also done some refactoring in the restore snapshot function, to strengthen the types a bit.

**Status**

### DIFF
--- a/meteor/server/api/snapshot.ts
+++ b/meteor/server/api/snapshot.ts
@@ -671,10 +671,10 @@ export async function restoreFromRundownPlaylistSnapshot(
 	// Migrate old data:
 	// 1.12.0 Release 24:
 	const partSegmentIds: { [partId: string]: SegmentId } = {}
-	_.each(snapshot.parts, (part) => {
+	for (const part of snapshot.parts) {
 		partSegmentIds[unprotectString(part._id)] = part.segmentId
-	})
-	_.each(snapshot.pieces, (piece) => {
+	}
+	for (const piece of snapshot.pieces) {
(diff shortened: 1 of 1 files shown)


meteor/server/api/snapshot.ts
```diff
@@ -681,74 +688,127 @@ export async function restoreFromRundownPlaylistSnapshot(
...
+	const rundownIdMap = new Map<RundownId, RundownId>()
 	const getNewRundownId = (oldRundownId: RundownId) => {
-		return rundownIdMap[unprotectString(oldRundownId)]
+		const rundownId = rundownIdMap.get(oldRundownId)
+		if (!rundownId) {
+			throw new Meteor.Error(500, `Could not find new rundownId for "${oldRundownId}"`)
+		}
+		return rundownId
 	}
 	_.each(snapshot.rundowns, (rd) => {
```
gundelsby, line 699: Since you're touching the code anyway I guess I'd like to see _.each replaced by for..of

gundelsby (commented):
I assume that replace the ids in the snapshot with new random ids is to make packages "new" in the eyes of PM, then?

nytamin:
> I assume that replace the ids in the snapshot with new random ids is to make packages "new" in the eyes of PM, then?

Not exactly, we change the ids of all objects we import in the snapshot. This is to avoid overwriting any existing rundown-data, and possibly cross-contaminate data between rundowns.

PR #646, opened 2022-01-04
feat: pass currentPartInstanceId into UserActions.take call

Julusian:

Bug fix

Users will sometimes press take multiple times close to each other when it looks like the first press did nothing.
If the second press is within 1s of the first it will be rejected.
If it is longer, then both takes will be performed

When the user does a take, the currentPartInstanceId is passed to the backend. This avoids the double-take problem as the currentPartInstance will have changed when the second one attempts to be executed and so will fail

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/RundownView.tsx
+++ b/meteor/client/ui/RundownView.tsx
@@ -477,7 +477,9 @@ const RundownHeader = withTranslation()(
 						},
 					})
 				} else {
-					doUserAction(t, e, UserAction.TAKE, (e) => MeteorCall.userAction.take(e, this.props.playlist._id))
+					doUserAction(t, e, UserAction.TAKE, (e) =>
+						MeteorCall.userAction.take(e, this.props.playlist._id, this.props.playlist.currentPartInstanceId)
+					)
 				}
 			}
 		}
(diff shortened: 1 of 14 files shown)


baltedewit (approved):

baltedewit:
I do wonder if this can be done entirely server side so the API can be a bit more concise?

Julusian:
@baltedewit 
> I do wonder if this can be done entirely server side so the API can be a bit more concise?

Possibly, but it will add extra complexity to the server side instead and changes the intention of the check a bit.

With it being in the api (which we do elsewhere for some methods, but not most https://github.com/nrkno/tv-automation-server-core/blob/master/meteor/lib/api/userActions.ts#L93), it becomes very explicit that the take action does a take out of the partInstance the user specifies.
If core was to do the check, then there would be a few ms more 'grey time', and while very rare I think there will still be some edge cases where a busy core will perform a double take.

That said, I am warming to the idea of keeping it all server-side. I was concerned for a bit that it wont play well with the worker-thread changes, or with the way we do locks/queues right now, but apart from requiring an extra load of the playlist from the db, I think it would be fine

PR #647, opened 2022-01-04
fix: use MeteorCall instead of Meteor.call for better type safety

Julusian:

code quality

Inside of core, (primarily inside unit tests) we use `Meteor.call(` to invoke methods as a gateway or client would.
This works, but is not at all type safe, so makes updating things hard.

For example, `rundownLayout.createRundownLayout` wasn't validating that `region` was a string or set. The unit tests were omitting that newly added parameter

Where possible `MeteorCall` has been used instead to make the calls be type safe

* **Other information**:

**Status**


jstarpl:
Wait, wasn't that the point? I think the tests specifically had the *API enums copied and not imported from `lib` so that the tests would verify that the names of the method didn't accidentally change.

Julusian:
Maybe, but they aren't verifying that method signatures hadn't changed, so does it matter that the name stays the same? 

If that is something we want to do then I shall  write some tests dedicated to that, but we will need to rules on how and when we can modify/extend/remove existing methods.

Let's make this be an agenda point for Friday

PR #648, opened 2022-01-05
fix: strict select of pieceinstances in timelineTriggerTime

baltedewit:
Only load pieceInstances that we know we the timings may need to be updated for, ie ones that come from the current or next part. This reduces data transferred from the database and reduces the search space for when we use the cache.

All piece instances in the currently active playlist are loaded.

* **Other information**:
I think the current and next part covers all piece instances that could possibly have a 'now' time that will be updated but do please correct me if I'm wrong here.

**Status**

### DIFF
--- a/meteor/server/api/playout/playout.ts
+++ b/meteor/server/api/playout/playout.ts
@@ -1085,23 +1085,18 @@ export namespace ServerPlayoutAPI {
 						activePlaylist,
 						PlayoutLockFunctionPriority.CALLBACK_PLAYOUT,
 						async () => {
-							const rundownIDs = Rundowns.find({ playlistId }, { fields: { _id: 1 } }).map(
-								(r) => r._id
-							)
+							const rundownIDs = Rundowns.find({ playlistId }, { fields: { _id: 1 } }).map((r) => r._id)
 							const partInstanceIDs = [activePlaylist.currentPartInstanceId].filter(
 								(id): id is PartInstanceId => id !== null
 							)
 
 							// We only need the PieceInstances, so load just them
-							const pieceInstanceCache = await DbCacheWriteCollection.createFromDatabase(
-								PieceInstances,
-								{
-									rundownId: { $in: rundownIDs },
-									partInstanceId: {
-										$in: partInstanceIDs,
-									},
-								}
-							)
+							const pieceInstanceCache = await DbCacheWriteCollection.createFromDatabase(PieceInstances, {
+								rundownId: { $in: rundownIDs },
+								partInstanceId: {
+									$in: partInstanceIDs,
+								},
+							})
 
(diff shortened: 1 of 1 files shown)


meteor/server/api/peripheralDevice.ts
```diff
@@ -256,12 +257,19 @@ export namespace ServerPeripheralDeviceAPI {
 							PlayoutLockFunctionPriority.CALLBACK_PLAYOUT,
 							async () => {
 								const rundownIDs = Rundowns.find({ playlistId }).map((r) => r._id)
+								const partInstanceIDs = [
```
Julusian, line 260: We shouldn't rely on `activePlaylist` inside the lock, as other playout operations could be done between loading that playlist and aquiring the lock causing us to use stale data.

meteor/server/api/peripheralDevice.ts
```diff
@@ -256,12 +257,19 @@ export namespace ServerPeripheralDeviceAPI {
...
+									activePlaylist.currentPartInstanceId,
+									activePlaylist.nextPartInstanceId,
```
Julusian, line 262: I suspect we can consider just the currentPartInstanceId.

next hasnt started playback, and so shouldnt have any adlib pieces which might need updating.
previous has finished playback (or just the overlap remaining) and so probably shouldnt have any left to update

meteor/server/api/peripheralDevice.ts
```diff
@@ -256,12 +257,19 @@ export namespace ServerPeripheralDeviceAPI {
 							PlayoutLockFunctionPriority.CALLBACK_PLAYOUT,
 							async () => {
 								const rundownIDs = Rundowns.find({ playlistId }).map((r) => r._id)
```
Julusian, line 259: If you want to optimise further, you could make this query load only the `_id` field and not the whole document

meteor/server/api/peripheralDevice.ts
baltedewit: Aren't we acquiring a lock & cache for that specific playlist anyway implying that if the playlist changes we are working with stale/wrong data anyway?

Julusian: I dont follow that

baltedewit: In line 256 the playlistId is used to load the cache, reusing that id inside the function that has the cache seems safe.

Julusian: using the id to load the cache is different to relying on properties on the playlist to not have changed

but I completely missed that this is inside of a `runStudioOperationWithCache` lock, so ignore me. that guarantees that nothing else will be done in the middle (I know, the locks in sofie at the moment are a bit confusing..)

Julusian (approved):

nytamin (approved):
Nice optimization!

PR #655, opened 2022-01-12
fix: dashboard panel scale affecting position

ianshade:
Bug fix (UI)

Setting a Scale other than 1 on some dashboard panels (NextInfoPanel and PieceCountdownPanel) affects their position change. Cause: changing fontSize on an element positioned using `em`.

Positioning on these panels is fixed to work exactly as other panels.

* **Other information**:
The additional variable (`--dashboard-panel-scale`) is not used by these panels, but used by TV 2's keyboard map. Can it stay there?

**Status**

### DIFF
--- a/meteor/client/ui/Shelf/NextInfoPanel.tsx
+++ b/meteor/client/ui/Shelf/NextInfoPanel.tsx
@@ -54,9 +54,11 @@ export class NextInfoPanelInner extends MeteorReactComponent<INextInfoPanelProps
 					}
 				)}
 			>
-				<span className="next-info-panel__name">{showAny && this.props.panel.name} </span>
-				{segmentName && <span className="next-info-panel__segment">{segmentName}</span>}
-				{partTitle && <span className="next-info-panel__part">{partTitle}</span>}
+				<div style={{ fontSize: isDashboardLayout ? 'var(--dashboard-panel-scaled-font-size)' : undefined }}>
+					<span className="next-info-panel__name">{showAny && this.props.panel.name} </span>
+					{segmentName && <span className="next-info-panel__segment">{segmentName}</span>}
+					{partTitle && <span className="next-info-panel__part">{partTitle}</span>}
+				</div>
 			</div>
 		)
 	}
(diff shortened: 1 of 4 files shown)


meteor/client/ui/Shelf/NextInfoPanel.tsx
```diff
@@ -54,9 +54,11 @@ export class NextInfoPanelInner extends MeteorReactComponent<INextInfoPanelProps
 					}
 				)}
 			>
-				<span className="next-info-panel__name">{showAny && this.props.panel.name} </span>
-				{segmentName && <span className="next-info-panel__segment">{segmentName}</span>}
-				{partTitle && <span className="next-info-panel__part">{partTitle}</span>}
+				<div style={{ fontSize: isDashboardLayout ? 'var(--dashboard-panel-scaled-font-size)' : undefined }}>
```
jstarpl, line 57: Wouldn't it make sense to move it to a class? I really don't like appending reactive styling using a CSS variable. It just mixes and mingles the CSS, JS code and JSX code everywhere. I think just using a CSS declaration here would be much preferable and cleaner.

meteor/client/ui/Shelf/PieceCountdownPanel.tsx
```diff
@@ -85,6 +85,7 @@ export class PieceCountdownPanelInner extends MeteorReactComponent<
 					className={ClassNames('piece-countdown-panel__timecode', {
 						overtime: Math.floor(this.state.displayTimecode / 1000) > 0,
 					})}
+					style={{ fontSize: isDashboardLayout ? 'var(--dashboard-panel-scaled-font-size)' : undefined }}
```
jstarpl, line 88: Same as above: why use inline style when using a CSS variable anyway?

jstarpl (changes requested):

meteor/client/ui/Shelf/NextInfoPanel.tsx
ianshade: I did something. It might not fit the CSS class naming convention perfectly, though. Maybe all the dashboard panels should have the `dashboard-panel` class in addition to their own like `next-info-panel`?

jstarpl (approved):

PR #660, opened 2022-01-17
Fix/media issues notification level

nytamin:
This PR does some refactoring of the mediaStatus-generation and introduces some changes:

* Added `PieceStatusCode.SOURCE_HAS_ISSUES` and uses that for the "black/freeze frames detected" notifications instead of the previous `PieceStatusCode.SOURCE_BROKEN`.

This affects the GUI, by reducing "black/freeze frames detected" notification-levels to `Notes`, instead of `Warnings`.

The motivation for this is to reduce the amount of warnings for the end-users.


meteor/client/lib/notifications/notifications.ts
```diff
@@ -477,13 +477,20 @@ export class Notification extends EventEmitter {
...
 
 export function getNoticeLevelForPieceStatus(statusCode: RundownAPI.PieceStatusCode): NoticeLevel | null {
-	return statusCode !== RundownAPI.PieceStatusCode.OK && statusCode !== RundownAPI.PieceStatusCode.UNKNOWN
-		? statusCode === RundownAPI.PieceStatusCode.SOURCE_NOT_SET
-			? NoticeLevel.CRITICAL
-			: // : innerPiece.status === RundownAPI.PieceStatusCode.SOURCE_MISSING ||
-			  // innerPiece.status === RundownAPI.PieceStatusCode.SOURCE_BROKEN
-			  NoticeLevel.WARNING
-		: null
+	if (statusCode === RundownAPI.PieceStatusCode.OK || statusCode === RundownAPI.PieceStatusCode.UNKNOWN) {
```
LindvedKrvang, line 480: Why not use a switch?

meteor/client/lib/notifications/notifications.ts
```diff
@@ -477,13 +477,20 @@ export class Notification extends EventEmitter {
...
+		return null
```
LindvedKrvang, line 481: Returning null is generally a bad practice, but I see that the code allready did that before the refactor.

nytamin: yes, I changed as little as possible with the refactor.

meteor/client/lib/notifications/notifications.ts
nytamin: Good point! I'll switchify it!

jstarpl (approved):
Looks good.

PR #667, opened 2022-01-25
chore: light run script

jstarpl:

Workflow improvement

Currently, starting up Core for development takes up to 2 minutes on my machine, 25% of that time spent setting up watchers on packages I'm unlikely to change (playout-gateway, mos-gateway, server-core-integration, blueprints-integration). I understand the usefulness of having everything rebuild all the time and staying in sync, however in my particular case that usefulness is limited, while the bloat is hindering my workflow.

Adds support of a `--ui-only` flag to the default `yarn dev` script, in which all the packages are built once (in case something changed in a commit by someone else) and then only the type-check watcher and a regular `meteor` are run concurrently.

This gives a 25% increase in build time. I've also added coloring to the various sub-process prefixes, for easier navigation of the console.

obraz

* **Other information**:

**Status**

### DIFF
--- a/scripts/run.js
+++ b/scripts/run.js
@@ -1,4 +1,49 @@
+const process = require('process');
 const concurrently = require("concurrently");
+const args = process.argv.slice(2);
+
+const config = {
+	uiOnly: (args.indexOf('--ui-only') >= 0) || false
+}
+
+function watchPackages() {
+	return [
+		{
+			command: "yarn watch",
+			cwd: "packages",
+			name: "PACKAGES-TSC",
+			prefixColor: 'red',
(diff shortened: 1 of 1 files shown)


Julusian (approved):
Looks good

Interestingly, It feels like its mongo startup that is the slow point on my machine, but that might be because im on an M1 mac..

jstarpl:
> Looks good
> 
> Interestingly, It feels like its mongo startup that is the slow point on my machine, but that might be because im on an M1 mac..

Oh, yeah, mongo startup is _the worst_, but I can't avoid that 😞

PR #670, opened 2022-01-27
feat: First-class Sofie PWA

jstarpl:

Feature

Sofie supports a rudimentary PWA deployment model

The new model uses all relevant, newly available Web APIs (see Chromium's project Fugu: https://fugu-tracker.web.app/) and uses them to create the smoothest possible experience for the Users. Most importantly, this uses Keyboard Lock API and FullScreen API to reduce the amount of situations of lost focus and hotkey conflicts.

* **Other information**:

**Status**


meteor/client/ui/App.tsx
```diff
@@ -168,6 +169,44 @@ export const App = translateWithTracker(() => {
...
+								}
+							)
+						})
+						.catch((e) => console.error('Could not get FullScreen when running as a PWA', e))
+
+					e.preventDefault()
+
+					// Use Keyboard API to lock the keyboard and disable all browser shortcuts
+					if ('keyboard' in navigator) {
+						//@ts-ignore
```
gundelsby, line 197: Add a comment as described here right away, and we won't have to come back to this line when we down the line enable ban-ts-comment: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/ban-ts-comment.md

meteor/lib/typings/webmanifest.d.ts
```diff
@@ -0,0 +1,206 @@
...
+	url: string
+	/**
+	 * The icons member of a shortcut item serves as iconic representations of the shortcut in various contexts.
+	 */
+	icons?: ManifestImageResource[]
+	[k: string]: unknown
+}
+
+export interface ProtocolHandler {
+	protocol:
```
gundelsby, line 182: I'm going to assume we're not gonna be using Sofie UI for bitcoin transactions? :)

jstarpl: Sofie NFT: we're going to automatically mint only 19050607 individual seconds of newscasts, so you can own a slice of reality! And an NFT is forever, so it's potentially worth billions!

jstarpl: In all seriousness, ProtocolHandler interface is just a direct translation of this standard document: https://wicg.github.io/manifest-incubations/#protocol_handlers-member and the related definition of white-listed protocols here: https://html.spec.whatwg.org/multipage/system-state.html#normalize-protocol-handler-parameters, with the addition of the new scheme of `web+some_arbitrary_token`.

meteor/client/ui/App.tsx
jstarpl: Good idea. I've added a description of what we're expecting, why, why it's fine, and where the definiton of the proposed Keyboard API is.

Julusian: I have a feeling our linter rules reject all `ts-ignore`, but do allow `ts-expect-error`, so that once it is no longer an error we know to remove the comment before it starts hiding a new real error

gundelsby (approved):

PR #673, opened 2022-01-31
Fix: log errors

nytamin:
This PR contains various fixes to the error logging.

* fix: use `stringifyError` in log functions. This is to ensure that errors are logged properly (error message, stack etc..).
* Removed unnesessary uses of `.stack` (since this is already handled in stringifyError)
* fix: improve `stringifyError` slightly. Refactor to clarify, and do JSON.stringify as last resort for objects.
* fix: improve client-side uncaughtErrorHandler (adding another stack trace).

### DIFF
--- a/meteor/__mocks__/meteor.ts
+++ b/meteor/__mocks__/meteor.ts
@@ -363,8 +363,7 @@ export async function runInFiber<T>(fcn: () => T | Promise<T>): Promise<T> {
 				}
 			} catch (e: any) {
 				// Note: we cannot use
-				console.log('Error: ' + stringifyError(e))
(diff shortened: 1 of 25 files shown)


jstarpl:
Question: Wouldn't it make more sense as a part of a logger wrapper? It could then output objects where compatible (like DevTools, VS Code, etc.) and output a stringified version of whatever non-string arguments where passed in where this isn't possible (I assume Logstash? Elasticsearch?). It would also simplify this change a lot, since it could be just done in the logger implementation and not have to modify all these files.

nytamin:
Good points @jstarpl ! I'm going to close this PR and instead go in that direction in https://github.com/nrkno/sofie-core/pull/675

PR #674, opened 2022-01-31
feat: sync previous Part Instances

jstarpl:

Feature / bug fix (sort of)

If there is a single Part A, and then the user adLibs PartInstances B and C, B and C can receive all of their timing through a displayTimingGroup. If Current Part Instance is B, an update to the `expectedDuration` of A will not be synced, because at this stage it's a "past" Part Instance and not considered for syncing Part Instances. As a result, it's impossible to change the countdown to the end of B and C.

If the current PartInstance is `orphaned: adlib-part`, the most "previous" Part/PartInstance in the current Segment is found and offered to the Blueprints for syncing.

* **Other information**:

**Status**


packages/blueprints-integration/src/api.ts
```diff
@@ -146,7 +146,7 @@ export interface ShowStyleBlueprintManifest extends BlueprintManifestBase {
 		context: ISyncIngestUpdateToPartInstanceContext,
 		existingPartInstance: BlueprintSyncIngestPartInstance,
 		newData: BlueprintSyncIngestNewData,
-		playoutStatus: 'current' | 'next'
+		playoutStatus: 'previous' | 'current' | 'next'
```
Julusian, line 149: I think it would be good to add a note to the method docs to state when this will be called for 'previous'

meteor/server/api/ingest/syncChangesToPartInstance.ts
```diff
@@ -24,48 +24,66 @@ import {
...
+					cache,
+					ingestCache,
+					playlistPartInstances.currentPartInstance,
+					playlistPartInstances.previousPartInstance,
+					'current'
+				)
+				// If the currentPartInstance is adlibbed we probably also need to find the earliest
+				// non-adlibbed Part within this segment and check it for updates too. It may have something
+				// changed (like timing) that will affect what's going on.
+				if (playlistPartInstances.currentPartInstance.orphaned === 'adlib-part') {
```
Julusian, line 60: I think this needs to be done before the `findPartAndInsertToSyncedInstanceCandidates ` for `currentPartInstance`, because the input to the currentPartInstance will depend on the pieceInstances the previousPartInstance has, so this order will cause some changes to behave weirdly

meteor/server/api/ingest/syncChangesToPartInstance.ts
```diff
@@ -179,3 +197,102 @@ export async function syncChangesToPartInstances(
...
+ * PartInstance that matches that Part. Returns them if found or returns `null` if it can't find anything.
+ */
+function findLastUnorphanedPartInstanceInSegment(
+	cache: CacheForPlayout,
+	currentPartInstance: PartInstance
+): {
+	partInstance: PartInstance
+	part: Part
+} | null {
+	const previousParts = cache.Parts.findFetch(
```
Julusian, line 259: I don't think we need to be looking at the Parts for this.
The strategy I would use is to iterate backwards through the PartInstances (ordered by takeCount, and with a matching playlistActivationId), to find the first one without `orphaned: 'adlib'`.
Then do the same criteria you have below, but if the partinstance doesnt match then return null.

What you have here will find that PartInstance, but if its already been reset, or the part has been deleted in enps, it will find the one before. 
Is that actually desirable? I think this will lead to some confusing behaviour. 
Such as PartA has been modified in enps, but not synced because this is algorithm is finding PartB. PartB gets reset because they set it as next, and now PartA suddenly starts updating.

Julusian (changes requested):
I think the refactoring makes sense here.
Ive added some thoughts about the implementation of `findLastUnorphanedPartInstanceInSegment`, let me know if you want to discuss it

meteor/server/api/ingest/syncChangesToPartInstance.ts
jstarpl: Didn't thought of that. Moved the "previous" Part Instance to be the first in the list.

meteor/server/api/ingest/syncChangesToPartInstance.ts
jstarpl: Okay, how about this way? I guess this will be way more performant because it actually just requires a single find and sort through the PartInstances and then a single `_id` find in Parts.

packages/blueprints-integration/src/api.ts
jstarpl: Added a note on when `playStatus: previous` and which PartInstance is received.

Julusian (approved):

Julusian:
@jstarpl Is there a blueprints side to this, to sync just the duration? It doesnt feel like we should be syncing any piece changes in something that has completed playback

jstarpl:
> @jstarpl Is there a blueprints side to this, to sync just the duration? It doesnt feel like we should be syncing any piece changes in something that has completed playback

Yes, there is, and for some reason I forgot to create a PR for it.

PR #675, opened 2022-02-02
Fix: log errors 2

nytamin:
Replaces https://github.com/nrkno/sofie-core/pull/673

This PR contains various fixes to the error logging.

- [x]  fix: use stringifyError in log functions. This is to ensure that errors are logged properly (error message, stack etc..).
- [x]  fix: improve stringifyError slightly. Refactor to clarify, and do JSON.stringify as last resort for objects.
- [x]  fix: improve client-side uncaughtErrorHandler (adding another stack trace).
- [x]  Removed unnesessary uses of .stack (since this is already handled in stringifyError)
- [x] Fix logging in worker threads (parent & children)
- [ ] Code review


meteor/lib/logging.ts
```diff
@@ -69,6 +59,7 @@ if (Meteor.isServer) {
 } else {
 	const getLogMethod = (type) => {
 		return (...args) => {
+			// TODO: Should we also send errors and warnings to the server?
```
jstarpl, line 62: As a rule, our Front-end code doesn't really log anything anywhere. Any errors that happen, happen either in React, Meteor or are uncaught exceptions, so I don't think there's any point. You shouldn't be using the logger in FE code in the first place.

Julusian:
@nytamin have you checked how logs coming from inside the job-worker behave?
Each worker_thread is using its own instance of winston, with a custom transport to pipe it to its parent. So its possible that some errors will need serializing before being sent over ipc (I might have handled this already, I am not sure)

nytamin:
> @nytamin have you checked how logs coming from inside the job-worker behave?

Not yet, that would be the next thing on my todo-list

jstarpl (approved):

packages/job-worker/src/workers/events/child.ts
```diff
@@ -31,10 +31,11 @@ export class EventsWorkerChild {
 	constructor(
 		emitLockEvent: (event: AnyLockEvent) => Promise<void>,
 		queueJob: QueueJobFunc,
-		logLine: LogLineFunc,
+		logLine: LogLineWithSourceFunc,
 		fastTrackTimeline: FastTrackTimelineFunc | null
 	) {
-		interceptLogging(logLine)
+		// Intercept logging to pipe back over ipc
+		interceptLogging('events-child', logLine)
```
Julusian, line 38: This will lose the context of which 'events-child' thread it was. was it studio0 or studio1?
I assume its the same for the other threads

PR #678, opened 2022-02-08
Feat: ready on air playlists

nytamin:
Blueprint interface change

in getRundown(), the blueprints will be able to call `getPlaylists()` in order to be able to determine which playlist the rundown should be tied to.

This can be used when several rundowns are set to be put in a READY TO AIR-playlist.
Before, the rundowns used a fixed `externalPlaylistId` to join up (which caused the READY TO AIR-playlist to be fixed to a specific _id). This allows the rundowns to make any playlist to be the READY TO AIR-playlist (or convert a non-readytoAir into one, on the fly, while on air)

* **Other information**:

**Status**


packages/blueprints-integration/src/context.ts
```diff
@@ -118,6 +119,13 @@ export interface IShowStyleContext extends ICommonContext, IStudioContext {
 
 export interface IShowStyleUserContext extends IUserNotesContext, IShowStyleContext, IPackageInfoContext {}
 
+export interface IGetRundownContext extends IShowStyleUserContext {
+	/** Returns a list of the Playlists in the studio */
+	getPlaylists: () => Promise<Readonly<IBlueprintRundownPlaylist[]>>
+	getCurrentPlaylist: () => Promise<Readonly<IBlueprintRundownPlaylist> | undefined>
```
Julusian, line 125: It would be good to have some docs on the other two methods here

packages/job-worker/src/ingest/generation.ts
```diff
@@ -565,11 +558,26 @@ export async function getRundownFromIngestData(
...
+		rundownBaselinePackages,
+		async () => {
+			// Note: This can cause a mild race-condition, in the case of two Rundowns being created at the same time.
+			// But we're just ignoreing that for now.
+			return context.directCollections.RundownPlaylists.findFetch({
+				studioId: context.studioId,
+			})
+		},
+		async () => {
+			return context.directCollections.Rundowns.findFetch({
```
Julusian, line 570: It would be good to reduce the number of properties loaded here. The metaData property can be quite heavy, and only the playlistId looks to be needed

Julusian (approved):

PR #682, opened 2022-02-10
fix: ensure timelinediv is an HtmlElement before trying to calculate …

gundelsby:
…its width

Bug fix

We have occurrences in the logs of `Error in Kibana: 'getComputedStyle': parameter 1 is not of type 'Element'` which is believed to originate from this component.

This fix ensures that the timeline div actually is an HTMLElement before trying to calculate its width.

* **Other information**:

**Status**


jstarpl (approved):
`0` seems like the best available alternative, IMHO. I don't think we have a better fallback.

PR #687, opened 2022-02-20
Feat/migration context expose get triggered actionsid

jesperstarkar:
* **What kind of change does this PR introduce?** Feature, adds a new method to `MigrationContextWithTriggeredActions`

Blueprints aren't able to generate a hashed triggeredActionId, which is needed for Migrations i Blueprints to perform any comparisons between existing and potential new actionTriggers

API change exposing `getTriggeredActionsId` in the migration context of Blueprints Integration.

* **Other information**:
It also fixes a bug in `removeTriggeredAction` where the db selector was missing a default value, and all comparisons failed resulting in failed attempts to remove any triggeredActions.

**Status**

### DIFF
--- a/meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
+++ b/meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
@@ -304,8 +304,8 @@ export const TriggeredActionsEditor: React.FC<IProps> = function TriggeredAction
 			.catch(console.error)
 	}, [props.showStyleBaseId])
 
-	const onRemoveTriggeredAction = useCallback((triggeredActionsId: TriggeredActionId) => {
-		MeteorCall.triggeredActions.removeTriggeredActions(triggeredActionsId).catch(console.error)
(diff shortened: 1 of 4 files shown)


jesperstarkar:
I'm not sure if "triggeredActionsId" should become "triggeredActionid". Perhaps subject for another PR?

jstarpl:
> I'm not sure if "triggeredActionsId" should become "triggeredActionid". Perhaps subject for another PR?

This is plural "TriggeredActions" and not "TriggeredAction", because a single Trigger can have multiple Actions assigned to it.

Julusian:
> I'm not sure if "triggeredActionsId" should become "triggeredActionid". Perhaps subject for another PR?

Yeah it would make more sense as `triggeredActionid`. It would be good to do in this pr, as renaming it later on will require blueprints to be updated to match. I generally avoid doing api tidying later on because of that

Julusian (approved):

jesperstarkar:
I'll rename then, if you agree @jstarpl?

jesperstarkar:
Renamed, Blueprints changed too. Briefly tested locally, seems to work.

PR #688, opened 2022-03-08
fix: avoid regenerating segments when only the rank changes

Julusian:
* ** (Bug fix, feature, docs update, ...)

Fix

When receiving a sendROMoveStories or some other mos operations, we will regenerate all the segments following the change

For most of those segments, only the rank has changed, which we have a shortcut for. Avoid re-running blueprints for these segments to speed up the ingest operations

* **Other information**:

**Status**


Julusian:
@nytamin Im not sure if this should target r38 or r39

nytamin (approved):

PR #689, opened 2022-03-11
Fix: bucket items showstyle variants

nytamin:
A fix for the usefulness of bucket items.

The Bucket items are tied to a showStyleVariant, which means that a bucket-item created from one Rundown, might not be playable when moving to the next rundown (with another showStyleVariant).

This PR changes a few things.
1. By default, it generates one Bucket-item for EACH variant
2. Adds the `allVariants` option, so that the blueprints can choose to create just one Bucket-item that is valid for all variants.
3. Adds the `uniquenessId` propertly to Bucket items, so that only one is shown, if there are several ones (like when having different variants)

* **Other information**:

**Status**


packages/job-worker/src/blueprints/postProcess.ts
```diff
@@ -315,21 +316,25 @@ export function postProcessBucketAdLib(
...
-			)
-		),
+		_id: id,
 		externalId,
 		studioId: innerContext.studioIdProtected,
+		showStyleBaseId: innerContext.showStyleCompound._id,
 		showStyleVariantId: innerContext.showStyleCompound.showStyleVariantId,
 		bucketId,
 		importVersions,
+		uniquenessId: uniquenessId ?? unprotectString(id),
```
Julusian, line 337: I dont think this should be forced here,
uniquenessId is something the blueprints can set in their types, so we should respect that value (if they have set it). But I wouldnt be opposed to filling in a default value to match the externalId if it hasnt been set

meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -737,11 +751,34 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
...
 
 				render() {
 					const { connectDragSource, connectDragPreview, connectDropTarget } = this.props
 
 					if (this.props.showStyleBase) {
+						// Hide duplicates;
+						// only the first adLib found with a given uniquenessId will be displayed
+						const uniqueAdlibs = new Map<string, BucketAdLibItem>()
+						for (const adLibPiece of this.state.adLibPieces) {
+							const uniquenessId = adLibPiece.uniquenessId || unprotectString(adLibPiece._id)
```
Julusian, line 769: Is there a reason for using the uniquenessId ere instead of the externalId?

As I see it, in the backend we are using the externalId to figure out if they are the same, then in the frontend we are using the uniquenessId, with no clearly defined relationship between the properties.

To me it would make sense for two passes, first to filter out duplicate externalIds, then to filter out duplicate uniquenessIds.

uniquenessId was originally added so that adlibs from different parts which are identical could be identified and hidden. But this pr 'claims' uniquenessId as a core property for bucket adlibs, even though externalId should be functionally identical

meteor/server/api/buckets.ts
```diff
@@ -169,7 +170,8 @@ export namespace BucketsAPI {
 				externalId: '', // TODO - is this ok?
```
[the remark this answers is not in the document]
nytamin: hm, yeah that was last from a _long_ time ago.. should we just do `externalId: action.externalId,` instead, do you think?

PR #691, opened 2022-03-17
tsconfig issue with SonarCloud

nytamin:
don't omit the file extension when extending a tsconfig, as this results in errors in the SonarCloud scan.


nytamin:
This didn't help :(

PR #694, opened 2022-03-18
Merge 'tv2/develop' into release41

Julusian:
Co-authored-by: Julian Waller <Julusian@users.noreply.github.com>
Co-authored-by: Krzysztof Zegzula <ianshade@users.noreply.github.com>
Co-authored-by: Rasmus Lindved <LindvedKrvang@users.noreply.github.com>

* **Other information**:

**Status**


meteor/lib/mediaObjects.ts
```diff
@@ -416,23 +425,28 @@ export function checkPieceContentStatus(
...
+												frames,
+												type,
+												count: frames,
+											}),
+										})
+									}
 								}
 							}
-							if (deepScan?.blacks) {
-								addFrameWarning(deepScan.blacks, t('black'), t)
```
Julusian, line 432: removing `t(` here will cause translations to not be generated for this string anymore, which is probably not desired?

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -56,16 +62,39 @@ export const RundownTimingProvider = withTracker<
...
+		segments = incomingSegments
 		const partInstances = RundownPlaylistCollectionUtil.getActivePartInstances(props.playlist)
 
 		const currentPartInstance = partInstances.find((p) => p._id === props.playlist?.currentPartInstanceId)
+		const previousPartInstance = partInstances.find((p) => p._id === props.playlist?.previousPartInstanceId)
+
 		currentRundown = currentPartInstance ? rundowns.find((r) => r._id === currentPartInstance.rundownId) : rundowns[0]
+		// These are needed to retrieve the start time of a segment for calculating the remaining budget, in case the first partInstance was removed
+		segmentEntryPartInstances.push(
+			..._.compact([
```
jstarpl, line 84: This can be done without underscore, using `[].filter(Boolean)`.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -121,16 +152,21 @@ export const RundownTimingProvider = withTracker<
...
 		}
+		lowResDurations: RundownTimingContext = {
+			isLowResolution: true,
+		}
 		refreshTimer: number
 		refreshTimerInterval: number
 		refreshDecimator: number
 
 		private timingCalculator: RundownTimingCalculator = new RundownTimingCalculator()
+		private lastSyncedTime: number = 0
```
jstarpl, line 169: Some information what this variable does would be good, like "Last time the `timeupdateSynced` event was sent."

meteor/client/ui/RundownView/RundownViewEventBus.ts
```diff
@@ -59,6 +62,10 @@ export interface ShelfStateEvent extends IEventContext {
 	state: boolean | 'toggle'
 }
 
+export interface MiniShelfQueueAdLibEvent extends IEventContext {
+	forward: boolean
```
jstarpl, line 66: Some explanation here is needed. What does `forward` mean? What if `forward: false`? Does that mean "backward"? Also the other "moving" implementations use a `delta`, I think it would be nice to repeat that pattern here. Then it would be clear what `delta: 1` and `delta: -1` mean.

meteor/client/ui/SegmentTimeline/Parts/OutputGroup.tsx
```diff
@@ -42,6 +42,7 @@ interface IOutputGroupProps {
 	onContextMenu?: (contextMenuContext: IContextMenuContext) => void
 	indexOffset: number
 	isPreview: boolean
+	showDurationSourceLayers?: Set<string>
```
jstarpl, line 45: What does `showDurationSourceLayers` do? Why is it a `Set<string>`?

meteor/client/ui/SegmentTimeline/Renderers/CustomLayerItemRenderer.tsx
```diff
@@ -182,6 +200,23 @@ export class CustomLayerItemRenderer<
 		) : null
 	}
 
+	renderDuration() {
+		const uiPiece = this.props.piece
+		const innerPiece = uiPiece.instance.piece
+		const content = innerPiece.content
+		const duration = content && content.sourceDuration
+		if (duration && this.props.showDurationSourceLayers?.has(innerPiece.sourceLayerId)) {
```
jstarpl, line 208: Why pass an entire set down to the `CustomLayerItemRender`? Shouldn't just the renderer be informed that it either should or shouldn't render the duration? It has no reason to know what other sourceLayers should have this enabled. This should just be a simple boolean at this stage.

meteor/client/ui/SegmentTimeline/Renderers/DefaultLayerItemRenderer.tsx
```diff
@@ -52,7 +53,16 @@ export class DefaultLayerItemRenderer extends CustomLayerItemRenderer<IProps, IS
 							ref={this.setLeftLabelRef}
 							style={this.getItemLabelOffsetLeft()}
 						>
-							<span className="segment-timeline__piece__label">{this.props.piece.instance.piece.name}</span>
+							<span
+								className={ClassNames(
+									'segment-timeline__piece__label',
+									'with-duration',
```
jstarpl, line 59: Why does this add the `with-duration` class, even though `this.renderDuration()` can choose not to render the duration at all?

meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx
```diff
@@ -303,70 +307,50 @@ export class VTSourceRendererBase extends CustomLayerItemRenderer<IProps & WithT
...
-
 			const piece = this.props.piece
 			if (piece.contentPackageInfos) {
 				// TODO: support multiple packages:
 				const contentPackageInfos = Object.values(piece.contentPackageInfos)
 				if (contentPackageInfos[0]?.deepScan?.scenes) {
-					const scenes = _.compact(
-						contentPackageInfos[0].deepScan.scenes.map((i) => {
-							const sceneTime = i * 1000
-							if (sceneTime < itemDuration) {
```
jstarpl, line 316: Doesn't this capping of `sceneTime` at `itemDuration` serve some purpose? I'm almost sure this filter here was a result of a bug somewhere.

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
jstarpl: Also, this feels like it should be optimized, no reason to scan the PartInstances collection this many times, especially since on most takes the currentPartInstance will be set.

meteor/client/lib/triggers/TriggersHandler.tsx
```diff
@@ -163,14 +177,24 @@ export interface MountedGenericTrigger {
...
+	/** Hint that all actions of this trigger are adLibs */
+	adLibOnly: boolean
 }
 
 export const MountedGenericTriggers = new Mongo.Collection<MountedGenericTrigger>(null)
 
+export function isMountedAdLibTrigger(
+	mountedTrigger: MountedAdLibTrigger | MountedGenericTrigger
+): mountedTrigger is MountedAdLibTrigger {
+	return !!mountedTrigger['targetId']
```
jstarpl, line 195: `MountedAdLibTrigger`  and `MountedGenericTrigger` are not extensions of a shared base, they come from two different collections. It's not possible to mix them up, at least not here - why is this function here then? The check of `!!mountedTrigger['tragetId']` can't actually guarantee that this is an `MountedAdLibTrigger`; if the topology of `MountedAdLibTrigger` changes, this will not show up in any compilation message.

meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx
nytamin: I had doubts on this last week as well.
This is actually a bug-fix; IIRC @ianshade explained that this is fixing a situation where scenes/blacks/freezes was filtered for only the original duration of a Piece (like when it was capped in a Part). If the used decided to play the Part longer than planned, the scenes/blacks/freezes wouldn't show up for the lengthened part of the Piece.

The scenes are [ultimately filtered later](https://github.com/nrkno/sofie-core/blob/sync/release41/meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx#L546), so this should be OK.

With this being said, I've got this on my list of Things-to-test, to ensure that it still works properly.

jstarpl: That's a good find, I totally forgot that they are filtered later! Yeah, if they are filtered later to make sure we don't produce more DOM elements than we need, then this definitely makes sense.

meteor/client/ui/SegmentTimeline/Parts/OutputGroup.tsx
thomasslee97: It's a set of the sourcelayers for which the duration of the piece should be shown (e.g. only VTs and VOs should have timing shown on the Piece. I don't know why it's a set instead of an array.

thomasslee97: Documentation is on the `RundownViewLayout` interface https://github.com/nrkno/sofie-core/blob/0633774ad3c69186194a9bc7dd5d2f062143e6f9/meteor/lib/collections/RundownLayouts.ts#L367

meteor/client/lib/triggers/TriggersHandler.tsx
thomasslee97: This function also appears to be unused

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
ianshade: `filter(Boolean)` doesn't narrow the type down. How about if I added this to lib/lib?
``` ts
export function compact<T>(array: (T | false | undefined | null | '' | 0)[]): T[] {
	return array.filter(Boolean) as T[]
}
```
As for the second comment, SegmentInstances would solve it. Otherwise the Ids of the entry partInstances would need to be stored somewhere else, perhaps on the playlist

meteor/client/lib/triggers/TriggersHandler.tsx
ianshade: It is used by TV 2 in the code excluded from contribution, where we need to tell apart `MountedAdLibTrigger`s from `MountedGenericTrigger`s. 
you do mix them together sometimes, for example in `mountedTriggerToHotkeyList`.
Should I make them share a base class then?

meteor/client/ui/SegmentTimeline/Parts/OutputGroup.tsx
jstarpl: Yeah, but why Set of `string` as opposed to Set of `ISourceLayer['_id']`? String is very opaque.

meteor/client/lib/triggers/TriggersHandler.tsx
jstarpl: @ianshade Okay, so let's put this into a lib file or just put it next to where they are combined. Alternatively, let's refactor this so that they have a shared base and a clear discriminator for one or the other.

jstarpl: What I'm specifically getting at, is that the triggers ecosystem was specifically designed to avoid using type predicates and allow the interface typings to implicitly state the types of various objects and available properties. Type predication is a huge problem with the whole `IAdLib*Ui` environment that I would like to clean up at some point: the predicates aren't very good (as they generally aren't type-safe themselves) and we use them more as a tool to make the compiler happy rather than providing actual type safety.

meteor/client/ui/SegmentTimeline/Parts/OutputGroup.tsx
ianshade: Using `ISourceLayer['_id']` doesn't seem to do anything in terms of type checking. It still resolves to `Set<string>` and I still can do silly things like `showDurationSourceLayers.has(sourceLayer.name)`, but you want it to just be a hint to the developers?

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -121,16 +159,22 @@ export const RundownTimingProvider = withTracker<
 	{
 		static childContextTypes = {
 			durations: PropTypes.object.isRequired,
+			lowResDurations: PropTypes.object.isRequired,
```
jstarpl, line 162: What is the point of this split between `durations` and `lowResDurations`?

meteor/client/ui/RundownView/RundownTiming/withTiming.tsx
```diff
@@ -102,12 +119,57 @@ export function withTiming<IProps, IState>(
...
-				if (!!this.filterGetter && durations.isLowResolution !== !expandedOptions.isHighResolution) {
+				if (componentIsDirty(this.filterGetter, highResDurations, expandedOptions.dataResolution)) {
 					this.isDirty = true
 				}
 
-				return <WrappedComponent {...this.props} timingDurations={durations} />
+				return (
+					<WrappedComponent
+						{...this.props}
+						timingDurations={rundownTimingDataFromDataResolution(expandedOptions.dataResolution, {
```
jstarpl, line 129: This will create an object on every update and then Garbage-Collect the old inline object. The point of doing `timingDurations={durations}` was to remove that garbage being created so that there's nothing to GC (because we're just passing references).

meteor/client/ui/RundownView.tsx
```diff
@@ -1996,81 +2106,149 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
...
-					countdownToSegmentRequireLayers={this.state.rundownViewLayout?.countdownToSegmentRequireLayers}
-					fixedSegmentDuration={this.state.rundownViewLayout?.fixedSegmentDuration}
-				/>
-			)
+			if (adlibPiece.invalid) {
+				NotificationCenter.push(
+					new Notification(
+						t('Invalid AdLib'),
+						NoticeLevel.WARNING,
+						t('Cannot play this AdLib because it is marked as Invalid'),
```
jstarpl, line 2117: This code is duplicated in `RundownViewShelf.tsx` in this same PR, apart from alrady being present in some other components (namely `AdLibPanel.tsx`, `AdLibRegionPanel.tsx` and others). I think this should be refactored to limit the amount of copies of these code blocks.

meteor/client/ui/RundownView.tsx
```diff
@@ -1996,81 +2106,149 @@ export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((
...
+				}
+			}
+
+			if (pieceToQueue) {
+				this.queueAdLibPiece(pieceToQueue, e)
+				this.setState({ keyboardQueuedPiece: pieceToQueue })
+			}
+		}
+
+		queueNextMinishelfAdLib = (e: any) => {
```
jstarpl, line 2246: This doesn't seem to be used anywhere? That would also mean all the adjacent methods here are unused and not needed?

meteor/client/ui/SegmentContainer/withResolvedSegment.ts
```diff
@@ -84,10 +85,14 @@ export interface IProps {
 	isLastSegment: boolean
 	ownCurrentPartInstance: PartInstance | undefined
 	ownNextPartInstance: PartInstance | undefined
+	adLibSegmentUi?: AdlibSegmentUi
 	isFollowingOnAirSegment: boolean
 	rundownViewLayout: RundownViewLayout | undefined
 	countdownToSegmentRequireLayers: string[] | undefined
 	fixedSegmentDuration: boolean | undefined
+	minishelfRegisterHotkeys?: boolean
```
jstarpl, line 93: Does this do anything now? I don't see this property being used anywhere.

meteor/client/ui/RundownView/RundownViewShelf.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+interface IRundownViewShelfProps {
+	studio: Studio
+	segment: SegmentUi
+	playlist: RundownPlaylist
+	showStyleBase: ShowStyleBase
+	adLibSegmentUi: AdlibSegmentUi
+	hotkeyGroup: string
+	studioMode: boolean
+	registerHotkeys?: boolean
```
jstarpl, line 37: This doesn't seem to be used anywhere or do anything.

meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx
```diff
@@ -484,11 +405,17 @@ export class VTSourceRendererBase extends CustomLayerItemRenderer<IProps & WithT
 			<span className="segment-timeline__piece__label" ref={this.setLeftLabelRef} style={this.getItemLabelOffsetLeft()}>
 				{noticeLevel !== null && <PieceStatusIcon noticeLevel={noticeLevel} />}
 				<span
-					className={ClassNames('segment-timeline__piece__label', {
-						'overflow-label': end !== '',
-					})}
+					className={ClassNames(
+						'segment-timeline__piece__label',
+						'with-duration',
```
jstarpl, line 410: Same problem as in https://github.com/nrkno/sofie-core/blob/05a90536a680bc75ca10ad950a53f2ae2ce9ac01/meteor/client/ui/SegmentTimeline/Renderers/DefaultLayerItemRenderer.tsx#L59 - the `with-duration` class is added, even though `this.renderDuration()` makes an independent decision if it will render the duration or not.

meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx
```diff
@@ -561,10 +488,18 @@ export class VTSourceRendererBase extends CustomLayerItemRenderer<IProps & WithT
...
 		) {
-			const endOfContentAt = (this.props.piece.renderedInPoint || 0) + (vtContent.sourceDuration - seek)
+			const lastFreeze = this.state.freezes && this.state.freezes[this.state.freezes.length - 1]
+			const endingFreezeStart =
+				lastFreeze &&
+				lastFreeze.start >= vtContent.sourceDuration &&
+				lastFreeze.start < vtContent.sourceDuration + (vtContent.postrollDuration || 0) &&
+				lastFreeze.start
+			const endOfContentAt =
+				(this.props.piece.renderedInPoint || 0) +
```
jstarpl, line 501: This is a cool improvement, though some explanation would be great here, and maybe it should be optional somehow? Whether the countdown to freeze-frame should count down to actual freeze-frames or just the end of the clip is I think editorially-dependent (say, in an obit clip ending with a black-and-white photo of the deceased).

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -528,41 +549,48 @@ export const SegmentTimelineContainer = withResolvedSegment(
...
+							: partOffset + lastTakeOffset
 
-				this.setState({
-					livePosition: newLivePosition,
-					scrollLeft: this.state.followLiveLine
-						? Math.max(newLivePosition - LIVELINE_HISTORY_SIZE / this.state.timeScale, 0)
-						: this.state.scrollLeft,
-				})
-			}
+					const budgetDuration = this.getSegmentBudgetDuration()
```
jstarpl, line 582: Does it really make sense to calculate this budgetDuration on every tick? `getSegmentBudgetDuration` seems to only rely on data from the Database, there doesn't seem to be any dependency on current time at all. Couldn't it just be calculated when the reactive data changes and then reused?

meteor/client/ui/SegmentTimeline/SmallParts/SegmentTimelineSmallPartFlag.tsx
```diff
@@ -114,6 +116,7 @@ export const SegmentTimelineSmallPartFlag = ({
 				isLastInSegment={isLastInSegment}
 				totalSegmentDuration={partDurations}
 				parentTimeScale={timeScale}
+				showDurationSourceLayers={showDurationSourceLayers}
```
jstarpl, line 119: I think we should make a point to refactor `showDurationSourceLayers` _et consortes_ into some sort of a React context. I hate how this needs to be passed around like that. Not something to be addressed right now, but soon-ish.

meteor/client/ui/SegmentTimeline/SourceLayerItem.tsx
```diff
@@ -81,8 +121,12 @@ export const SourceLayerItem = withTranslation()(
 					top: 0,
 					left: 0,
 				},
-				scrollLeftOffset: 0,
+				cursorRawPosition: {
+					clientX: 0,
+					clientY: 0,
+				},
 				cursorTimePosition: 0,
+				elementWidth: 0,
```
jstarpl, line 129: this.state.elementWidth is not used anymore.

meteor/client/ui/SegmentTimeline/SourceLayerItem.tsx
```diff
@@ -356,6 +411,58 @@ export const SourceLayerItem = withTranslation()(
...
+		// 		const width = getElementWidth(this.state.itemElement) || 0
+		// 		if (this.state.elementWidth !== width) {
+		// 			this.setState({
+		// 				elementWidth: width
+		// 			})
+		// 		}
+		// 	}
+		// }
+
+		private measureElement = () => {
```
jstarpl, line 427: I don't think this is needed anymore, since `this.state.elementWidth` is no longer used.

meteor/client/ui/SegmentTimeline/SourceLayerItem.tsx
```diff
@@ -356,6 +411,58 @@ export const SourceLayerItem = withTranslation()(
...
+						elementWidth: width,
+					})
+				}
+			}
+		}
+
+		private mountResizeObserver() {
+			if (this.props.isLiveLine && !this._resizeObserver && this.itemElement) {
+				this._resizeObserver = new ResizeObserver(this.onResize)
+				this._resizeObserver.observe(this.itemElement)
```
jstarpl, line 453: The `ResizeObserver` was refactored away because it added unnecessary overhead. We settled on using a different way of calculating this information - see https://github.com/nrkno/sofie-core/blob/a6f19b4d0644b5572280cdffc6cbb8f168b57d57/meteor/client/ui/SegmentTimeline/SourceLayerItem.tsx#L744

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
```diff
@@ -1037,6 +1052,119 @@ const HotkeyLegendSettings = withTranslation()(
 			})
 		}
 
+		onDownloadAHKScript = () => {
```
jstarpl, line 1055: Does it make sense to include this in the public release, considering it uses a hard-coded, private TV2 AHK mapping?

meteor/lib/tv2/AHKkeyboardMap.ts
```diff
@@ -0,0 +1,75 @@
+import * as _ from 'underscore'
```
jstarpl, line 1: Should this file even be included in the repo?

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
```diff
@@ -1049,6 +1177,10 @@ const HotkeyLegendSettings = withTranslation()(
 						>
 							<th className="settings-studio-custom-config-table__name c2">{hotkeyHelper.shortcutLabel(item.key)}</th>
 							<td className="settings-studio-custom-config-table__value c3">{item.label}</td>
+							<td className="settings-studio-custom-config-table__value c2">{item.platformKey || ''}</td>
```
jstarpl, line 1180: Are we importing this hostKey/platformKey feature? If so, this needs to be well-documented.

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
```diff
@@ -1091,6 +1223,47 @@ const HotkeyLegendSettings = withTranslation()(
...
+												className="input text-input input-l dropdown"
+												mutateUpdateValue={(v) => (v ? v : undefined)}
+											/>
+										</div>
+										<div className="mod mvs mhs">
+											<label className="field">
+												{t('Key color')}
+												<EditAttribute
+													modifiedClassName="bghl"
+													attribute={'hotkeyLegend.' + index + '.buttonColor'}
```
jstarpl, line 1258: I don't think this is used anywhere outside of the private TV2 Keyboard map feature, that we're apparently not importing. If it's not being used, it should be removed.

meteor/client/ui/Settings/components/triggeredActions/actionEditors/actionSelector/ActionSelector.tsx
```diff
@@ -162,6 +167,8 @@ function actionToLabel(t: TFunction, action: SomeAction['action']): string {
 			return t('Go to On Air line')
 		case ClientActions.showEntireCurrentSegment:
 			return t('Show entire On Air Segment')
+		case ClientActions.miniShelfQueueAdLib:
+			return t('MiniShelf queue AdLib')
```
jstarpl, line 171: I think we had a disussion about _queue_ vs _cue_ and _cue_ won. I actually wonder if this should be renamed "select", since _cue_ suggests a playout operation being performed, while this seems to be an entirely client-side operation?

packages/blueprints-integration/src/rundown.ts
```diff
@@ -519,6 +530,10 @@ export interface IBlueprintAdLibPiece<TMetadata = unknown> extends IBlueprintPie
 	nextPieceTags?: string[]
 	/** String that can be used to identify adlibs that are equivalent to each other */
 	uniquenessId?: string
+	/** When not playing, display in the UI as playing, and vice versa. Useful for Adlibs that toggle something off when taken */
+	invertOnAirState?: boolean
+	/** Do not assign a hotkey to this adlib */
+	noHotKey?: boolean
```
jstarpl, line 536: This doesn't seem to do anything.

packages/blueprints-integration/src/action.ts
```diff
@@ -34,6 +34,10 @@ export interface IBlueprintActionManifestDisplay {
 	nextPieceTags?: string[]
 	/** String that can be used to identify adlibs that are equivalent to each other */
 	uniquenessId?: string
+	/** When not playing, display in the UI as playing, and vice versa. Useful for Adlibs that toggle something off when taken */
+	invertOnAirState?: boolean
+	/** Do not assign a hotkey to this adlib */
+	noHotKey?: boolean
```
jstarpl, line 40: I don't think this actually does anything. I don't see anything reading from `noHotKey`.

meteor/client/ui/SegmentContainer/withResolvedSegment.ts
```diff
@@ -331,17 +353,7 @@ export function withResolvedSegment<T extends IProps, IState = {}>(
...
-						(i) =>
-							i.instance._id === props.playlist.currentPartInstanceId ||
-							i.instance._id === nextProps.playlist.currentPartInstanceId
-					) ||
-						data.parts.find(
-							(i) =>
-								i.instance._id === props.playlist.nextPartInstanceId ||
-								i.instance._id === nextProps.playlist.nextPartInstanceId
-						))) ||
+					((data.parts && findNextOrCurrentPart(data.parts)) || data.segmentui?.showShelf)) ||
```
ianshade, line 356: `data.segmentui?.showShelf` doesn't seem right here. I don't remember the reason it was added, but it causes segments with Mini Shelves to update way too often.

meteor/client/ui/Settings/components/triggeredActions/actionEditors/actionSelector/ActionSelector.tsx
ianshade: It is consistent with `isQueueable`, `innerStartOrQueueAdLibPiece` and so on. It is not just a client-side operation. When triggered, it executes the START_ADLIB userAction and may result in queueing a Part.

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
ianshade: Is any part of the hotkey legend something you can use at all? If not, I think maybe we should exclude the entire HotkeyLegendSettings

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
ianshade: I wouldn't say the AHK mapping is hardcoded. It'll remap almost any combo for you. There seems to be only one character specific to Danish layout, which should go away.

meteor/client/ui/SegmentTimeline/SmallParts/SegmentTimelineSmallPartFlag.tsx
ianshade: Let's use some central state management. MobX perhaps? 😃

meteor/client/ui/RundownView/RundownViewShelf.tsx
ianshade: Leftovers from pre-actionTriggers times. 
(But that also means we seem to have a lost feature here, because it was supposed to show the assigned shortcut on the button)

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
ianshade: A bit of a naming confusion. `lowResDurations` are for all the components that update in a synced fashion, so that when they re-render for non-timing related reasons, they still hold their time throughout the entire second. I'll fix that.

meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx
ianshade: @jstarpl So the idea is that this class does not decide if the duration should be shown or not and it does not break the label if the duration is not shown later on. It just means that there might be a duration displayed in that label

meteor/client/ui/SegmentTimeline/Renderers/VTSourceRenderer.tsx
nytamin: Agreed, this should be an opt-in.
We're disabling this one for now, to be re-enabled by tv2 later and put behind a settings-flag, or blueprints property or something.

meteor/client/ui/SegmentTimeline/SmallParts/SegmentTimelineSmallPartFlag.tsx
nytamin: To be continued then.. :)

meteor/client/ui/SegmentTimeline/SourceLayerItem.tsx
ianshade: @jstarpl I fixed this, but I feel like this function gets called too often now, please give another look

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
nytamin: In the interest of keeping the code-base as unified as possible, we'll allow keeping this. TV2 promises to keep the number of hard-coded things to a minimum :)

meteor/lib/tv2/AHKkeyboardMap.ts
nytamin: We'll allow it for now. Ref: https://github.com/nrkno/sofie-core/pull/694#discussion_r841429702

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
nytamin: We're keeping this in the spirit of keeping our codebases aligned as much as possible and even though NRK won't use most parts of it it won't hurt us to keep it.
We've put it behind a feature-flag for now, to reduce noise to the user in GUI.

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
nytamin: ref: https://github.com/nrkno/sofie-core/pull/694#discussion_r841458278

meteor/client/ui/Settings/components/triggeredActions/actionEditors/actionSelector/ActionSelector.tsx
nytamin: We're pretty inconsistent atm with queue vs cue anyway, so we'll leave this for now.

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
nytamin: @ianshade will look into this in a bit

meteor/client/ui/Settings/components/triggeredActions/actionEditors/actionSelector/ActionSelector.tsx
jstarpl: If this executes `START_ADLIB` then this needs to be documented somewhere. I wouldn't expect a "queue" action to "start" something.

PR #695, opened 2022-03-18
doc: package manager powershell note

imaretic:
Docs update

Some Powershell users won't be able to start package-manager successfully - they need to add an extra pair of dashes.

The note is added so that all Powershell users who encounter the issue will know how to fix it.


jstarpl (approved):
Good note! I wonder why this is a problem...

Julusian:
@jstarpl I was curious and it looks like powershell is using the first `--` for its own command line parsing, which cmd does not do.
cmd does appear to be happy with having the extra `--` in the command, so this could be simplified down to always use the three `--` version. But that could of course break in the future as Im not sure at what point that `--` is being parsed/ignored so it may suddenly have meaning

PR #697, opened 2022-03-25
chore: replace standard version with custom script

baltedewit:

Replacing standard version with a custom script. Move the changelog up to the root of the repository. This reduces the number of dependencies and allows us to create more tailored workflows in the future.

**Status**


scripts/updateVersion.mjs
```diff
@@ -0,0 +1,128 @@
...
+	writeFile(
+		"./meteor/package.json",
+		JSON.stringify(packageFile, undefined, "\t")
+	);
+
+	// update the library files
+	await execPromise("cd ./packages && yarn sync-version-and-changelog");
+
+	// git commit
+	await execPromise(`git add *`);
```
Julusian, line 120: It would be good to do a very specific git add. I have a bunch of local files that should not be included which this will probably pick up (bash scripts to run various components with predefined environment variables)

scripts/updateVersion.mjs
```diff
@@ -0,0 +1,128 @@
+import semver from "semver";
+import { exec } from "child_process";
+import { readFile, writeFile } from "fs/promises";
+
+const REPO_URL = "https://github.com/nrkno/sofie-core";
+const START_OF_LAST_RELEASE_PATTERN =
+	/(^#+ \[?[0-9]+\.[0-9]+\.[0-9]+|<a name=)/m;
+const HEADER = `# Changelog\n\nAll notable changes to this project will be documented in this file. See [Convential Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) for commit guidelines.\n\n`;
+
+const execPromise = (command) =>
```
Julusian, line 10: I would recommend https://www.npmjs.com/package/zx to simplify running of stuff like this
I know Jan raised some concern about windows compatability when I suggested it before, but I havent yet had any complaints from companion after I rewrote a bunch of scripts to zx.

scripts/updateVersion.mjs
```diff
@@ -0,0 +1,128 @@
+import semver from "semver";
+import { exec } from "child_process";
+import { readFile, writeFile } from "fs/promises";
+
+const REPO_URL = "https://github.com/nrkno/sofie-core";
```
Julusian, line 5: Can we pull this from one of the package.json files?

scripts/updateVersion.mjs
```diff
@@ -0,0 +1,128 @@
...
+const oldContentStart = oldContent.search(START_OF_LAST_RELEASE_PATTERN);
+if (oldContentStart !== -1) {
+	oldContent = oldContent.substring(oldContentStart);
+}
+
+if (!isDryRun) {
+	await writeFile("./CHANGELOG.md", HEADER + md + oldContent);
+
+	// update the package.json
+	packageFile.version = nextVersion;
```
Julusian, line 110: could use a `yarn version 1.3.2` to set the version instead of manually writing the file? makes sure we wont mess with the formatting

scripts/updateVersion.mjs
```diff
@@ -0,0 +1,128 @@
...
+	for (const [type, commits] of Object.entries(changes)) {
+		if (!groups[type]) continue;
+		md += "\n\n### " + groups[type] + "\n";
+
+		for (const change of commits) {
+			md += `\n* ${change.scope ? `**${change.scope}** ` : ""}${
+				change.description
+			} [${
+				change.short
+			}](${REPO_URL}/nrkno/sofie-core/commit/${change.hash.trim()})`;
```
Julusian, line 94: this is doubling the `/nrkno/sofie-core` portion of the url

Julusian:
Ive only skimmed it while curious for how things had been approached, but looks like a good start :)

scripts/updateVersion.mjs
baltedewit: I've not used zx for this because it requires a minimum node version of 16

scripts/updateVersion.mjs
baltedewit: The repository field is only available in to the packages, not in the root or meteor package.json. But I think we should add it.

baltedewit:
We intend to overhaul the changelogs in a different way than this PR, to be continued...

PR #698, opened 2022-03-25
feat: reduce number of loaded PartInstances

Julusian:

feature/fix

Rebase of https://github.com/nrkno/sofie-core/pull/696 for r41

Over time, the number of PartInstances loaded into the Cache increases. It only gets reduced when parts/segments are reset, which does not happen automatically (by design)

To avoid the amount of data continually growing, we can limit the PartInstances to be the Segments containing the previous, current and next PartInstances. It might be possible to reduce this further, but this much is a good starting point.

* **Other information**:

**Status**


jstarpl (approved):
Looks good to me!

PR #702, opened 2022-03-30
fix: missing/broken device debug logging in playout-gateway

Julusian:

fix

enabling 'debug' logging in playout gateway doesnt produce any logging from any device

debug logging is now logged

* **Other information**:

**Status**

### DIFF
--- a/packages/playout-gateway/src/coreHandler.ts
+++ b/packages/playout-gateway/src/coreHandler.ts
@@ -219,6 +219,10 @@ export class CoreHandler {
 			if (logLevel !== this.logger.level) {
 				this.logger.level = logLevel
 
+				for (const transport of this.logger.transports) {
+					transport.level = logLevel
+				}
+
 				this.logger.info('Loglevel: ' + this.logger.level)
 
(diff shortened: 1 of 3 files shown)


Julusian:
@nytamin It is very possible I have misinterpreted the relationship between `device.debugLogging` and `this._coreHandler.logDebug` here. But as it was I was getting nothing logged for any value of `device.debugLogging`.

Additionally there is a formatting change for the debug logging here as it was always producing `[object Object]`. This might want to go to r41, as I suspect this was a consequence of the winston update?

nytamin (changes requested):
Yeah, I believe that this PR is a result of both a misinterpretation and a bug.

The bug was that the default logLevel was set to verbose, which excluded debug-log-messages from being logged to console. I pushed a fix to this bug just now: https://github.com/nrkno/sofie-core/commit/304241b3ffe7b0186500fb91e74ca2581c6e0883

I'm happy with the formatting-change though, so I suggest that we keep the original if-clause, but use your formatting. (Just make sure you log to `this.logger.debug` and not `this.logger.info`

PR #703, opened 2022-03-30
fix: update supertimeline

Julusian:

fix

We are stuck on an old version of supertimeline, as we rely on some bugs in the behaviour of how groups overlap

This behaviour has been fixed in supertimeline, and the way we convert pieces to the timeline has been updated to not rely on broken overlap behaviour.
Additionally, the preroll of pieces is really clear in the new structure.

* **Other information**:

The change is relatively simple.
Previously, we had a piece_group, with a first_object inside.
The piece_group is the group that sits on the source_layer and houses all the child objects. The first_object defined the started/stopped callbacks sent to core. This meant all timing was done on the piece_group, and so it was trying to overlap with the previous piece when preroll was used. And as it was on a layer, it is not allowed to overlap.

Now, we have a control_object and a piece_group.
The base timing (without preroll) is on the control_object. This object also defines the started/stopped callback. This object is on the defined source_layer, and so produces the timing on when the piece is visible.
The piece_group, is now timed relative to the control_object, offsetting for preroll (and in the future postroll). As it is not on a layer, it is allowed to overlap.

Before merging, I need to remove the resolution forcing the version of supertimeline and update the version tsr is using directly.

**Status**


baltedewit (approved):
Good change, makes the timing of the pieces regarding preroll / postroll a bit simpler too

PR #704, opened 2022-04-04
[r43] Update Meteor to 2.7 and update MongoDB-related typings

Julusian:

* **Other information**:

A basic test of playout looks to be happy, and the ui loads and looks as I expect it to.

As part of this I have completely isolated our MongoDb typings from those in @types/meteor. Initially I tried to make it work, but it is not possible to override a type in the library, so we would have to change userId back to a string instead of a ProtectedString. 
But this is fine, as we already had a very simple class wrapping their mongo collection to be able to improve the errors it generated, so it as at that point that we isolate the typings. Unfortonately there are some casts to any in there, as the typings for mongo queries are fundamentally incompatible, or would get typescript stuck in infinite recursion. The scope for future error is very contained to a small location, in an api that is unlikely to undergo any major changes without it being clearly documented, so I think there is minimal risk of issues being introduced here. 

**Status**


meteor/lib/lib.ts
```diff
@@ -487,3 +489,7 @@ export enum LogLevel {
 	ERROR = 'error',
 	NONE = 'crit',
 }
+
+export function convertCorelibToMeteorMongoQuery<T>(query: CoreLibMongoQuery<T>): MongoQuery<T> {
```
nytamin, line 493: This function could use a short doc, describing its purpose

meteor/lib/collections/lib.ts
```diff
@@ -81,17 +76,19 @@ export function ObserveChangesForHash<DBInterface extends { _id: ProtectedString
 	}
 }
 
+/**
+ * Create a fully featured MongoCollection
+ * @param name Name of the collection in mongodb
+ * @param options Open options
+ */
```
nytamin, line 83: No issues here, just a thanks and a 👍  for taking the time to improve the docs as you go! :)

nytamin (approved):
I've given it a read, it looks good to me!

Due to the size of this PR, I'd like us to wait with merging it until **Release43**, since it won't affect functionality anyway.
I'd like the TV2-devs to give it a quick read-through as well, I'll ping them.

Julusian:
@nytamin actually, this doesn't affect the mongodb typings that much. Our `MongoCollection` now has no relation to Meteor's `Mongo.Collection`, but `MongoQuery` is virtually unchanged. Previously we were using a copy of their typings (not sure why), but we are now simply using theirs, which as far as I can tell is identical.  
Most of the files changed are because of imports that moved, as I didnt add re-exports for all of them when I moved types.

PR #708, opened 2022-04-07
feat: use Sorensen prevent default down option and configurable poison key

KvelaGorrrrnio:
Bug fix

Shortcuts that are bound on keyUp and conflicts with browser shortcuts doesn't prevent default behaviour.

Shortcuts prevent default behaviour on keyDown.

* **Other information**:
The code that prevents default behaviour for `Ctrl+Digit` and function keys has not been removed, since they might be preferred prevented, even when no action triggers have been defined for that combo.

**Status**


meteor/client/lib/triggers/TriggersHandler.tsx
```diff
@@ -304,11 +306,15 @@ export const TriggersHandler: React.FC<IProps> = function TriggersHandler(
 		]
 		const systemActionKeys = ['Enter', 'NumpadEnter', 'Tab']
 
+		const poisonKey: string | null = Meteor.settings.poisonKey
```
jstarpl, line 309: In the way this is currently set up, this needs to import `Settings` from `meteor\lib\Settings.ts` and use `Settings.poisonKey` as that's where the "defaulted" version of settings lives.

jstarpl (changes requested):
Looks good, apart from the default `poisonKey` which wouldn't work, as the code is now.

meteor/client/lib/triggers/TriggersHandler.tsx
KvelaGorrrrnio: @jstarpl Thanks for the catch. It is fixed now.

jstarpl (approved):
Looks good now!

PR #709, opened 2022-04-07
fix: auto update lobby errors

imaretic:

Bugfix.

On the Lobby page, the system status is fetched only when the page loads.
More info here: https://app.asana.com/0/1201969011790422/1200582151644811/f

System status will update every 5 seconds with the latest logs.

* **Other information**:

**Status**


Julusian:
I know we don't have any examples of this (yet), but instead of polling it might be nice to put the messages into a custom publication, so that they get pushed to the client?
It will be a lot more involved in the backend, as the data source doesnt look written appropriately for a custom publication currently (cc @nytamin)

PR #710, opened 2022-04-07
feat(Prompter): Support for multiple Joycon controller for prompter

jesperstarkar:
New feature/enhanced capatibility

Only a singe L or R or LR(pair) joycon(s) can be connected.
Known problems when going from LR mode to either L or R, where input might get stuck until windows restart.

Multiple LL or RR controllers supported. Less/no problems when going LR to L or R. 

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Prompter/controller/joycon-device.ts
+++ b/meteor/client/ui/Prompter/controller/joycon-device.ts
@@ -1,8 +1,10 @@
 import { ControllerAbstract } from './lib'
 import { PrompterConfigMode, PrompterViewInner } from '../PrompterView'
 import Spline from 'cubic-spline'
+import { isUndefined } from 'underscore'
 
-type JoyconMode = 'L' | 'R' | 'LR'
+type JoyconWithData = {index: number, mode: JoyconMode, axes: readonly number[], buttons: number[]}
+type JoyconMode = 'L' | 'R' | 'LR' | null
 
 /**
  * This class handles control of the prompter using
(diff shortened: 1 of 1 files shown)


jstarpl (changes requested):
Overall, seems good, one line seems like it can be removed, one question about typings.

meteor/client/ui/Prompter/controller/joycon-device.ts
```diff
@@ -208,75 +201,94 @@ export class JoyConController extends ControllerAbstract {
...
-		if (this.lastUsedJoyconMode === 'L' || this.lastUsedJoyconMode === 'R') {
-			// L or R mode
-			if (Math.abs(input.axes[0]) > this.deadBand) {
-				if (this.lastUsedJoyconMode === 'L') {
-					return input.axes[0] * -1 // in this mode, L is "negative"
-				} else if (this.lastUsedJoyconMode === 'R') {
-					return input.axes[0] * 1.4 // in this mode, R is "positive"
+			} else if (joycon.mode === 'LR') {
+				// L + R mode
+				// get the first one that is moving outside of the deadband, priorotizing the L controller
```
baltedewit, line 240: Behaviour appears to be incorrect compared to this comment. Currently the right one will overwrite left if it moves out of the deadband.

baltedewit (changes requested):

meteor/client/ui/Prompter/controller/joycon-device.ts
```diff
@@ -2,7 +2,8 @@ import { ControllerAbstract } from './lib'
 import { PrompterConfigMode, PrompterViewInner } from '../PrompterView'
 import Spline from 'cubic-spline'
 
-type JoyconMode = 'L' | 'R' | 'LR'
+type JoyconWithData = { index: number; mode: JoyconMode; axes: readonly number[]; buttons: number[] }
```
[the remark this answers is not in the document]
jesperstarkar: @jstarpl it comes from the standard typings for gamepads - I have not invented this.

TypeScript says that the number we are reading from is readonly, and requires that when we assign it we keep that type. I just did so, and didn't think more about it. I don't want to add stricter types than needed, neither do I want to cast or work around this warning.

Agree?

image

meteor/client/ui/Prompter/controller/joycon-device.ts
jesperstarkar: Fixed by adding the second clause in elseif instead of another if.

PR #716, opened 2022-04-13
feat: store user event timestamps

jstarpl:

Feature

The time it takes for a user event initiating an Action in Sofie and that being delivered to Sofie Core isn't tracked

Whenever a Browser event initiates a Sofie Action, the timeStamp of that event is stored, transformed into "Sofie" time and then sent along into Core for storing along side the UserActionLog entry. This allows tracking how long did it take for the user action to be processed in the GUI and how long it spent on the wire.

* **Other information**:

**Status**


nytamin (commented):
The expression `msg.timestamp - msg.clientTime` displayed in the GUI might be misleading, since the clients' time _might_ not be in sync with the server. So there is a slight risk that the GUI time is due to a mis-sync.

If we want to get a proper GUI time, we would need to do something like:

```
Client                 Server

CallUserActionX
                   \
                    \  
                      ReplyHasStarted    executeActionX
                    /                       |
                   /                        |
CalcAndSendDeltaTime                        |
                    \                       |
                     \                      |
                      StoreDeltaTime        |
                                            |
                                            v
                                            Finish
                                            StoreExecutionTimes
```

But of course this PR is a step in the right direction and I would be happy with us improving it in a later PR.

jstarpl:
> The expression `msg.timestamp - msg.clientTime` displayed in the GUI might be misleading, since the clients' time _might_ not be in sync with the server. So there is a slight risk that the GUI time is due to a mis-sync.
> 
> If we want to get a proper GUI time, we would need to do something like:
> (...)
> But of course this PR is a step in the right direction and I would be happy with us improving it in a later PR.

Yeah, so I think the GUI is now supposed to resync to Core every 5 minutes. I think it's safe to assume it's "good enough", although I agree that it might not be entirely accurate.

One thing that I would like to ask about your graph there: How would `CalcAndSendDeltaTime` work? Would it assume that half of the difference between _local client timestamp_ for `CallUserActionX`_ and _local timestamp_ for `CalcAndSendDeltaTime` is the GUI + wire time? Because I can still see situations where that assumption might not be correct.

jstarpl:
I guess one important thing now is to add code documentation about these innaccuracies we know of, so that we don't try to get accuracy beyond that point.

PR #718, opened 2022-04-28
feat: move more logic into workers

Julusian:

feature

Meteor is sending messages from the ExternalMessageQueue in the background.
There is no need for meteor to do this work, and as it is written it will not scale very well with meteor

The ExternalMessageQueue sending loop/timer/watcher has been moved to the 'events' thread. This thread was previously tasked with low-priority 'events' (yellow line to enps, metadata sending), so was appropriate to add some more timing insensitive logic to.
No new jobs were added, the queue remains running on its own timers and database watching. This shouldnt impact performance negatively, as nothing else in the thread is time sensitive.
As this queue is not being run as a Job, it is the one place in the worker where setInterval/setTimeout are allowed to be used.

In the future, I could see us moving more background tasks to this thread, such as the cronjob to check if there are any playlists with no rundowns, or generating/updating status messages.

Additionally, I have moved the playlist snapshot generation into the worker, so that it is an 'atomic' operation that is done inside of the playlist lock. This will ensure that the snapshot isnt taken mid-way though a save. I am unsure if this is worth the complexity it adds, so am open to discussing whether this should be kept.

* **Other information**:

**Status**


Julusian:
I need to move some tests still, and I havent tested it yet (I am a little unsure how best to do that), but the general architecture and approach can be reviewed

packages/job-worker/src/interval/playlist.ts
```diff
@@ -0,0 +1,107 @@
+// import { RundownPlaylistId } from '@sofie-automation/corelib/dist/dataModel/Ids'
```
Julusian, line 1: I didn't mean to stage this file, it was from another experiment

Julusian:
I have given slack based external messages a test, and they get sent properly. I do not expect to test rabbitmq, as the changes to the specific sections are not invasive enough to feel it is justified.

All that is left of this is to migrate over the unit tests

packages/job-worker/src/db/collections.ts
```diff
@@ -92,44 +113,76 @@ export interface IDirectCollections {
 	ExternalMessageQueue: ICollection<ExternalMessageQueueObj>
 }
 
-export function getMongoCollections(client: MongoClient, dbName: string): Readonly<IDirectCollections> {
+export function getMongoCollections(
+	client: MongoClient,
+	dbName: string,
+	allowWatchers: boolean
```
nytamin, line 119: This parameter could use some documentation. What is it's purpose?
(same thing with `#allowWatchers` in `WrappedCollection`)

packages/job-worker/src/workers/events/child.ts
```diff
@@ -81,6 +96,8 @@ export class EventsWorkerChild {
 		}
 
 		try {
+			if (this.#externalMessageQueue) this.#externalMessageQueue.invalidateCaches(data)
```
nytamin, line 99: -> `this.#externalMessageQueue?.invalidateCaches(data)`

packages/job-worker/src/db/collections.ts
```diff
@@ -58,6 +60,25 @@ export interface ICollection<TDoc extends { _id: ProtectedString<any> }> {
 	replace(doc: TDoc | ReadonlyDeep<TDoc>): Promise<boolean>
 
 	bulkWrite(ops: Array<AnyBulkWriteOperation<TDoc>>): Promise<unknown>
+
+	/**
+	 * Watch the collection for changes
+	 * Note: Not available in all workers/situations
```
nytamin, line 66: This is a quite scary comment :) could you elaborate?

nytamin (approved):
I left a few minor comments, other than that this looks good!

packages/job-worker/src/db/collections.ts
Julusian: better?

packages/job-worker/src/db/collections.ts
Julusian: better?

PR #719, opened 2022-05-04
fix: OAuth code returned via redirect_uri

thomasslee97:

This fixes the OAuth authentication procedure used by the Spreadsheet Gateway.

Currently the Spreadsheet Gateway relies on the user approving access to their Google Drive, then copying and pasting a code from a page displayed by Google into a form hosted in Sofie.

Since this was first implemented, Google has deprecated (and subsequently removed) support for this copy-paste method of entering access codes. Instead, it is required that Sofie provides a callback url which can be provided with the code as a URL parameter.

Two new URL handlers have been added:
- `/devices/:deviceId/oauthResponse`
  - This handles the callback from the OAuth process and forwards the access code to the gateway
- `/devices/:deviceId/resetAuth`
  - The settings form for the OAuth flow now provides a button for resetting / clearing the auth settings from a peripheral device, offering a way to "sign out" from the Drive integration

* **Other information**:

There have also been a couple of QoL improvements made to the OAuth form, as well as removing references to "spreadsheet" to allow the form to be reused for other OAuth-based services.

**Status**


meteor/server/api/peripheralDevice.ts
```diff
@@ -609,6 +610,95 @@ PickerPOST.route('/devices/:deviceId/uploadCredentials', (params, req: IncomingM
...
+	} catch (e) {
+		res.statusCode = 500
+		content = e + ''
+		logger.error('Upload credentials failed: ' + e)
+	}
+
+	res.end(content)
+})
+
+PickerGET.route('/devices/:deviceId/resetAuth', (params, req: IncomingMessage, res: ServerResponse) => {
```
Julusian, line 667: This should not be a GET. Perhaps a POST or maybe DELETE?

meteor/server/api/peripheralDevice.ts
```diff
@@ -609,6 +610,95 @@ PickerPOST.route('/devices/:deviceId/uploadCredentials', (params, req: IncomingM
...
+
+		let accessToken = url.searchParams.get('code') || undefined
+		const scopes = url.searchParams.get('scope') || undefined
+
+		check(accessToken, String)
+		check(scopes, String)
+
+		accessToken = (accessToken + '').trim()
+		if (accessToken && accessToken.length > 5) {
+			PeripheralDeviceAPI.executeFunction(deviceId, 'receiveAuthToken', accessToken)
```
Julusian, line 642: Is it intentional that we do the 302 redirect even if this executeFunction fails? A comment to state this and why would be good

meteor/server/api/peripheralDevice.ts
```diff
@@ -609,6 +610,95 @@ PickerPOST.route('/devices/:deviceId/uploadCredentials', (params, req: IncomingM
...
+				.then(() => {
+					logger.info(`Sent auth token to device "${deviceId}"`)
+				})
+				.catch(logger.error)
+		}
+
+		const coreUrl = new URL(
+			`/settings/peripheralDevice/${deviceId}`,
```
Julusian, line 650: I feel like it is possible to omit the `http://localhost:3000` from this redirect, but it has been a while

jstarpl: Yes, it is perfectly safe to not provide protocol & host in the Location header.

Julusian (approved):

PR #720, opened 2022-05-12
fix(Prompter): If collection changes are slow, Prompter may receive incomplete updates, and render some Script missing

jstarpl:

Bug fix

In non-ideal network situations, it's possible that the Piece/PieceInstance collection updates will not stream at once into the Prompter. If that takes more than 250 ms, then the prompter will update once (without a Piece of script) and then once again (with a Piece of Script shown again).

This change introduces an alternative timer that, if a Script Piece is not present in an update, a 2s delay will wait for that Script Piece to re-appear, before rendering it gone. This in turn, should prevent the Prompter scroll position locking to lock to a new Piece, because the old Piece was gone for a second.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Prompter/PrompterView.tsx
+++ b/meteor/client/ui/Prompter/PrompterView.tsx
@@ -26,6 +26,10 @@ import { StudioScreenSaver } from '../StudioScreenSaver/StudioScreenSaver'
 import { RundownTimingProvider } from '../RundownView/RundownTiming/RundownTimingProvider'
 import { OverUnderTimer } from './OverUnderTimer'
 import { Rundown, Rundowns } from '../../../lib/collections/Rundowns'
+import { PieceId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+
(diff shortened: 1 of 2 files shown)


meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -696,9 +700,45 @@ export const Prompter = translateWithTracker<IPrompterProps, {}, IPrompterTracke
...
+
+			const currentPrompterPieces = _.flatten(
+				prompterData?.segments.map((segment) =>
+					segment.parts.map((part) =>
+						// collect all the PieceId's of all the non-empty pieces of script
+						_.compact(part.pieces.map((dataPiece) => (dataPiece.text !== '' ? dataPiece.id : null)))
+					)
+				) ?? []
+			) as PieceId[]
+			const nextPrompterPieces = _.flatten(
```
nytamin, line 715: nit-picking: This implementation is identical to the one on line 707

nytamin (approved):
Seems like a reasonable fix for a troublesome issue :)

PR #721, opened 2022-05-13
Fix: mos gateway command timeout

nytamin:

Improvement of error message and handling of timeouts in MOS messages.

If a MOS message times out, Core aborts too early with a generic and unhelpful timeout message.

The MOS-Gateway is given a specified timeout time, and calls to it are set to wait a certain (longer) time, ensuring that timed out MOS-messages are handled primarily in MOS-gateway, so errors should be more helpful.

* **Other information**:

**Status**


packages/mos-gateway/package.json
```diff
@@ -66,7 +66,8 @@
 	],
 	"dependencies": {
 		"@sofie-automation/server-core-integration": "^1.42.0-in-testing.0",
-		"mos-connection": "^1.0.6",
+		"@sofie-automation/corelib": "^1.42.0-in-testing.0",
```
Julusian, line 69: The gateways arent supposed to rely on corelib
There isnt any particular reason for that other than it was written as a shared ground between meteor and the job-worker.
Could these constants go in `@sofie-automation/server-core-integration`?
Or is this the motivation needed to split `@sofie-automation/server-core-integration` in half to have a typings half and a '@sofie-automation/sofie-gateway-connection` half?

nytamin: I can move the constants into `@sofie-automation/server-core-integration`, good catch! 👍

nytamin: Correction: I'll tear the constants apart for now, we'll be adding a share libary later.

PR #730, opened 2022-06-09
shared-lib package

nytamin:
This PR adds another package to the mono-repo.

The **shared-lib** package is intended to be used Core, but also by gateways.
I've just put some constants in there for now, but I will in later PRs move other constants and enums to this package, such as the method definitions for the peripheral-devices.


packages/mos-gateway/package.json
```diff
@@ -65,6 +65,7 @@
 	],
 	"dependencies": {
 		"@sofie-automation/server-core-integration": "1.44.0-in-development",
+		"@sofie-automation/shared-lib": "link:../shared-lib",
```
Julusian, line 68: This should probably be an done with a version number lib server-core-integration.

There isnt technically much/any difference, but consistency

.github/workflows/node.yaml
```diff
@@ -299,6 +300,9 @@ jobs:
           - node-version: 14.x
             package-name: corelib
             send-coverage: true
+          - node-version: 14.x
+            package-name: shared-lib
```
Julusian, line 304: This should probably be in the `package-name` array above, as we should test it for the same runtimes as `server-core-integration`

packages/shared-lib/package.json
```diff
@@ -0,0 +1,60 @@
+{
+	"name": "@sofie-automation/shared-lib",
+	"version": "1.44.0-in-development",
+	"private": true,
```
Julusian, line 4: this should not be set to private, as we expect to publish it to npm

packages/shared-lib/package.json
```diff
@@ -0,0 +1,60 @@
...
+		"/dist",
+		"/CHANGELOG.md",
+		"/README.md",
+		"/LICENSE"
+	],
+	"dependencies": {
+		"tslib": "^2.4.0"
+	},
+	"prettier": "@sofie-automation/code-standard-preset/.prettierrc.json",
+	"husky": {
```
Julusian, line 47: didnt I remove this husky block from the other libs?

packages/shared-lib/package.json
```diff
@@ -0,0 +1,60 @@
...
+		"build:main": "tsc -p tsconfig.build.json",
+		"lint:raw": "eslint --ext .ts --ext .js --ignore-pattern dist",
+		"lint": "yarn lint:raw .",
+		"unit": "jest",
+		"test": "yarn lint && yarn unit",
+		"watch": "jest --watch",
+		"cov": "jest --coverage; open-cli coverage/lcov-report/index.html",
+		"cov-open": "open-cli coverage/lcov-report/index.html",
+		"release": "yarn reset && yarn test && yarn changelog",
+		"reset": "git clean -dfx && git reset --hard && yarn",
```
Julusian, line 29: This was probably in the other libs, but this reset line is odd to have deep in here. Itll affect the whole clone, so should it really be in here?
(Also, Ive never used it, so question its purpose)

Julusian (commented):
I think I need to read through some of the existing package.json and tidy out some junk that has been accumulating, there are a few quesitonable bits that I were copied into here.

But in general looks fine :)

packages/shared-lib/package.json
nytamin: doesn't look like it. I based this package.json from the one in corelib. I leave it here and we can remove it when cleaning up the libs later

packages/shared-lib/package.json
nytamin: I agree, it's also present in corelib, and job-worker. I'm leaving it here as well and we'll remove it in an upcoming cleanup.

packages/shared-lib/package.json
nytamin: It's still present in corelib and job-worker. I'll leave it for now and we can clean it up later when going through all the libs.

PR #733, opened 2022-06-20
chore: Move various types to shared-lib

nytamin:
This PR moves a few types to the previously added shared-lib package.

The idea is to reduce duplications and enable exposure of shared types, constants and enums to gateways or other external applications that talk to Sofie.

This PR does not aim to be the _ultimate move of all shared types_. Moving of other types will follow later.

Co-authored by @Julusian


meteor/client/ui/RundownView.tsx
```diff
@@ -106,7 +106,7 @@ import { RundownLayoutsAPI } from '../../lib/api/rundownLayouts'
 import { TriggersHandler } from '../lib/triggers/TriggersHandler'
 import { SorensenContext } from '../lib/SorensenContext'
 import { PlaylistTiming } from '@sofie-automation/corelib/dist/playout/rundownTiming'
-import { CASPARCG_RESTART_TIME } from '@sofie-automation/shared-lib'
+import { CASPARCG_RESTART_TIME } from '@sofie-automation/shared-lib/dist/core/constants'
```
LindvedKrvang, line 109: What is the reason for importing from the `dist` folder?

meteor/lib/api/mediaManager.ts
```diff
@@ -1,22 +1,8 @@
+import * as Shared from '@sofie-automation/shared-lib/dist/peripheralDevice/mediaManager'
 export namespace MediaManagerAPI {
```
LindvedKrvang, line 2: Isn't this class redundant now and the places that need the WorkStep Enums should just import them from where they reside now?

packages/blueprints-integration/src/status.ts
```diff
@@ -1,14 +1 @@
...
-	/** Not everything is OK, but normal operation is not affected */
-	WARNING_MINOR = 2,
-	/** Not everything is OK, normal operation might be affected */
-	WARNING_MAJOR = 3,
-	/** Normal operation is affected, automatic recover might be possible */
-	BAD = 4,
-	/** Normal operation is affected, automatic recover is not possible (manual interference is required) */
-	FATAL = 5,
-}
+export { StatusCode } from '@sofie-automation/shared-lib/dist/lib/status'
```
LindvedKrvang, line 1: Maybe it's my Typescript knowlegde that is lacking, but what is the idea of importing something with the sole reasons to just re-export it and nothing else? (You do this through out the codebase of Sofie)
It seems to add an extra unecssary layer of dependencies and not much else?

This file is a sole line that re-export an Enum. What does this file add of value? What's wrong with just importing the Enum from where it is defined in the files that actually needs the Enum?

meteor/client/ui/RundownView.tsx
nytamin: @Julusian can probably answer this even better, but as I understood it there is a bug in Typescript that prevents us from importing from the proper path (ie `@sofie-automation/shared-lib/core/constants`).

It would of course be possible to add exports to the main index.ts-file in shared-lib and just import everything from `@sofie-automation/shared-lib`, but that might become a hassle later as the library grows and contains many different kinds of types/constants/helper functions.

So we opted to import using the `/dist/` path for now, and when the Typescript-bug is solved we can remove it later.

packages/blueprints-integration/src/status.ts
nytamin: However, if you think it would be better to change the imports and handle the merge conflicts on our next sync, I would be happy to remove the import-exports right away.

Julusian: Yeah we did the same thing when doing refactoring for the job-worker, to minimise the amount of churn happening all at once. We probably should be working on removing the aliases introduced by the job-worker, but its now a challenge of motivating the refactoring (it is very tedious to fix it all up)

meteor/client/ui/RundownView.tsx
Julusian: Yes, I believe we need typescript 4.7 (we currently use 4.5) to be able to drop the `/dist/` portion of the import path. https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing
We felt it was better to have this there temporarily (we made the same decision when doing the job-worker, back when support for this was aiming for 4.5), rather than to export everything through the index.ts, as that would require us to rename a lot of the exports to ensure their name/purpose is clear without the context of file you are importing from

LindvedKrvang: Isn't the bug just that Typescripts suggest the wrong path for auto-import? The right path should still work?

Nevertheless. Glad is temporary

packages/blueprints-integration/src/status.ts
LindvedKrvang: Yeah the reason I asked is because this is not the first time I'm seeing this in Sofie, so I thought that this might be a practice that you were doing, so I wanted to understand it. Glad that it's not your normal practice. 

Personally, I favour just using the correct import and fix the merge-conflict (import conflict is useally quickfixes).
But I can see an argument for waiting untill we atleast get our changes back to you.

nytamin: Agreed, we'll keep this policy for a little while longer then.

LindvedKrvang (approved):

meteor/client/ui/RundownView.tsx
Julusian: When compiling it also fails to resolve the correct path, causing it to report every import as missing a file.
The right path does work at runtime, but with typescript unable to resolve those paths at all, we have been a little stuck

packages/blueprints-integration/src/status.ts
Julusian: To me this isnt solely about merge conflicts with TV2, but also reducing the noise in the PRs. Asking someone to review a PR with 15 changed files instead of 100. Even though its a simple change in most of those, its a nicer review to have two PRs, but we never got to doing the second

PR #735, opened 2022-06-21
Tv2/big sync aftermatch changes

LindvedKrvang:
As in agreement with @nytamin we are creating a PR so NRK can start looking at the TV 2 changes made after the first round of the BigSync


Julusian (commented):
There is a lot of changes for the tv2 repo structure included here, which I am ignoring.

From a skim of the backend/packages portions, I only saw a couple of small issues

PR #737, opened 2022-06-28
Feat/batched playout callbacks

nytamin:
Note: This PR is based on https://github.com/nrkno/sofie-core/pull/733

Refactor + Performance improvement

Playout-Gateway sends individual updates on:

  * Part playback started (used for auto-nexting & sending metadata)
  * Part playback stopped (used to send metadata)
  * Piece playback started (used to send metadata)
  * Piece playback stopped (used to send metadata)

Playout-Gateway batches the above updates together and sends them as a single call (`playoutPlaybackChanged`).
This should improve performance a bit, since multiple updates can be done within a single PlayoutCache and Playlist-lock.

* **Other information**:

**Status**


packages/shared-lib/src/peripheralDevice/peripheralDeviceAPI.ts
```diff
@@ -26,16 +26,41 @@ export interface PiecePlaybackStartedResult extends PiecePlaybackCallbackData {
...
-export type PlayoutChangedResults = PlayoutChangedResult[]
+export type PlayoutChangedResults = {
+	rundownPlaylistId: RundownPlaylistId
+	changes: PlayoutChangedResult[]
+}
+export enum PlayoutChangedType {
+	partPlaybackStarted = 'partPlaybackStarted',
+	partPlaybackStopped = 'partPlaybackStopped',
+	piecePlaybackStarted = 'piecePlaybackStarted',
+	piecePlaybackStopped = 'piecePlaybackStopped',
```
ianshade, line 37: Can we avoid camel case in enum members?

packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -227,24 +234,97 @@ export class TSRHandler {
...
-					})
-			} else {
-				this.logger.error(`Unknown callback method "${callbackName}"`)
+
+		let changedResults: PlayoutChangedResults | undefined = undefined
+		let sendCallbacksTimeout: NodeJS.Timer | undefined = undefined
+
+		this.tsr.on(
+			'timelineCallback',
+			(time, objId, callbackName0, data: PartPlaybackCallbackData | PiecePlaybackCallbackData) => {
```
ianshade, line 243: `init` got quite long. Can implementations of these callbacks be extracted out of it?

packages/job-worker/src/playout/playout.ts
```diff
@@ -981,36 +939,92 @@ export async function onPartPlaybackStopped(context: JobContext, data: OnPartPla
...
+		data,
+		(cache) => {
+			const playlist = cache.Playlist.doc
+			if (!playlist) throw new Error(`RundownPlaylist "${data.playlistId}" not found!`)
+		},
+		async (cache) => {
+			const playlist = cache.Playlist.doc
+			if (!playlist) throw new Error(`RundownPlaylist "${data.playlistId}" not found!`)
+
+			for (const change of data.changes) {
```
Julusian, line 995: I would like for us to consider doing this as state rather than commands. It may not be something to do with this PR though.
There is another documented weakness in the current callback flow, where if core doesnt receive/handle a callback then playout-gateway never retries it and it gets lost. Maybe this could be fixed with some retry logic, but I expect it will be simpler to make durable if we transfer state rather than commands

packages/corelib/src/worker/studio.ts
```diff
@@ -36,10 +37,16 @@ export enum StudioJobs {
 	RemovePlaylist = 'removePlaylist',
 	RegeneratePlaylist = 'regeneratePlaylist',
 
+	/** @deprecated use OnPlayoutPlaybackChanged instead */
```
Julusian, line 40: Why are we keeping these deprecated methods?
We already require server-core-integration to be the in the same semver minor range.

Im worried that there is some complexity added by keeping these methods around, and if we don't clean it up now then I dont know when we will

ianshade: If we're going to take these Core and Playout Gateway changes simultaneously, would these methods even be reachable? I don't think we have any other devices that could call them.

packages/job-worker/src/playout/playout.ts
```diff
@@ -837,140 +795,140 @@ export async function onPartPlaybackStarted(context: JobContext, data: OnPartPla
...
+			await _onPartPlaybackStarted(context, cache, data)
+		}
+	)
+}
+async function _onPartPlaybackStarted(context: JobContext, cache: CacheForPlayout, data: OnPartPlaybackStartedProps) {
+	const playingPartInstance = cache.PartInstances.findOne(data.partInstanceId)
+	if (!playingPartInstance)
+		throw new Error(`PartInstance "${data.partInstanceId}" in RundownPlayst "${data.playlistId}" not found!`)
+
+	// make sure we don't run multiple times, even if TSR calls us multiple times
```
LindvedKrvang, line 807: Comments like this that explains what is happening instead of why it's happening is a bad practice and should be avoided.
If you feel a need to include a comment like this then that's a sign that your code isn't clean enough.

packages/job-worker/src/playout/playout.ts
```diff
@@ -837,140 +795,140 @@ export async function onPartPlaybackStarted(context: JobContext, data: OnPartPla
...
-						? cache.Rundowns.findOne(currentPartInstance.rundownId)
-						: undefined
-					const showStyleRundown = currentRundown ?? rundown
-					const showStyle = await context.getShowStyleCompound(
-						showStyleRundown.showStyleVariantId,
-						showStyleRundown.showStyleBaseId
+		if (playlist.currentPartInstanceId === data.partInstanceId) {
+			// this is the current part, it has just started playback
+			if (playlist.previousPartInstanceId) {
+				if (!previousPartInstance) {
```
LindvedKrvang, line 826: You are four layers of nesting deep here... :( 
A good rule of thumb is to try and only have one layer of nesting for a method

packages/job-worker/src/playout/playout.ts
```diff
@@ -837,140 +795,140 @@ export async function onPartPlaybackStarted(context: JobContext, data: OnPartPla
...
-							)
-						} else if (!previousPartInstance.timings?.duration) {
-							onPartHasStoppedPlaying(cache, previousPartInstance, data.startedPlayback)
-						}
-					}
+			await _onPartPlaybackStarted(context, cache, data)
+		}
+	)
+}
+async function _onPartPlaybackStarted(context: JobContext, cache: CacheForPlayout, data: OnPartPlaybackStartedProps) {
```
LindvedKrvang, line 802: This method is almost 130 lines long.
A good rule of thumb is to try and keep them around 25-30 lines.
A good place to start refactoring this to smaller methods would be to look at your comments that explain what is happening and extract those sections.

packages/job-worker/src/playout/playout.ts
```diff
@@ -981,36 +939,92 @@ export async function onPartPlaybackStopped(context: JobContext, data: OnPartPla
...
-				rundownId: { $in: rundowns.map((r) => r._id) },
-			})
+	const partInstance = cache.PartInstances.findOne({
+		_id: data.partInstanceId,
+		rundownId: { $in: rundowns.map((r) => r._id) },
+	})
 
-			if (partInstance) {
-				// make sure we don't run multiple times, even if TSR calls us multiple times
+	if (partInstance) {
```
LindvedKrvang, line 958: Another method that could benefit from inverting the if-statement

packages/shared-lib/src/lib/lib.ts
```diff
@@ -1,2 +1,12 @@
 export type Time = number
 export type TimeDuration = number
+
+export function assertNever(_never: never): void {
```
LindvedKrvang, line 4: I'm curious. Is this method usefull?
If you try to call it, wouldn't you just get an error saying you are trying to pass something to type never, making a call to this method uncompilable?
If above is true - then what's the purpose of having a method in the code we can't call?
Or am I missing something about the "never" type?

packages/shared-lib/src/peripheralDevice/peripheralDeviceAPI.ts
LindvedKrvang: I would highly suggest that we use SNAKE_CASE

packages/job-worker/src/playout/playout.ts
Julusian: Ive never gotten along with the number of lines rule. Ive often found it to want me to split a method with no logical split point, or resulting in a method that is then problematic because of number of parameters.
And it also makes it harder for future improvements when you to change how data flows between these methods.

I agree that some of the methods on Sofie are way too long, but Im not convinced this is one of those

packages/shared-lib/src/lib/lib.ts
Julusian: This method is for doing some type checks. You can't pass any valid value to this method (number/string/object/any is not assignable to never). So we use this method to help us make sure that a path is dead code.

For example, if a new value `Test = 'test'` was added to the enum, the typescript compiler would error on the line this was called saying `Test is not assignable to never`. 
Without using `assertNever` we would be relying on manual testing and unit testing to ensure that everywhere we branch based on the enum value handles the new value. But instead we can use the typescript compiler error as a much quicker view of places that need to be updated. It doesnt catch everything, but it will catch a lot of them

LindvedKrvang: The reason I am asking is because I tried using it once and if I parsed something as a parameter I got a compiler error saying that I couldn't parse something to type never.
If I didn't parse any arguments I got a compiler error saying that the method was expecting a parameter but got none.
Either way, I couldn't compile my code if I tried to use it, so I fail to see the use of it.
But maybe I am just using it wrong?

packages/job-worker/src/playout/playout.ts
LindvedKrvang: > Ive often found it to want me to split a method with no logical split point

Hence it's a rule of thumb and not a set in stone rule :) 

> makes it harder for future improvements

I would argue that having these long method makes it even harder for future improvements. A variable that is used multiple times over the span of 100+ lines becomes near impossible to keep track of and makes it so much easier to introduce unknown side effects. 
Also, the sheer amount of logic you need to consider when dealing with a long method compared to a small method makes the method more brittle, makes it more difficult for newer team members to understand the method and increases the amount of time it takes to understand the method well enough to make changes to it.

> resulting in a method that is then problematic because of number of parameters

I would could this a side effect of having the long methods. Having these longs methods where we have access to our local variables all of the time allows us to be "lazy" with our design. Since we just have everything available all of the time we start using everything all of time, making "spaghetti code" within a single function which makes it a parameter hell trying to extract the long methods into smaller methods.

> but Im not convinced this is one of those

Let's agree to disagree then :)

LindvedKrvang: If you want more opinions than mine then I would highly recommend that you read "Clean Code" by Robert C. Martin if you haven't allready.

packages/corelib/src/worker/studio.ts
nytamin: Ok, I will remove them

meteor/server/api/peripheralDevice.ts
```diff
@@ -355,6 +356,33 @@ export namespace ServerPeripheralDeviceAPI {
...
+		const transaction = profiler.startTransaction('playoutPlaybackChanged', apmNamespace)
+
+		// This is called from the playout-gateway when a part starts playing.
+		// Note that this function can / might be called several times from playout-gateway for the same part
+		const peripheralDevice = checkAccessAndGetPeripheralDevice(deviceId, token, context)
+
+		if (!peripheralDevice.studioId)
+			throw new Error(`PeripheralDevice "${peripheralDevice._id}" sent piecePlaybackStarted, but has no studioId`)
+
+		if (changedResults.changes.length) {
```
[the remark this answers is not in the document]
nytamin: That would not be a good idea, as that would skip the rest of the function (`transaction?.end()`).

packages/job-worker/src/blueprints/events.ts
```diff
@@ -130,50 +128,54 @@ export function reportPartInstanceHasStopped(
...
-	playlist: ReadonlyDeep<DBRundownPlaylist>,
-	pieceInstance: Pick<PieceInstance, '_id' | 'partInstanceId' | 'infinite'>,
+	cache: CacheForPlayout,
+	pieceInstance: PieceInstance,
 	timestamp: Time
-): Promise<void> {
-	await Promise.all([
-		context.directCollections.PieceInstances.update(pieceInstance._id, {
+): void {
+	if (pieceInstance.startedPlayback !== timestamp) {
```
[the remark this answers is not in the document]
nytamin: I consider this to be two different schools of code styles / personal taste. I've used both in different situations and I'd be happy to discuss this in one of our upcoming talks.

packages/job-worker/src/blueprints/events.ts
```diff
@@ -130,50 +128,54 @@ export function reportPartInstanceHasStopped(
...
-): Promise<void> {
-	await context.directCollections.PieceInstances.update(pieceInstance._id, {
-		$set: {
-			stoppedPlayback: timestamp,
-		},
-	})
-
-	handlePartInstanceTimingEvent(context, playlist._id, pieceInstance.partInstanceId)
+): void {
+	if (pieceInstance.stoppedPlayback !== timestamp) {
```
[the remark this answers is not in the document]
nytamin: ref: https://github.com/nrkno/sofie-core/pull/737#discussion_r912742322

packages/job-worker/src/playout/playout.ts
```diff
@@ -693,139 +701,89 @@ export async function disableNextPiece(context: JobContext, data: DisableNextPie
...
-						`onPiecePlaybackStarted: Playout reports pieceInstance "${
-							data.pieceInstanceId
-						}" has started playback on timestamp ${new Date(data.startedPlayback).toISOString()}`
-					)
-					await reportPieceHasStarted(context, playlist, pieceInstance, data.startedPlayback)
+function _onPiecePlaybackStarted(context: JobContext, cache: CacheForPlayout, data: OnPiecePlaybackStartedProps) {
+	const playlist = cache.Playlist.doc
+	const pieceInstance = cache.PieceInstances.findOne(data.pieceInstanceId)
+
+	if (pieceInstance) {
```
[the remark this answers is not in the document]
nytamin: ref: https://github.com/nrkno/sofie-core/pull/737#discussion_r912742322

packages/job-worker/src/playout/playout.ts
```diff
@@ -693,139 +701,89 @@ export async function disableNextPiece(context: JobContext, data: DisableNextPie
...
+		async (cache) => {
+			_onPiecePlaybackStopped(context, cache, data)
 		}
 	)
 }
+function _onPiecePlaybackStopped(context: JobContext, cache: CacheForPlayout, data: OnPiecePlaybackStoppedProps) {
+	const playlist = cache.Playlist.doc
+	const pieceInstance = cache.PieceInstances.findOne(data.pieceInstanceId)
+
+	if (pieceInstance) {
```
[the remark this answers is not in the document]
nytamin: ref: https://github.com/nrkno/sofie-core/pull/737#discussion_r912742322

packages/shared-lib/src/lib/lib.ts
jstarpl: @LindvedKrvang you can see it in use here: https://github.com/nrkno/sofie-core/blob/a6fe9c87c3ce2a2e9ca627368a77bd47fcf95e7b/packages/job-worker/src/playout/timeline/piece.ts#L52 and here: https://github.com/nrkno/sofie-core/blob/a12f7d8082fe1c08b5c68d6d536cbc5b32c1aebf/meteor/lib/api/triggers/actionFactory.ts#L545

I find `assertNever` insanely useful. Makes adding new features very easy, as long as you're following a certain pattern of typing objects. If you passed something as a parameter, and got a compiler error that means that whatever you were passing was in fact not of type `never`, which means that there was some case that you haven't handled in your code and it was in fact possible to get into the codepath where `assertNever` was used: either because of not-so-strict typing or an actual case that was unhandled. The good thing is that the error that TypeScript will give on that `assertNever` will hint at what case needs to be handled.

packages/job-worker/src/playout/playout.ts
nytamin: That line does not originate from this PR, so I will not adress this here.
It is an interesting thought you bring up though, I'll make sure we talk about it in our upcoming talk.

packages/job-worker/src/playout/playout.ts
nytamin: ref: https://github.com/nrkno/sofie-core/pull/737#discussion_r912742322

packages/job-worker/src/playout/playout.ts
nytamin: I'll make sure to bring this up at our upcoming talk.

packages/job-worker/src/playout/playout.ts
nytamin: ref: https://github.com/nrkno/sofie-core/pull/737#discussion_r912742322

packages/shared-lib/src/lib/lib.ts
nytamin: Here's an example:
```
enum Animal {
	DOG,
	CAT,
}
function isRound(animal: Animal): boolean {
	switch (animal) {
		case Animal.DOG:
			return true
		case Animal.CAT:
			return false
		default:
			// Ensure we've handled all of the cases:
			assertNever(animal)
			return false
	}
}
```
This compiles without errors.
But if you add a `FISH` to the enum, you'll get a compiler error as a reminder that you have to remember to handle that case in the switch too. So it's basically a guard against future bugs/unhandled cases.

packages/shared-lib/src/peripheralDevice/peripheralDeviceAPI.ts
nytamin: Good point! SNAKE_CASE it is!

PR #742, opened 2022-07-01
feat: strongly type meteor publications

Julusian:

code quality

Most of the meteor publications are using parameters which are implicitly any. A few were defining types, but many were not.
In the UI, when subscribing to these, there was no type safety in place except for the id of the publication.

Our wrappings around the meteor methods have been updated to have strongly defined typings.
There is an interface describing the possible combinations which with some generics and some infered types, provides us with a type safe usages where typescript knows what the valid parameters are and what we are calling it with.

* **Other information**:

So far this is limited to meteor (core and ui), and not the gateways. We should look at getting the same types there too, but that is more effort than doing it inside of meteor.

As a follow up we should discuss whether we should have these mongo selectors as parameters to the publications. While it is easy to implement, it would be safer if we were to construct the mongo queries on the server side based on a few simple parameters provided.
Also, almost every publication accepts an auth token as the last parameter. I think this is for gateways to be able to use them. Should gateways be able to use all the publications? And is this token parameter the best way to achieve this?

**Status**

### DIFF
--- a/meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
+++ b/meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
@@ -11,7 +11,7 @@ import {
 import { faCaretDown, faCaretRight, faDownload, faPlus, faUpload } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { TriggeredActionEntry, TRIGGERED_ACTION_ENTRY_DRAG_TYPE } from './TriggeredActionEntry'
-import { literal, omit, protectString, unprotectString } from '../../../../../lib/lib'
+import { literal, omit, unprotectString } from '../../../../../lib/lib'
 import { TriggersHandler } from '../../../../lib/triggers/TriggersHandler'
 import {
 	RundownPlaylist,
@@ -97,9 +97,7 @@ export const TriggeredActionsEditor: React.FC<IProps> = function TriggeredAction
 	}
 
 	useSubscription(PubSub.triggeredActions, showStyleBaseSelector)
-	useSubscription(PubSub.rundowns, {
-		showStyleBaseId: showStyleBaseId ?? protectString('Ignore All'),
-	})
+	useSubscription(PubSub.rundowns, null, showStyleBaseId ? [showStyleBaseId] : [])
 
 	useEffect(() => {
 		const debounce = setTimeout(() => {
@@ -213,13 +211,8 @@ export const TriggeredActionsEditor: React.FC<IProps> = function TriggeredAction
 		null
 	)
 
-	useSubscription(PubSub.partInstances, {
-		rundownId: rundown?._id ?? protectString('Ignore All'),
-		playlistActivationId: rundownPlaylist?.activationId,
-	})
-	useSubscription(PubSub.parts, {
-		rundownId: rundown?._id ?? protectString('Ignore All'),
-	})
+	useSubscription(PubSub.partInstances, rundown ? [rundown._id] : [], rundownPlaylist?.activationId)
+	useSubscription(PubSub.parts, rundown ? [rundown._id] : [])
 
 	const previewContext = useTracker(
 		() => {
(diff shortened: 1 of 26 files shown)


meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
```diff
@@ -208,11 +214,11 @@ export const TriggeredActionsEditor: React.FC<IProps> = function TriggeredAction
 	)
 
 	useSubscription(PubSub.partInstances, {
-		rundownId: rundown?._id ?? false,
+		rundownId: rundown?._id ?? protectString('Ignore All'),
```
jstarpl, line 217: I suspect that as we move over to React Hooks, this is a pattern that we might see more of. Maybe it should be wrapped into something like `protectedInvalidId()` or something, which would return a static protectedString?

jstarpl (changes requested):
I really like the typed publications, should make maintenance of them much easier. Regarding not passing in the Mongo Selectors, but rather constructing the queries server-side: I suppose that's not a bad idea, but then the flexibility to subscribe to data in almost any shape is very handy when optimizing the amount of data loaded client-side. I feel like it may make more sense to wait with that until we try out the custom, UI-specific publications.

meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
Julusian: yeah that could work, or my suggestion of `we should discuss whether we should have these mongo selectors as parameters to the publications.` would be a proper solution.
In this particular case the publication could be done as `useSubscription(PubSub.partInstances, rundownPlaylist?.activationId, rundown?._id)`. That would let the backend easily handle this and simply not run the mongo query when the rundownId is undefined.

We could probably do something similar with the current api. If this publication gets a subscriber with no rundownId, then return an empty collection. It feels a little dangerous to assume that we will always a rundownId, so the api change would be better, but it might be safe for the current queries (I havent checked yet)

Ill try and remember to bring up this publications api topic at the next devmeet

meteor/lib/api/pubsub.ts
```diff
@@ -59,7 +103,96 @@ export enum PubSub {
 	expectedPackagesForDevice = 'expectedPackagesForDevice',
 }
 
-export function meteorSubscribe(name: PubSub, ...args: any[]): Meteor.SubscriptionHandle {
+export interface PubSubTypes {
```
nytamin, line 106: Should these go into `shared-lib` instead?

Or perhaps the intention is for these to be internal in Core, if so perhaps that could be documented in the code, or even rename to something like `InternalPubSubTypes`? (Or, perhaps such a change would be included in the upcoming PublicAPI-discussion/task.)

nytamin (approved):
Looks good!

meteor/client/ui/Settings/components/triggeredActions/TriggeredActionsEditor.tsx
Julusian: Or more shortterm, to preserve `the flexibility to subscribe to data in almost any shape is very handy`, perhaps we add a new `isValid: boolean` parameter to some of the publications? That way the ui can flag a query as invalid/incomplete and meteor will skip executing it

meteor/lib/api/pubsub.ts
Julusian: Ideally the ones which are relevant to the gateways should be defined in shared-lib, but that adds a lot of typings complications. 
The first problem I hit when looking at moving `timelineForDevice` was that it relies on TSR types, and we don't have that in shared-lib (yet?). Once I found that I gave up on it for now.
Perhaps this is another section of the shared-lib we should pair program? 

Im not sure about the naming for now, I think that will depend on how we decide to split it up between what is for the ui and what is for gateways. (should the gateway bits also be available for the ui?). Are there other valid use cases too that should have their own set?

nytamin: Other possible use cases could be future UI plug-ins, or a public API, or other specific Gateways (package manager, mos-gw, playout-gw, input/xkeys/streamdeck/vsm gateway)
Let's have a look at it together some time later.

Julusian:
Following the discussion, I have reworked the publications where I used `protectedString('Ignore All')` to be more parameter driven.
Based on our current usages, the parts and partInstances ones translated nicely. They were using the same parameters in every call.
The rundowns one was a bit more complicated as they would either provide playlistIds or showStyleBaseIds. But as this is a simple difference, the publication expects both to be provided but if that parameter is undefined then it will be ignored.

Changing the publications like this, means that meteor can see that the 'query' has no rundownIds and so it can ignore the query as it is 'incomplete'

jstarpl (approved):

PR #747, opened 2022-07-19
fix: Prevents "stuck" joycon when going from single to paired or vice…

jesperstarkar:
… versa AND have active joystick input when doing so.

Known bug on MacOS where a "stuck" active joystick won't recover before re-pairing (going to L+R-mode) again.

Fix: Solves an edge case where a joycon can get "stuck" and continue to scroll when going from a L+R pair to a single controller, without any way of stopping it. It is not stop-able.

When having a "stuck" joycon input, nothing can be done to stop it before a L+R pair is connected again.

When having a "stuck" controller, any movement/input on any other joycon unit will take over the control

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Prompter/controller/joycon-device.ts
+++ b/meteor/client/ui/Prompter/controller/joycon-device.ts
@@ -24,6 +24,7 @@ export class JoyConController extends ControllerAbstract {
 	private reverseSpeedSpline: Spline
 
 	private updateSpeedHandle: number | null = null
+	private timestampOfLastUsedJoyconInput: number = 0
 	private currentPosition = 0
 	private lastInputValue = ''
 	private lastButtonInputs: { [index: number]: { mode: JoyconMode; buttons: number[] } } = {}
@@ -225,11 +226,10 @@ export class JoyConController extends ControllerAbstract {
 
 	private getActiveInputsOfJoycons(joycons: JoyconWithData[]): number {
 		let lastSeenSpeed = 0
-		let lastSeenSpeedTimestamp = 0
 
 		for (const joycon of joycons) {
 			// sort/filter by gamepad timestamp to use the most up-to-date input, in order to prevent the "stuck" dead joycon when going from pairs to singles
-			if (joycon.timestamp > lastSeenSpeedTimestamp) {
+			if (joycon.timestamp >= this.timestampOfLastUsedJoyconInput) {
 				// handle buttons at the same time as evaluating stick input
 				this.handleButtons(joycon)
 
@@ -243,7 +243,7 @@ export class JoyConController extends ControllerAbstract {
 							lastSeenSpeed = joycon.axes[0] * 1.4 // in this mode, R is "positive"
 							// factor increased by 1.4 to account for the R joystick being less sensitive than L
 						}
-						lastSeenSpeedTimestamp = joycon.timestamp
+						this.timestampOfLastUsedJoyconInput = joycon.timestamp
 					}
 				} else if (joycon.mode === 'LR') {
 					// L + R mode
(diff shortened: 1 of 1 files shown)


jesperstarkar (commented):
Description of logic

meteor/client/ui/Prompter/controller/joycon-device.ts
```diff
@@ -219,30 +225,36 @@ export class JoyConController extends ControllerAbstract {
...
-			// hadle speed input
-			if (joycon.mode === 'L' || joycon.mode === 'R') {
-				// L or R mode
-				if (Math.abs(joycon.axes[0]) > this.deadBand) {
-					if (joycon.mode === 'L') {
-						lastSeenSpeed = joycon.axes[0] * -1 // in this mode, L is "negative"
-					} else if (joycon.mode === 'R') {
-						lastSeenSpeed = joycon.axes[0] * 1.4 // in this mode, R is "positive"
+			// sort/filter by gamepad timestamp to use the most up-to-date input, in order to prevent the "stuck" dead joycon when going from pairs to singles
+			if (joycon.timestamp > lastSeenSpeedTimestamp) {
```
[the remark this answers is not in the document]
jstarpl: So I had a little bit of a think when doing the dishes today, and I realized that we need to have a persistent (I mean across-frames-persistent, not stored in localStorage) `lastSeenSpeedTimestamp` variable, here's why:

Suppose you have two joycons, one active, the other one stuck in a +50 position. 
1. The Prompter is scrolling, because of the stuck one.
2. You move the active one away from deadband, starts scrolling at the speed of the active one, since that's the one with the latest timestamp
3. You move the active one back into the deadband, which then means that the stuck one has the latest timestamp again, and everything starts scrolling, because the current `lastSeenSpeedTimestamp` is scoped to this, single-frame function, `getActiveInputsOfJoycons`.

The thing is, with a persistent `lastSeenSpeedTimestamp` this comparison needs to be `joycon.timestamp >= lastSeenSpeedTimestamp`, because we actually want to keep scrolling when someone goes full bore and the gamepad value doesn't update.

meteor/client/ui/Prompter/controller/joycon-device.ts
```diff
@@ -221,28 +228,33 @@ export class JoyConController extends ControllerAbstract {
...
+							lastSeenSpeed = joycon.axes[0] * -1 // in this mode, L is "negative"
+						} else if (joycon.mode === 'R') {
+							lastSeenSpeed = joycon.axes[0] * 1.4 // in this mode, R is "positive"
+							// factor increased by 1.4 to account for the R joystick being less sensitive than L
+						}
+						this.timestampOfLastUsedJoyconInput = joycon.timestamp
+					}
+				} else if (joycon.mode === 'LR') {
+					// L + R mode
+					// get the first one that is moving outside of the deadband, prioritizing the L controller
```
jesperstarkar, line 250: There's still the tiiiiiiniest small edgecas:

When, and only when, the R controller is stuck and gets disconnected. If giving the L some input, it takes over as expected, but when releasing the L again, the R will continue scrolling. This comes from both sides having their timestamp updated at the same time when paired, which they appear to be for a second or two before going to the single L config.

This can only happen when going from LR to L, with R stuck.

This would not have happened with a > comparison at :232, but tests shows that the UX of > is broken, as @jstarpl expected when going on a steady pace.

jesperstarkar: I think this is the best we both could and should do.

jstarpl: Well, I can't say that I experienced it: I just predicted that it would happen, based on the algorithm we had.

jstarpl (approved):

PR #749, opened 2022-07-28
fix: make `isolatedAutorunWithCleanup` independent from `currentComputation`

ianshade:

Bugfix regarding Action Triggers

MountedAdLibTriggers are lost when ShowStyle config is updated (and in other scenarios causing the tracker in RundownView to rerun)

* **Other information**:

In a scenario when `isolatedAutorunWithCleanup` is called within an effect, the `currentComputation` might be stopped and replaced, triggering the cleanup, but the effect might not rerun due to the dependencies being identical. Manually stop trackers made with `isolatedAutorunWithCleanup` when no longer needed.

**Status**


jstarpl (approved):
Looks good!

PR #750, opened 2022-08-03
feat: syncIngestUpdateToPartInstance is called when a Part is orphaned, and allows for removing the PartInstance

Julusian:

feature

The blueprint method `shouldRemoveOrphanedPartInstance` has been combined into `syncIngestUpdateToPartInstance`.
To achieve this, a new method has been added to the context `removePartInstance()`. It is only valid to call this when `playStatus: 'next'`, at any other time it will throw.

This was to let `syncIngestUpdateToPartInstance` decide to remove a PartInstance when the Part has become invalid. That case didn't fit under the `shouldRemoveOrphanedPartInstance` concept.
Once it was able to do this, it felt like it made sense to merge `shouldRemoveOrphanedPartInstance` into it too, so that there was a single blueprint call that would cover all cases.

This does require some blueprints updates, as `syncIngestUpdateToPartInstance` is called with `undefined` passed as the Part for any orphaned PartInstances.
Additionally, for any orphaned PartInstances, core was previously propogating changes to `onEnd` infinites. This is no longer done, and should be handled inside of `syncIngestUpdateToPartInstance` if that behaviour is desired.

* **Other information**:

**Status**


packages/job-worker/src/blueprints/context/lib.ts
```diff
@@ -121,10 +121,20 @@ function convertPieceInstanceToBlueprintsInner(pieceInstance: PieceInstance): Co
 	return obj
 }
 
+/**
+ * Convert a PieceInstance into IBlueprintPieceInstance, for passing into the blueprints
+ * @param pieceInstance the PieceInstance to convert
+ * @returns a cloned complete and clean IBlueprintPieceInstance
+ */
 export function convertPieceInstanceToBlueprints(pieceInstance: PieceInstance): IBlueprintPieceInstance {
 	return convertPieceInstanceToBlueprintsInner(pieceInstance)
```
LindvedKrvang, line 130: I have seen this naming convention throughout Sofie where `Inner` is postfixed to method names.
I'm curious: What is the purpose/benefit of this?

Julusian: I typically use it when the method is the same as the non `Inner` version, but takes more localised input parameters.
In this case, its because I want the return type of `convertPieceInstanceToBlueprintsInner` to be `Complete<IBlueprintPieceInstance>`, as the method is also used by `convertResolvedPieceInstanceToBlueprints`, but I want callers outside of the file to be given `IBlueprintPieceInstance`.
`Inner` may not be the best description, but I dont know what would be better. `EnsuringComplete`? without understanding the custom return type that isnt any clearer

packages/blueprints-integration/src/api.ts
```diff
@@ -150,6 +149,8 @@ export interface ShowStyleBlueprintManifest extends BlueprintManifestBase {
 	 * `playStatus: previous` means that the currentPartInstance is `orphaned: adlib-part`
 	 * and thus possibly depends on an already past PartInstance for some of it's properties. Therefore
 	 * the blueprint is allowed to modify the most recently played non-adlibbed PartInstance using ingested data.
+	 *
+	 * `newData: undefined` happens when the PartInstance is orphaned
```
nytamin, line 153: Is this comment correct? The types say that `newData.part` is `| undefined`

packages/job-worker/src/ingest/syncChangesToPartInstance.ts
```diff
@@ -28,7 +28,7 @@ import {
 	convertPartToBlueprints,
 	convertPieceInstanceToBlueprints,
 } from '../blueprints/context/lib'
-import { saveIntoCache } from '../cache/lib'
+// import { saveIntoCache } from '../cache/lib'
```
nytamin, line 31: This commented our code should be removed :)

nytamin (approved):
This looks good to me!

PR #751, opened 2022-08-10
chore: move SettingsMenu to separate file

Julusian:

feature

The settings ui requires a lot of scrolling to navigate.
There are a lot of sections to it which can become very long when configured that require navigating past.

This made using the settings really tedious, especially when you want a section further down and meteor reloads the page due to some changes.

The studio and showstyle settings have been split into a tab/sub-page per section:
Screenshot from 2022-08-10 15-24-21

In the sidebar, the headings are now click to expand, so take up no more space unless either manually expanded or you navigate to a page inside the collapsible group.

The sidebar css isnt the best, but it works and doesnt look too terrible. 

* **Other information**:

For reviewing, the first few commits are purely splitting up large files to make it easier to navigate. They can be ignored

**Status**


meteor/client/ui/Settings/SettingsMenu.tsx
```diff
@@ -0,0 +1,511 @@
...
+			</NavLink>
+			{expanded
+				? links.map((link, i) => (
+						<NavLink
+							key={i}
+							activeClassName="selectable-selected"
+							className="settings-menu__settings-menu-item selectable clickable"
+							to={`${basePath}/${link.subPath}`}
+						>
+							<div className="selectable clickable">
```
jstarpl, line 207: Is this div useful for anything? The parent is already "selectable clickable", so this doesn't seem to be neccessary.

meteor/client/ui/Settings/SettingsMenu.tsx
```diff
@@ -0,0 +1,511 @@
...
+		// Auto-expand if the current location starts with the basePath
+		if (location.pathname.startsWith(basePath)) {
+			setExpanded(true)
+		}
+	}, [location.pathname, basePath])
+
+	return (
+		<>
+			<NavLink
+				activeClassName="selectable-selected"
```
jstarpl, line 192: This doesn't seem like good UX. I feel only one item in the list should be marked as selected. Also, I feel like there should be some sort of a marker that clicking on this header is going to expand it.

meteor/client/styles/responsiveOverrides.scss
```diff
@@ -1,11 +1,11 @@
 @import '_variables';
 
-.gutter.has-statusbar {
-	margin-right: $statusbar-width;
-}
+// .gutter.has-statusbar {
```
jstarpl, line 3: I don't think there's a point in keeping these as comments.

jstarpl (changes requested):
I'm not very happy with the UX of having two elements in that list have a selected state. I also would like to insert some sort of a collapse chevron to visualize the collapsing/expanding mechanism.

Julusian:
image

jstarpl (approved):
Yeah, I think this looks good!

PR #752, opened 2022-08-11
fix: move setup of process and logging of Gateways to server-core-integration

nytamin:
Bug fix

There are issues in mos-gateway which causes som logging to not work.
Examples:
```
const err = new Error('Error')
logger.error(err)
logger.error('here's my error', err, err.stack)
logger.info('these', 'are', 'many', 'arguments')
```

Because MOS-gateway and Playout-gateway have duplicates of the code setting up logging, they don't always behave the same.

The examples above work now, and works the same between MOS-gw and Playout-gw.

* **Other information**:

**Status**

### DIFF
--- a/packages/server-core-integration/package.json
+++ b/packages/server-core-integration/package.json
@@ -77,7 +77,7 @@
 		"got": "^11.8.5",
 		"tslib": "^2.4.0",
 		"underscore": "^1.13.4",
-		"winston": "^3.5.1"
+		"winston": "^3.8.2"
 	},
 	"lint-staged": {
 		"*.{js,css,json,md,scss}": [
--- a/packages/server-core-integration/package.json
+++ b/packages/server-core-integration/package.json
@@ -76,7 +76,8 @@
 		"faye-websocket": "^0.11.4",
 		"got": "^11.8.5",
 		"tslib": "^2.4.0",
-		"underscore": "^1.13.4"
+		"underscore": "^1.13.4",
+		"winston": "^3.5.1"
 	},
 	"lint-staged": {
 		"*.{js,css,json,md,scss}": [
--- a/packages/server-core-integration/src/lib/gatewaySetup.ts
+++ b/packages/server-core-integration/src/lib/gatewaySetup.ts
@@ -1,0 +1,223 @@
+import * as Winston from 'winston'
+
+/**
+ * Used when JSON.stringifying values that might be circular
+ * Usage: JSON.stringify(value, JSONStringifyCircular()))
+ */
+function JSONStringifyCircular() {
+	const cacheValues: any[] = []
+	const cacheKeys: any[] = []
+	const stringifyFixer = (key: string, value: any) => {
+		if (typeof value === 'object' && value !== null) {
+			const i = cacheValues.indexOf(value)
+			if (i !== -1) {
+				// Duplicate reference found
+				try {
+					// If this value does not reference a parent it can be deduped
+					return JSON.parse(JSON.stringify(value))
+				} catch (error) {
+					// discard key if value cannot be deduped
+					return '[circular of ' + (cacheKeys[i] || '*root*') + ']'
+				}
+			}
+			// Store value in our collection
+			cacheValues.push(value)
+			cacheKeys.push(key)
+		}
+		return value
+	}
+	return stringifyFixer
+}
+const { printf } = Winston.format
+const myLogFormat = Winston.format.combine(
+	// Winston.format.errors({ stack: true }),
+	// Winston.format.metadata(),
+	// Winston.format.json()
+	// Winston.format.simple()
+	printf((obj): string => {
+		if (obj instanceof Error) return obj.stack || obj.toString()
+		return JSON.stringify(obj, JSONStringifyCircular())
+	})
+)
+
+/**
+ * Intended to be run first in the Gateways.
+ * Sets up logging, error handling etc.
+ */
+export function setupGatewayProcess(options: { logPath: string; logLevel: string | undefined }) {
+	let logger: Winston.Logger
+	if (options.logPath) {
+		// Log json to file, human-readable to console
+		const transportConsole = new Winston.transports.Console({
+			level: options.logLevel || 'silly',
+			handleExceptions: true,
+			handleRejections: true,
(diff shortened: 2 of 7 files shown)


jesperstarkar:
Tested. Works.

packages/server-core-integration/src/lib/gatewaySetup.ts
```diff
@@ -0,0 +1,215 @@
...
+		}
+	}
+
+	return {
+		logger,
+	}
+}
+
+/** Do various loggings, to test that logging works as intended */
+export function testLogging(logger: Winston.Logger) {
```
LindvedKrvang, line 157: This seems like something that belongs in a test file

packages/server-core-integration/src/lib/gatewaySetup.ts
```diff
@@ -0,0 +1,215 @@
+import * as Winston from 'winston'
+
+/**
+ * Used when JSON.stringifying values that might be circular
+ * Usage: JSON.stringify(value, JSONStringifyCircular()))
+ */
+function JSONStringifyCircular() {
```
LindvedKrvang, line 7: To continue our talk from Tuesday.
A small thing one can do to improve the readability of the code a lot is to declare the return type so you don't have to read the implementation to figure it out.

This scenario highlights that since `JSONStringify` is a popuplar method and returns a String. Then when I see a method named `JSONStringifyCircular` I would also expect that that also returns a String, since the name of the method expands on the name of `JSONStringify`. Instead it returns `(key: string, value: any) => any`, which is vastly different from the return type of `JSONStringify`.

It seems small, but it means that a maintenance developer doesn't have to spend time reading the implementation of `JSONStringifyCircular` when said developer is trying to understand another function that uses `JSONStringifyCircular`.
And not having to go read the implementations of other methods all the time to figure the return type is a big time saver.

packages/server-core-integration/src/lib/gatewaySetup.ts
```diff
@@ -0,0 +1,215 @@
...
+	})
+)
+
+/**
+ * Intended to be run first in the Gateways.
+ * Sets up logging, error handling etc.
+ */
+export function setupGatewayProcess(options: { logPath: string; logLevel: string | undefined }) {
+	let logger: Winston.Logger
+	if (options.logPath) {
```
LindvedKrvang, line 49: To continue our talk from Tuesday about trying to make smaller functions.
This if-statement presents a good opportunity for that.

As far as I can see the body of the if- and else-statment only uses the variable `logger` and the parameter `options` and the purpose of the bodies is to create the logger. So you could create two methods (one for each of the bodies) that each creates a logger:

```
export function setupGatewayProcess(options: {...}) {
  const logger = options.logPath 
            ? createLoggerWithPath(options.logLevel, options.logPath) 
            : createLogger(options.logLevel)
  // Rest of your function
}

function createLoggerWithPath(...) {
  // The body of your if-statement
}

function createLogger(...) {
  // The body of your else statement
} 
```

Now, as a maintenance developer, I don't have to first read through the code of creating the logger, if my current task has nothing to do with the creation of the logger.

Disclaimer: My naming of the methods might not be accurate. I haven't spent the time to understand the code 100%

packages/server-core-integration/src/lib/gatewaySetup.ts
baltedewit: We should consider enabling [@typescript-eslint/explicit-function-return-type](https://github.com/typescript-eslint/typescript-eslint/blob/v2.34.0/packages/eslint-plugin/docs/rules/explicit-function-return-type.md), afaIk it's already part of our code standard

baltedewit (approved):
It should no longer be necessary to keep winston in the package.json's of the gateways now, right?

Since this is mostly just copying around functionality I will forgo a full review of the copied code.

packages/server-core-integration/src/lib/gatewaySetup.ts
LindvedKrvang: I would be for that :)

I would also suggest that we start adding the types to our variables more.
Offcourse if it's **obvious** what the type is, then it's not that necessary (but still nice).
So in a case like `const amountOfTries = 1` it's obvious it has the type number, so we can leave it out, but in a case like `const piece = fetchPiece()` it's not obvious what the type is (and what attributes the piece has, so it's nice to have the interface by hand).

packages/server-core-integration/package.json
```diff
@@ -74,6 +74,7 @@
 		"faye-websocket": "^0.11.4",
 		"got": "^11.8.5",
 		"tslib": "^2.3.1",
+		"winston": "^3.5.1",
```
Julusian, line 77: `winston` should be removed from the package.json of mos-gateway and playout-gateway with it defined here, or we will end up pulling in multiple versions

packages/server-core-integration/src/lib/gatewaySetup.ts
Julusian: I agree that I would expect `JSONStringifyCircular` to return a string, and not be the parameter to pass to `JSON.stringify`.

packages/server-core-integration/src/lib/gatewaySetup.ts
```diff
@@ -0,0 +1,215 @@
...
+
+	// Because the default NodeJS-handler sucks and wont display error properly
+	process.on('warning', (e: any) => {
+		logger.warn('Unhandled warning, see below')
+		logger.error('error', e)
+		logger.error('error.reason', e.reason || e.message)
+		logger.error('error.stack', e.stack)
+	})
+	// Override the logger methods, to handle a few common cases:
+	for (const method of ['error', 'warn', 'help', 'data', 'info', 'debug', 'prompt', 'http', 'verbose', 'input', 'silly']) {
```
Julusian, line 128: What is this block trying to solve? It looks new, and I dont know why we need to apply such a complex hack to these methods.
Is this something we could/should be doing as part of a winston formatter rather than overriding methods?

Julusian (changes requested):

nytamin:
Closing this PR due to inactivity. We'll likely do another try at unifying this logging at a later time.

PR #754, opened 2022-08-25
Chore/update dockerimage

siljekristensen:

- Updated node version from 14.19 to 14.20. 
- Changed the deploy image from using a Debian based image (10.12) to a Alpine (3.16.2) based image. This takes us from `Total: 136 (UNKNOWN: 0, LOW: 84, MEDIUM: 12, HIGH: 30, CRITICAL: 10)` known vulnerabilities in that layer of the image to `Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)`

* **Other information**:

**Status**

### DIFF
--- a/meteor/Dockerfile
+++ b/meteor/Dockerfile
@@ -13,21 +13,24 @@ RUN yarn install && yarn build
 COPY meteor /opt/core/meteor
 COPY scripts /opt/core/scripts
 WORKDIR /opt/core/meteor
+
 # Force meteor to setup the runtime
 RUN meteor --version --allow-superuser
 RUN meteor npm install -g yarn
 RUN meteor yarn install
+
 # Restore the NODE_ENV variable:
 ENV NODE_ENV $NODE_ENV_TMP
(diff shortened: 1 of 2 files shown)


Julusian (approved):
I think that we needed `-slim` in the past because in the first stage we were compiling a native binding. But as that dependency has been removed I see no reason to keep using `-slim`.

At some later time, we should look at whether we need first stage to be the full image or whether it would work with alpine too

jstarpl:
@Julusian @siljekristensen Should we also modify `Dockerfile.circle` in the same fashion? I think that's what being used to push public images to DockerHub? No reason to have those two be that different.

siljekristensen:
> I think that we needed `-slim` in the past because in the first stage we were compiling a native binding. But as that dependency has been removed I see no reason to keep using `-slim`.
> 
> At some later time, we should look at whether we need first stage to be the full image or whether it would work with alpine too

I tested building with Alpine too, but that didn't seem to work with Meteor (because it needs glibc). But I think building in a "larger" image and then just copying what is needed to a smaller production image should be fine.

siljekristensen:
> @Julusian @siljekristensen Should we also modify `Dockerfile.circle` in the same fashion? I think that's what being used to push public images to DockerHub? No reason to have those two be that different.

Ah, I didn't realize how it was being used. I can update this PR.

PR #756, opened 2022-08-30
feat: Input gateway support

jstarpl:

This is a feature

This PR aims at implementing an API for trigger gateway (input gateway) peripheral devices. This will allow for such things as native Stream Deck plugin integration, HTTP, MIDI, GPI, Serial triggers, etc.

The plan for this implementation is as follows:
* Add a Mounted Triggers publications for peripheralDevices, so that devices can know what they will trigger
* Add core peripheralDevice API methods for sending triggers into Core. The trigger payload will be a key-value object.
* Add a Last Trigger publication so that the UI can pick that up when assigning triggers to actions

This will allow us to listen for triggers in very much the same way we do for hotkeys, i.e., the Input Gateway will not need to present a manifest of possible triggers, but we should be able to expand on this API later, if needed (say, have an X-Y-Z joystick control into a DVE AdLib Action?)

* **Other information**:

**Status**

### DIFF
--- a/packages/documentation/docs/user-guide/configuration/settings-view.md
+++ b/packages/documentation/docs/user-guide/configuration/settings-view.md
@@ -109,10 +109,14 @@ This is a way to set up how - outside of the Point-and-Click Graphical User Inte
 
 The triggers are designed to be either client-specific or issued by a peripheral device module.
 
-Currently, the Action Triggers system only supports a single, client-specific trigger type: a Hotkey. Hotkeys can be either a single key, a combination of keys (*combo*) or a *chord* - a sequnece of key combinations pressed in a particular order. *Chords* are popular in some text editing applications and vastly expand the amount of actions that can be triggered from a keyboard, at the expense of the time needed to execute them. Currently, the Hotkey editor in Sofie does not support creating *Chords*, but they can be specified by Blueprints during migrations.
+Currently, the Action Triggers system supports setting up two types of triggeers: Hotkeys and Device Triggers. 
 
-To edit a given trigger, click on the trigger pill on the left of the Trigger-Action set. When hovering, a **+** sign will appear,
-allowing you to add a new trigger to the set.
+Hotkeys are valid in the scope of a browser window and can be either a single key, a combination of keys (*combo*) or a *chord* - a sequnece of key combinations pressed in a particular order. *Chords* are popular in some text editing applications and vastly expand the amount of actions that can be triggered from a keyboard, at the expense of the time needed to execute them. Currently, the Hotkey editor in Sofie does not support creating *Chords*, but they can be specified by Blueprints during migrations.
+
+To edit a given trigger, click on the trigger pill on the left of the Trigger-Action set. When hovering, a **+** sign will appear, allowing you to add a new trigger to the set.
+
+Device Triggers are valid in the scope of a Studio and will be evaluated on the currently active Rundown in a given studio. To use Device Triggers, you need to have at least a single [Input Gateway](../installation/installing-input-gateway) attached to a Studio and a Device configured in the Input Gateway. Once that's done, when selecting a **Device** trigger type in the pop-up, you can invoke triggers on your Input Device and you will see a preview of the input
+events shown at the bottom of the pop-up. You can select which of these events should be the trigger by clicking on one of the previews.
 
 #### Actions
 
--- a/packages/documentation/docs/user-guide/installation/installing-input-gateway.md
+++ b/packages/documentation/docs/user-guide/installation/installing-input-gateway.md
@@ -1,0 +1,24 @@
+# Input Gateway
+
+The Input Gateway handles control devices that are not capable of running a Web Browser. This allows Sofie to integrate directly with devices such as: Hardware Panels, GPI input, MIDI devices and external systems being able to send an HTTP Request.
+
+To install it, begin by downloading the latest release of [Input Gateway from GitHub](https://github.com/nrkno/sofie-input-gateway/releases). You can now run the `input-gateway.exe` file inside the extracted folder. A warning window may popup about the app being unrecognized. You can get around this by selecting _More Info_ and clicking _Run Anyways_. A terminal window will open and begin running the application.
+
+You can now open the _Sofie&nbsp;Core_, `http://localhost:3000`,  and navigate to the _Settings page_. You will see your _Input Gateway_ under the _Devices_ section of the menu. In _Input Devices_ you can add devices that this instance of Input Gateway should handle. Some of the device integrations will allow you to customize the Feedback behavior. The *Device ID* property will identify a given Input Device in the Studio, so this property can be used for fail-over purposes.
+
+## Supported devices and protocols
+
+Currently, input gateway supports:
+
+* Stream Deck panels
+* Skaarhoj panels - _Raw Panel_ mode
+* X-Keys panels
+* MIDI controllers
+* OSC
+* HTTP
+
+### Further Reading
+
+* [Input Gateway Releases on GitHub](https://github.com/nrkno/sofie-input-gateway/releases)
+* [Input Gateway GitHub Page for Developers](https://github.com/nrkno/sofie-input-gateway)
+


meteor/server/publications/mountedTriggers.ts
```diff
@@ -0,0 +1,121 @@
...
+				})
+			)
+		} else {
+			logger.warn(`Pub.mountedTriggersForDevicePreview: Not allowed: "${deviceId}"`)
+		}
+	}
+)
+
+function cursorCustomPublish<T extends { _id: ProtectedString<any> }>(pub: CustomPublish<T>, cursor: Mongo.Cursor<T>) {
+	let buffer: CustomPublishChanges<T> = {
```
Julusian, line 67: How clustered are these updates usually? I am wondering if it would be better to have the buffer be formed as
```
let buffer = {
  added: new Map(),
  changed: new Map(),
  removed: new Set()
}
```
and converted back to `CustomPublishChanges` inside the debounce.

That will avoid needing so much array searching in the observe callbacks. The cost of converting back might be too high to make it worthwhile

Julusian (approved):
This is rather hypocritical of me to say, but I feel like the different classes and things could do with some documentation on their relationships and purposes.
But overall I don't think it was overly hard to follow.

I do think this has a much easier to follow data flow than my recent publications have, but I do worry that this might be overly reactive to changes. Regenerating every item in the collection whenever anything changes feels like a lot of excessive computation, and corresponding garbage collection.
This is something I have tried to avoid in my publications, but equally I might be optimising prematurely at the cost of code readability.

I also see why you were asking about threads. I do wonder if this could end up fighting playout for cpu time. Porting this to use the native mongodb driver, and piping the output back through meteor (or its own websocket server?). Im not sure how some bits of it would work with threading, as it is relying on a global shared state with meteor, and some shared code, but that sounds solvable. Would be a good idea to look into at some point.

baltedewit (approved):
Approaching this review from the backend it really wasn't clear to me at first what the previews were used for. So I would think some more documentation on that would be good.

I also note that the triggers are relying on collections quite heavily, but we don't handle so many user interactions either so I think it will be acceptable.

meteor/server/lib/customPublication/observerGroup.ts
```diff
@@ -0,0 +1,130 @@
...
+	restart(): void
+}
+
+/**
+ * Helper to trigger reactivity inside of an OptimisedObserver whenever one of the other Mongo observers changes
+ * Note: care needs to be taken when using this, as Mongo observers call `added` for every document when they start. It is very easy to form an infinite loop of observer invalidations
+ * @param generator Function to generate the `Meteor.LiveQueryHandle`s
+ * @returns Handle to stop and restart the observer group
+ */
+export function ReactiveMongoObserverGroup(
```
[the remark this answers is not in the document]
Julusian: You can blame me for this one, this came from my not-yet opened PR which this was based on.

Yeah it probably could be a class, I think the style of writing here was rather influenced by writing react functions. But I feel like making it a class will cause it to be clunkier to use once it has to stop using fibers (as it will need to return a Promise), which was incredibly lazy of me to write it with fibers...

meteor/server/publications/mountedTriggers.ts
jstarpl: They will be generally very clustered. I think it is probably a good idea to convert these to Sets & Maps.

meteor/server/api/deviceTriggers/RundownsObserver.ts
```diff
@@ -0,0 +1,66 @@
...
+			}
+		)
+		this.#rundownsLiveQuery = cursor.observe({
+			added: (doc) => {
+				this.#rundownIds.add(doc._id)
+				this.updateRundownContent()
+			},
+			changed: (doc) => {
+				this.#rundownIds.add(doc._id)
+				this.updateRundownContent()
```
Julusian, line 35: I don't think this `changed` will ever be fired, I think I checked and a document starting/stopping matching the query is an added/removed.  
So this will only catch the change of the `projection` fields. Which in this case is nothing (_id is immutable)

But if it did fire, should it be blindly calling `this.updateRundownContent()`, or only when it added something to this.#rundownIds?

meteor/server/api/deviceTriggers/ReactiveCacheCollection.ts
```diff
@@ -0,0 +1,124 @@
...
+		return result
+	}
+
+	link(): ObserveCallbacks<Document> {
+		return {
+			added: (doc: Document) => {
+				this.upsert(doc._id, { $set: omit(doc, '_id') as Partial<Document> })
+			},
+			changed: (doc: Document) => {
+				this.upsert(doc._id, { $set: omit(doc, '_id') as Partial<Document> })
```
Julusian, line 113: I think this will be 'leaking' data, as the `upsert` won't remove properties that have become undefined. This could have unintentional side effects

meteor/server/api/deviceTriggers/StudioDeviceTriggerManager.ts
```diff
@@ -0,0 +1,227 @@
...
+
+	public get showStyleBaseId(): ShowStyleBaseId | null {
+		return this.#showStyleBaseId
+	}
+
+	public set showStyleBaseId(value: ShowStyleBaseId | null) {
+		this.#showStyleBaseId = value
+	}
+
+	updateTriggers(cache: ContentCache | null): void {
```
nytamin, line 32: minor note: 
this method could perhaps be split into two, for clarity?
* `updateTriggers`
* `clearTriggers` (using the code path where the cache is `null`)
I noticed that it is only ever called with a cache OR with `null`

meteor/server/api/peripheralDevice.ts
```diff
@@ -1150,5 +1152,16 @@ class ServerPeripheralDeviceAPIClass extends MethodContextAPI implements NewPeri
 	) {
 		await PackageManagerIntegration.removePackageInfo(this, deviceId, deviceToken, type, packageId)
 	}
+	// --- Triggers ---
+	async inputDeviceTrigger(
```
nytamin, line 1156: As a first-time-reader, this is not immediately clear when or why it is called.
Perhaps add a description here?
```
/**
This method is called when a button is pressed, a knob is turned etc..
@param value will be a boolean when a button is pressed (?), a number when a knob is turned (?)...

 */
```

(I know this is not documented on any of the other methods, but still....)

meteor/server/migration/currentSystemVersion.ts
```diff
@@ -49,4 +49,4 @@
  */
 
 // Note: Only set this to release versions, (ie X.Y.Z), not pre-releases (ie X.Y.Z-0-pre-release)
-export const CURRENT_SYSTEM_VERSION = '1.46.0'
+export const CURRENT_SYSTEM_VERSION = '1.47.0'
```
nytamin, line 52: TODO: ensure that this diff goes away during merge (or when updating the branch with the latest release-branch)

nytamin (approved):

PeterC89 (approved):

meteor/server/api/deviceTriggers/Queue.ts
```diff
@@ -0,0 +1,80 @@
...
+
+type AddOptions = {
+	className?: string
+}
+
+type Options = {
+	autoStart: boolean
+}
+
+export class PromiseQueue {
```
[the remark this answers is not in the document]
jstarpl: Is shared-lib published? My understanding was that it's shared package for the core mono-repo, but other than that, it's technically private. I see that it's got `private: false` set in `package.json`, but I don't see it being published to NPM. I don't think it makes sense to publish it?

meteor/server/publications/deviceTriggersPreview.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+async function setupDeviceTriggersPreviewsObservers(
+	args: ReadonlyDeep<DeviceTriggersPreviewArgs>,
+	triggerUpdate: TriggerUpdate<DeviceTriggersUpdateProps>
+): Promise<Meteor.LiveQueryHandle[]> {
+	const studioId = unprotectString(args.studioId)
+	const lastTriggersStudio = prepareBufferForStudio(studioId)
+
+	lastTriggersStudio.updated = () => {
```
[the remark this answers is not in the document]
jstarpl: I agree, but I don't see any other solution. Ultimately, this needs to be a in-memory collection. The alternative would be to use an in-memory minimongo collection, but I don't think that improves much.

meteor/server/api/peripheralDevice.ts
jstarpl: I agree, but should this be documented here (where this is an implementation of an interface) or should it be in `@sofie-automation\shared-lib\src\peripheralDevice\methodsAPI.ts`, where this documentation will be available to consumers? I think I'm going to go for the latter.

meteor/server/api/deviceTriggers/ReactiveCacheCollection.ts
```diff
@@ -0,0 +1,124 @@
+import { omit } from '@sofie-automation/corelib/dist/lib'
+import { ProtectedString } from '@sofie-automation/corelib/dist/protectedString'
+import { Mongo } from 'meteor/mongo'
+import { ObserveCallbacks } from '../../../lib/collections/lib'
+
+type Reaction = () => void
+
+export class ReactiveCacheCollection<
+	Document extends { _id: ProtectedString<any> }
+> extends Mongo.Collection<Document> {
```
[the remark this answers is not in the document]
jstarpl: The idea behind extending `Mongo.Collection` is to be able to use these collections in isomorphic code in `lib/api/triggers/actionFactory` and `lib/api/triggers/actionFilterChainCompilers` where they could be injected and used in place of the regular Mini-Mongo collections on the server side, thus saving a round-trip to the database whenever an action is triggered. Replacing this with `DbCacheWriteCollection` would mean that either the Client would now have to maintain a copy of the MiniMongo collection as a `DBCache*Collection` or would mean having two implementations of that code, utilizing different collection classes - one for the Client and another one for the Server side.

meteor/server/api/deviceTriggers/StudioDeviceTriggerManager.ts
```diff
@@ -0,0 +1,227 @@
...
+				const actionId = protectString(`${studioId}_${triggeredAction._id}_${key}`)
+				GlobalTriggerManager.setAction(actionId, compiledAction)
+
+				Object.entries(triggeredAction.triggers).forEach(([key, trigger]) => {
+					if (!isDeviceTrigger(trigger)) {
+						return
+					}
+
+					const deviceTriggerMountedActionId = protectString(`${actionId}_${key}`)
+					DeviceTriggerMountedActions.upsert(deviceTriggerMountedActionId, {
```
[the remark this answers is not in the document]
jstarpl: I guess, this is `replace`, however using something like replace would mean moving DbWriteCacheCollection from `job-worker` into something like `shared-lib` and then relinking everything. As such, I don't think the value of making this specifically `replace` warrants expanding the breadth of this PR even more.

meteor/server/api/deviceTriggers/Queue.ts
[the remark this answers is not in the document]
jstarpl: Moved to shared-lib.

meteor/server/api/deviceTriggers/ReactiveCacheCollection.ts
jstarpl: Fixed by iterating over the `fields` in `observeChanges` and finding the undefined values.

PR #758, opened 2022-09-01
fix: stoppedPlayback is sometimes incorrectly reported for PartInstances

jstarpl:

Bug fix

The `stoppedPlayback` timestamp can be incorrectly reported by the playout gateway, due to Core sending multiples of the same timeline.

The `published` property on the RoutedTimeline is removed, so that subsequent republishings of the same timeline through Timeline fast-track do not trigger a timeline update in playout-gateway. This is especially important when using the 'now' feature.

* **Other information**:

**Status**

### DIFF
--- a/packages/job-worker/src/playout/playout.ts
+++ b/packages/job-worker/src/playout/playout.ts
@@ -783,8 +783,8 @@ async function _onPartPlaybackStarted(
 		throw new Error(`PartInstance "${data.partInstanceId}" in RundownPlayst "${data.playlistId}" not found!`)
 
 	// make sure we don't run multiple times, even if TSR calls us multiple times
-	const isPlaying = !!playingPartInstance.timings?.startedPlayback
-	if (!isPlaying) {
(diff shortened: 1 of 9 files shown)


nytamin (approved):

packages/job-worker/src/playout/playout.ts
```diff
@@ -783,7 +783,7 @@ async function _onPartPlaybackStarted(
 		throw new Error(`PartInstance "${data.partInstanceId}" in RundownPlayst "${data.playlistId}" not found!`)
 
 	// make sure we don't run multiple times, even if TSR calls us multiple times
-	const isPlaying = playingPartInstance.timings?.startedPlayback && !playingPartInstance.timings?.stoppedPlayback
+	const isPlaying = !!playingPartInstance.timings?.startedPlayback
```
Julusian, line 786: this should probably be renamed to `hasStartedPlaying`, as it could have started and already stopped

Julusian (approved):

PR #762, opened 2022-09-21
chore: remove mongo-like queries from CacheCollection

Julusian:

code quality

Our CacheCollection and CacheObject types in the job-worker have long supported using mongo-like queries to find or mutate documents. This is using a custom implementation of those queries, and is not very efficient.

For a while we have supported using functions instead of the mongo-like queries, this PR changes it so that these functions are the only way to do these collection operations. This forces us to write cleaner and more typesafe code, and avoids the complexity of the query language.

This allows us to review and make some further optimisations, such as allowing us to more easily skip modifying a single document when doing a bulk update, and avoids needing to do multiple iterations when doing a bulk update.

In the rare scenario where a mongo-like query is desirable, the helpers we were using do still exist and can be invoked manually

* **Other information**:

There is a chance that some queries were broken here, as we have occasionally typed some of the modifiers as `any`, meaning typescript hasnt flagged them as incompatible. I think I caught them all, but it is possible some have sipped through

**Status**


packages/job-worker/src/cache/CacheCollection.ts
```diff
@@ -69,42 +63,13 @@ export class DbCacheReadCollection<TDoc extends { _id: ProtectedString<any> }> {
 		return this._collection.name
 	}
 
-	findFetch(selector: MongoQuery<TDoc> | TDoc['_id'] | SelectorFunction<TDoc>, options?: FindOptions<TDoc>): TDoc[] {
+	findFetch(selector: SelectorFunction<TDoc> | null, options?: FindOptions<TDoc>): TDoc[] {
```
jstarpl, line 66: I wonder, if we're changing the signatures this much, and are actively going away from pretending to be Mongo, should this be called just `find` or `findAll`, so that no suggestion of this being a Meteor Cursor or a Mongo query is made?

jstarpl: Also, I suppose the meaning of `null` should be clarified through some JSDoc, so that it can be intellisensed where it's being used.

packages/job-worker/src/cache/CacheCollection.ts
```diff
@@ -116,11 +81,20 @@ export class DbCacheReadCollection<TDoc extends { _id: ProtectedString<any> }> {
 		if (span) span.end()
 		return res
 	}
-	findOne(
-		selector: MongoQuery<TDoc> | TDoc['_id'] | SelectorFunction<TDoc>,
-		options?: FindOneOptions<TDoc>
-	): TDoc | undefined {
-		return this.findFetch(selector, options)[0]
+	findOne(selector: TDoc['_id'] | SelectorFunction<TDoc>, options?: FindOneOptions<TDoc>): TDoc | undefined {
```
jstarpl, line 84: Selector should probably allow `null` here as well, or is this intentional?

jstarpl (approved):
Feels like a good change, but I think maybe `findFetch` should be renamed, to reflect it no longer has anything to do with Mongo queries? Also, some JSDoc for findFetch and findOne would be good too?

packages/job-worker/src/cache/CacheCollection.ts
Julusian: yeah `findAll` would make sense
docs probably do need some work..

I was wondering about changing the structure of the options parameter too, as some bits it can do really arent that useful..
eg, we dont clone the objects except if a `fields` property is defined. and as there is no reactivity, we dont need to use it to limit that. so `fields` adds a cost than an optimization here (the opposite normal mongo queries).
And i think the sorting algorithm is fairly primitive/inefficient
But I'm leaving that for a follow up

packages/job-worker/src/cache/CacheCollection.ts
Julusian: This was intentional. `null` would pick the 'first' document in the map, which would be a rather random and potentially unstable result. Outside of unit tests I dont see it as having any value

PR #764, opened 2022-09-26
Use node.js-script instead of requiring powershell

nytamin:
**The problem**
currently, the script
```
"sync-version": "yarn set-version $(node -p \"require('../meteor/package.json').version\")",
```
requires yarn to use powerwhell on windows, requiring us to use the hack
```
yarn config set script-shell (Get-Command powershell.exe).Definition
or
yarn config set script-shell (Get-Command pwsh.exe).Definition
```
But this breaks other commands, so we need to run
```
yarn config delete script-shell
```
afterwards.

By instead running these type of scripts in a node.js-script right away, they'll work without any such shenanigans.


packages/package.json
```diff
@@ -20,11 +20,10 @@
 		"build": "yarn lerna run build",
 		"build:try": "yarn lerna run --no-bail build",
 		"watch": "lerna run --parallel build:main -- --watch --preserveWatchOutput",
-		"sync-version-and-changelog": "lerna version --exact --force-publish --conventional-commits --no-git-tag-version --no-push --yes $(node -p \"require('../meteor/package.json').version\")",
-		"postsync-version-and-changelog": "yarn stage-versions",
 		"sync-version": "yarn set-version $(node -p \"require('../meteor/package.json').version\")",
```
Julusian, line 23: Is this `sync-version` command used anywhere else still?

Julusian (approved):

packages/package.json
nytamin: no it's not, I thought I removed it but must have messed up :)

PR #766, opened 2022-09-29
fix(Playout Gateway): remove uninitialized devices from the list

jstarpl:

Bug fix

If a device fails to initialize correctly in a `_updateDevices` pass due to a crash or a timeout, it still stays in TSR's "devices" list. Because there is no change in device settings, the device is never retried to initialize, causing it to be potentially stuck forever, even if it could initialize later correctly (say, the hardware device comes back online) and the overall Peripheral Device status cannot be properly evaluated.

Remove them from TSR if they failed to initialize and retry initializing them in `_updateDevices`.

* **Other information**:

Requires https://github.com/nrkno/sofie-timeline-state-resolver/pull/220 to be merged

**Status**

### DIFF
--- a/packages/playout-gateway/src/tsrHandler.ts
+++ b/packages/playout-gateway/src/tsrHandler.ts
@@ -637,6 +637,9 @@ export class TSRHandler {
 					}
 
 					Promise.all(
+						// At this point in time, promiseOperations contains the promises that have timed out.
+						// If we tried to add or re-add a device, that apparently failed so we should remove the device in order to
+						// give it another chance next time _updateDevices() is called.
 						Object.values(promiseOperations)
 							.filter((op) => op.operation === 'add' || op.operation === 're-add')
 							.map(async (op) =>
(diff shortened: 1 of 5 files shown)


packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -621,15 +627,43 @@ export class TSRHandler {
...
+							})
+						})
+						.catch((e) => {
+							this.logger.error(
+								`Error when trying to remove unsuccessfully initialized devices: ${stringifyIds(
+									Object.values(promiseOperations).map((op) => op.deviceId)
+								)}`,
+								e
+							)
+							resolve({
```
nytamin, line 661: it would be slightly more DRY to call resolve({..}) in a `.finally()` instead

packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -621,15 +627,43 @@ export class TSRHandler {
...
+								.join(',')}`
+						)
 					}
-					resolve({
-						success: false,
-						reason: 'timeout',
-						details: keys,
-					})
+
+					Promise.all(
```
nytamin, line 639: While we're here, it would be nice to add a comment or two describing what we're doing.
```
// At this point in time, promiseOperations contains the promises that have timed out.
// If we tried to add or re-add a device, that apparently failed so we should remove the device in order to
// give it another chance next time _updateDevices() is called.
```

nytamin (approved):

packages/playout-gateway/src/tsrHandler.ts
jstarpl: Yeah, good point about explaining the intent here.

PR #767, opened 2022-09-29
Feat: timeline datastore

baltedewit:

* Implements the Datastore API from the timeline-state-resolver (https://github.com/nrkno/sofie-timeline-state-resolver/pull/219)
* Implements a new TimelinedataStore collection
* Implements a cleanup method and a debug view for said collection
* Implements a blueprint AdlibAction API for the datastore collection
* Splits the AdlibActionContext into 2 such that the blueprints can use a small, fast context for adlib actions utilising only the datastore

**Left todo**

- [x] Rebase on R47
- [ ] A round of reviews with some potential pair programming

**Status**

### DIFF
--- a/meteor/client/ui/TestTools/TimelineDatastore.tsx
+++ b/meteor/client/ui/TestTools/TimelineDatastore.tsx
@@ -1,0 +1,113 @@
+import * as React from 'react'
+import {
+	Translated,
+	translateWithTracker,
+	useSubscription,
+	useTracker,
+} from '../../lib/ReactMeteorData/react-meteor-data'
+import { MeteorReactComponent } from '../../lib/MeteorReactComponent'
+import { StudioSelect } from './StudioSelect'
+import { StudioId } from '../../../lib/collections/Studios'
+import { Mongo } from 'meteor/mongo'
+import { TimelineDatastoreEntry } from '../../../lib/collections/TimelineDatastore'
+import { PubSub } from '../../../lib/api/pubsub'
+
+const TimelineDatastore = new Mongo.Collection<TimelineDatastoreEntry>('timelineDatastore')
+
+interface ITimelineDatastoreViewProps {
+	match?: {
+		params?: {
+			studioId: StudioId
+		}
+	}
+}
+interface ITimelineDatastoreViewState {}
+const TimelineDatastoreView = translateWithTracker<ITimelineDatastoreViewProps, ITimelineDatastoreViewState, {}>(
+	(_props: ITimelineDatastoreViewProps) => {
+		return {}
+	}
+)(
+	class TimelineDatastoreView extends MeteorReactComponent<
+		Translated<ITimelineDatastoreViewProps>,
+		ITimelineDatastoreViewState
+	> {
+		constructor(props: Translated<ITimelineDatastoreViewProps>) {
+			super(props)
+		}
+
+		render() {
+			const { t } = this.props
+
+			return (
+				<div className="mtl gutter">
+					<header className="mvs">
+						<h1>{t('Timeline Datastore')}</h1>
+					</header>
+					<div className="mod mvl">
+						{this.props.match && this.props.match.params && (
+							<div>
+								{/* <ComponentMappingsTable studioId={this.props.match.params.studioId} /> */}
+								<ComponentDatastoreControls studioId={this.props.match.params.studioId} />
+							</div>
+						)}
+					</div>
+				</div>
+			)
+		}
+	}
+)
+
+interface IDatastoreControlsProps {
+	studioId: StudioId
+}
+function ComponentDatastoreControls({ studioId }: IDatastoreControlsProps) {
+	useSubscription(PubSub.timelineDatastore, { studioId })
+
+	const datastore = useTracker(() => TimelineDatastore.find().fetch(), [studioId])
+
+	const createOrEdit = (key: string, value: any) => {
+		const doc = datastore?.find((entry) => entry.studioId === studioId && entry.key === key)
+		if (doc) {
+			TimelineDatastore.update(doc._id, {
+				$set: {
+					value,
+					modified: Date.now(),
+				},
+			})
+		} else {
+			TimelineDatastore.insert({
+				studioId,
+
+				key,
+				value,
+
+				modified: Date.now(),
+			})
+		}
+	}
+
+	return (
+		<div>
+			<div>
+				<button className="btn btn-primary" onMouseDown={() => createOrEdit('input', 1)}>
+					1
+				</button>
+				<button className="btn btn-primary" onMouseDown={() => createOrEdit('input', 2)}>
+					2
+				</button>
+				<button className="btn btn-primary" onMouseDown={() => createOrEdit('input', 3)}>
+					3
+				</button>
+			</div>
+			<div>{JSON.stringify(datastore)}</div>
+		</div>
+	)
+}
+
+class TimelineDatastoreStudioSelect extends React.Component<{}, {}> {
+	render() {
+		return <StudioSelect path="timelinedatastore" title="Timeline Datastore" />
+	}
+}
+
+export { TimelineDatastoreView, TimelineDatastoreStudioSelect }
--- a/meteor/client/ui/TestTools/TimelineDatastore.tsx
+++ b/meteor/client/ui/TestTools/TimelineDatastore.tsx
@@ -1,113 +1,76 @@
 import * as React from 'react'
-import {
-	Translated,
-	translateWithTracker,
-	useSubscription,
-	useTracker,
-} from '../../lib/ReactMeteorData/react-meteor-data'
-import { MeteorReactComponent } from '../../lib/MeteorReactComponent'
+import { useSubscription, useTracker } from '../../lib/ReactMeteorData/react-meteor-data'
 import { StudioSelect } from './StudioSelect'
 import { StudioId } from '../../../lib/collections/Studios'
 import { Mongo } from 'meteor/mongo'
 import { TimelineDatastoreEntry } from '../../../lib/collections/TimelineDatastore'
 import { PubSub } from '../../../lib/api/pubsub'
+import { protectString, unprotectString } from '@sofie-automation/corelib/dist/protectedString'
+import { useTranslation } from 'react-i18next'
+import { useParams } from 'react-router-dom'
 
 const TimelineDatastore = new Mongo.Collection<TimelineDatastoreEntry>('timelineDatastore')
 
-interface ITimelineDatastoreViewProps {
-	match?: {
-		params?: {
-			studioId: StudioId
-		}
-	}
+interface TimelineDatastoreViewRouteParams {
+	studioId: string
 }
-interface ITimelineDatastoreViewState {}
-const TimelineDatastoreView = translateWithTracker<ITimelineDatastoreViewProps, ITimelineDatastoreViewState, {}>(
-	(_props: ITimelineDatastoreViewProps) => {
-		return {}
-	}
-)(
-	class TimelineDatastoreView extends MeteorReactComponent<
-		Translated<ITimelineDatastoreViewProps>,
-		ITimelineDatastoreViewState
-	> {
-		constructor(props: Translated<ITimelineDatastoreViewProps>) {
-			super(props)
-		}
 
-		render() {
-			const { t } = this.props
+const TimelineDatastoreView: React.FC = function TimelineDatastoreView() {
+	const { t } = useTranslation()
+	const { studioId } = useParams<TimelineDatastoreViewRouteParams>()
 
-			return (
-				<div className="mtl gutter">
-					<header className="mvs">
-						<h1>{t('Timeline Datastore')}</h1>
-					</header>
-					<div className="mod mvl">
-						{this.props.match && this.props.match.params && (
-							<div>
-								{/* <ComponentMappingsTable studioId={this.props.match.params.studioId} /> */}
-								<ComponentDatastoreControls studioId={this.props.match.params.studioId} />
-							</div>
-						)}
+	return (
+		<div className="mtl gutter">
+			<header className="mvs">
+				<h1>{t('Timeline Datastore')}</h1>
+			</header>
+			<div className="mod mvl">
+				{studioId && (
+					<div>
+						<ComponentDatastoreControls studioId={protectString(studioId)} />
 					</div>
-				</div>
-			)
-		}
-	}
-)
+				)}
+			</div>
+		</div>
+	)
+}
 
 interface IDatastoreControlsProps {
 	studioId: StudioId
 }
 function ComponentDatastoreControls({ studioId }: IDatastoreControlsProps) {
-	useSubscription(PubSub.timelineDatastore, { studioId })
+	useSubscription(PubSub.timelineDatastore, studioId)
 
 	const datastore = useTracker(() => TimelineDatastore.find().fetch(), [studioId])
-
-	const createOrEdit = (key: string, value: any) => {
-		const doc = datastore?.find((entry) => entry.studioId === studioId && entry.key === key)
-		if (doc) {
-			TimelineDatastore.update(doc._id, {
-				$set: {
-					value,
-					modified: Date.now(),
-				},
-			})
-		} else {
-			TimelineDatastore.insert({
-				studioId,
-
-				key,
-				value,
-
-				modified: Date.now(),
-			})
-		}
-	}
 
 	return (
 		<div>
 			<div>
-				<button className="btn btn-primary" onMouseDown={() => createOrEdit('input', 1)}>
-					1
-				</button>
-				<button className="btn btn-primary" onMouseDown={() => createOrEdit('input', 2)}>
-					2
-				</button>
-				<button className="btn btn-primary" onMouseDown={() => createOrEdit('input', 3)}>
-					3
-				</button>
+				<table className="testtools-timelinetable">
+					<tbody>
+						<tr>
+							<th>Key</th>
+							<th>Last modified</th>
+							<th>Type</th>
+							<th>Value</th>
+						</tr>
+						{datastore?.map((entry) => (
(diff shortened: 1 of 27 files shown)


meteor/lib/collections/TimelineDatastore.ts
```diff
@@ -0,0 +1,16 @@
+import { createMongoCollection } from './lib'
+import { registerIndex } from '../database'
+import { TimelineDatastoreEntryId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+export { TimelineDatastoreEntryId }
+import { CollectionName } from '@sofie-automation/corelib/dist/dataModel/Collections'
+
+import { DBTimelineDatastoreEntry } from '@sofie-automation/corelib/dist/dataModel/TimelineDatastore'
+export * from '@sofie-automation/corelib/dist/dataModel/Segment'
```
Julusian, line 8: Is this reexport of 'Segment' intentional?

meteor/lib/collections/TimelineDatastore.ts
```diff
@@ -0,0 +1,16 @@
+import { createMongoCollection } from './lib'
+import { registerIndex } from '../database'
+import { TimelineDatastoreEntryId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+export { TimelineDatastoreEntryId }
```
Julusian, line 4: It would be good to avoid re-exporting the Id, there is a task in the backlog to remove a lot of these re-exports

packages/blueprints-integration/src/action.ts
```diff
@@ -104,6 +104,8 @@ export interface IBlueprintActionManifest {
 	/** Optional ways of executing this action. The default option is computed from the display properties */
 	triggerModes?: IBlueprintActionTriggerMode[]
 
+	// supportsDataStoreFastRoute?: boolean
```
Julusian, line 107: unnecessary comment

packages/blueprints-integration/src/context.ts
```diff
@@ -148,7 +148,15 @@ export interface ISegmentUserContext extends IUserNotesContext, IRundownContext,
 }
 
 /** Actions */
-export interface IActionExecutionContext extends IShowStyleUserContext, IEventContext {
+export interface IDataStoreActionExecutionContext extends IShowStyleUserContext, IEventContext {
+	setTimelineDatastoreValue(key: string, value: any, mode: 'temporary' | 'indefinite'): Promise<void>
```
Julusian, line 152: These new methods would benefit from some description comments

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -64,27 +65,22 @@ export enum ActionPartChange {
...
-	private readonly rundown: DBRundown
-	private readonly playlistActivationId: RundownPlaylistActivationId
+export class DatastoreActionExecutionContext
+	extends ShowStyleUserContext
+	implements IDataStoreActionExecutionContext, IEventContext
+{
+	protected readonly _context: JobContext
 
 	/** To be set by any mutation methods on this context. Indicates to core how extensive the changes are to the current partInstance */
 	public currentPartState: ActionPartChange = ActionPartChange.NONE
```
Julusian, line 75: These currentPartState and nextPartState properties look unused, could they be removed?

packages/job-worker/src/playout/adlib.ts
```diff
@@ -109,10 +109,29 @@ export async function takePieceAsAdlibNow(context: JobContext, data: TakePieceAs
...
+						const playlist = cache.PlaylistId
 						const executeProps = pieceToCopy.allowDirectPlay
+						const showStyle = await context.getShowStyleCompound(
+							rundown.showStyleVariantId,
+							rundown.showStyleBaseId
+						)
+						const blueprint = await context.getShowStyleBlueprint(showStyle._id)
+
+						const currentPartInstance = playlist.currentPartInstanceId
+							? await context.directCollections.PartInstances.findOne(playlist.currentPartInstanceId)
```
Julusian, line 121: This should be pulling from the cache, not the database.

In fact, we already have the currentPartInstance as a variable (code might want some refactoring to clarify this) on line 86. We confirmed on line 63 that it is the currentPartInstance.

packages/blueprints-integration/src/context.ts
```diff
@@ -148,7 +148,15 @@ export interface ISegmentUserContext extends IUserNotesContext, IRundownContext,
...
+	removeTimelineDatastoreValue(key: string): Promise<void>
+}
+
+export interface IActionExecutionContext
+	extends IShowStyleUserContext,
```
Julusian, line 157: Should this be extending `IDataStoreActionExecutionContext`? Is there any value in letting blueprints set datastore values during the 'slow' method?

packages/job-worker/src/playout/cache.ts
```diff
@@ -124,6 +125,7 @@ export class CacheForPlayout extends CacheForPlayoutPreInit implements CacheForS
 	private toBeRemoved = false
 
 	public readonly Timeline: DbCacheWriteOptionalObject<TimelineComplete>
+	public readonly TimelineDatastore: DbCacheWriteCollection<DBTimelineDatastoreEntry>
```
Julusian, line 128: I am still currently thinking that it would make sense to not include this on the cache.
Because it is present during the action-execution (where adlib-actions could modify it), having it here feels risky.

As it only gets used during deactivation, I think it acceptable for that to be done directly on the database for now. We can always change it back once we have a clearer plan on where else it would be useful
(Idea for later: this could use a `DbCacheWriteSyncedCollection`? all modifications get immediately pushed to the database, as well as updating a cached copy. That might be useful in other places too)

packages/job-worker/src/playout/playout.ts
```diff
@@ -1093,51 +1102,135 @@ function timelineTriggerTimeInner(
...
+				fromPieceType: {
+					$in: [ExpectedPackageDBType.ADLIB_ACTION, ExpectedPackageDBType.BASELINE_ADLIB_ACTION],
 				},
-				async (actionContext, _rundown, _currentPartInstance, blueprint) => {
-					if (!blueprint.blueprint.executeAction) throw UserError.create(UserErrorMessage.ActionsNotSupported)
+			}
 
-					logger.info(`Executing AdlibAction "${data.actionId}": ${JSON.stringify(data.userData)}`)
+			if (blueprint.blueprint.executeDataStoreAction) {
+				const watchedPackages = await WatchedPackagesHelper.create(
```
Julusian, line 1142: We appear to be doing this `WatchedPackagesHelper.create()` both here and inside executeActionInner.
It would be nice to do it once and reuse it

meteor/client/ui/TestTools/TimelineDatastore.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+interface ITimelineDatastoreViewProps {
+	match?: {
+		params?: {
+			studioId: StudioId
+		}
+	}
+}
+interface ITimelineDatastoreViewState {}
+const TimelineDatastoreView = translateWithTracker<ITimelineDatastoreViewProps, ITimelineDatastoreViewState, {}>(
```
Julusian, line 26: Is there a reason some of these components are classes and some are functions?
Not a big deal, but would be nice for them to be consistent :)

meteor/lib/collections/TimelineDatastore.ts
baltedewit: No, copy past from Segment.ts. I missed that

meteor/lib/collections/TimelineDatastore.ts
baltedewit: Same here, it's consistent with the other files in this subfolder. I can't find the task but if there is a good reference / design to follow I'm happy to do so

packages/job-worker/src/playout/playout.ts
```diff
@@ -1093,51 +1102,135 @@ function timelineTriggerTimeInner(
...
+						},execution=${getRandomId()}`,
+						tempSendUserNotesIntoBlackHole: true, // TODO-CONTEXT store these notes
+					},
+					context,
+					showStyle,
+					context.getShowStyleBlueprintConfig(showStyle),
+					watchedPackages
+				)
+
+				logger.info(`Executing AdlibAction "${data.actionId}": ${JSON.stringify(data.userData)}`)
```
jstarpl, line 1162: This should probably specify that it's using the executeDataStoreAction API.

packages/job-worker/src/playout/playout.ts
```diff
@@ -1093,51 +1102,135 @@ function timelineTriggerTimeInner(
...
+
+				try {
+					await blueprint.blueprint.executeDataStoreAction(
+						actionContext,
+						data.actionId,
+						data.userData,
+						data.triggerMode
+					)
+				} catch (err) {
+					logger.error(`Error in showStyleBlueprint.executeAction: ${stringifyError(err)}`)
```
jstarpl, line 1172: Should this be `showStyleBlueprint.executeDataStoreAction`?

packages/job-worker/src/playout/playout.ts
```diff
@@ -1093,51 +1102,135 @@ function timelineTriggerTimeInner(
...
+				pieceId: data.actionDocId,
+				fromPieceType: {
+					$in: [ExpectedPackageDBType.ADLIB_ACTION, ExpectedPackageDBType.BASELINE_ADLIB_ACTION],
 				},
-				async (actionContext, _rundown, _currentPartInstance, blueprint) => {
-					if (!blueprint.blueprint.executeAction) throw UserError.create(UserErrorMessage.ActionsNotSupported)
+			}
 
-					logger.info(`Executing AdlibAction "${data.actionId}": ${JSON.stringify(data.userData)}`)
+			if (blueprint.blueprint.executeDataStoreAction) {
```
jstarpl, line 1141: Should there be an explanation of what `executeDataStoreAction` is trying to achieve here? I feel like we should have some sort of in-code or md file here, describing the intent: allowing a fast path to execute first and then pull in all the data. The timeline isn't obvious - unless you understand the underlying data IO ops underneath. `const fullCache: CacheForPlayout = await CacheForPlayout.fromInit(context, initCache)` doesn't seem as heavy as it is. Also, the fact that these operations are designed to work in-line and in-sequence, isn't immediately obvious. Someone could possibly try and put both of these in a Promise.all, with the side-effects not becoming apparent immediately.

jstarpl (changes requested):
I think some documentation of intent would be good to add, since the real-world timeline of events, underlying the whole concept isn't immediately obvious.

packages/blueprints-integration/src/context.ts
baltedewit: Well, if you need some of the data from the cache in order to make a decision you will have to use the 'slow' method. Maintenance wise it's cheap enough to leave it in, so I don't want to make the decision to limit the usage in that way right now.

meteor/server/publications/timeline.ts
```diff
@@ -31,6 +32,16 @@ meteorPublish(PubSub.timeline, function (selector, token) {
 	}
 	return null
 })
+meteorPublish(PubSub.timelineDatastore, function (selector, token) {
```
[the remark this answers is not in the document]
baltedewit: I think it's good to have this bundled, if you look at i.e. rundown.ts in the same folder you'll see that that has a lot of publications bundled as well. Since the datastore is so closely related to the timeline I think this makes sense.

meteor/server/publications/timeline.ts
```diff
@@ -50,6 +61,24 @@ meteorCustomPublishArray<RoutedTimeline>(
...
+			PeripheralDeviceReadAccess.peripheralDeviceContent({ deviceId: deviceId }, { userId: this.userId, token })
+		) {
+			const peripheralDevice = PeripheralDevices.findOne(deviceId)
+
+			if (!peripheralDevice) throw new Meteor.Error('PeripheralDevice "' + deviceId + '" not found')
+
+			const studioId = peripheralDevice.studioId
+			if (!studioId) return []
+
+			createObserverForTimelineDatastorePublication(pub, PubSub.timelineDatastoreForDevice, studioId)
```
nytamin, line 78: From what I see, there's no transformation of data or caching going in the custom publish function. If that's the case we should not use the `meteorCustomPublishArray` since that adds unnecessary overhead (performance penalty) for the publication, rather we should just do a `meteorPublish`:
```typescript
meteorPublish(PubSub.timelineDatastoreForDevice, function (deviceId: string, token: string) {
	if (PeripheralDeviceReadAccess.peripheralDeviceContent({ deviceId: deviceId }, { userId: this.userId, token })) {
			const peripheralDevice = PeripheralDevices.findOne(deviceId)

			if (!peripheralDevice) throw new Meteor.Error('PeripheralDevice "' + deviceId + '" not found')

			const studioId = peripheralDevice.studioId
			if (!studioId) return null
			const modifier: FindOptions<TimelineDatastoreEntry> = {
				fields: {},
			}

			return TimelineDatastore.find({studioId}, modifier)
	}
	return null
})
```

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -64,9 +65,72 @@ export enum ActionPartChange {
...
+			showStyle,
+			showStyleBlueprintConfig,
+			watchedPackages
+		)
+		this._context = context
+	}
+
+	async setTimelineDatastoreValue(key: string, value: unknown, mode: 'temporary' | 'indefinite'): Promise<void> {
+		const collection = this._context.directCollections.TimelineDatastores
+		const entry = await collection.findOne({ studioId: this._context.studioId, key })
```
nytamin, line 94: If I've understood this correctly: There will only be one entry per `studio+key`?

If so, a few notes:
1. The `_id`  should not be a random string, but rather be a unique hash consisting of the studioId+key.
2. This ^ gives: the update could be done using an `upsert` instead, since that would reduce the number of db-calls.

```typescript
const id = protectedString(`${studioId}_${key}`)
collection.upsert(id, {$set: {
	_id: id,
	studioId: this._context.studioId,

	key,
	value,

	modified: Date.now(),
	mode,
}})
```

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -64,9 +65,72 @@ export enum ActionPartChange {
...
+					mode,
+					modified: Date.now(),
+				},
+			})
+		}
+	}
+
+	async removeTimelineDatastoreValue(key: string): Promise<void> {
+		const collection = this._context.directCollections.TimelineDatastores
+		await collection.remove({ studioId: this._context.studioId, key })
```
nytamin, line 120: ref: my comment on line 94:
instead do a lookup using the hash-id, otherwise this would cause a full-db-search.

packages/job-worker/src/playout/datastore.ts
```diff
@@ -0,0 +1,23 @@
+import { deserializeTimelineBlob } from '@sofie-automation/corelib/dist/dataModel/Timeline'
+import { JobContext } from '../jobs'
+import { CacheForPlayout } from './cache'
+
+export async function cleanTimelineDatastore(context: JobContext, cache: CacheForPlayout): Promise<void> {
```
nytamin, line 5: This function could use a comment describing what it does.
"Remove entries from TimelineDatastore that have no references left in the Timeline"

packages/job-worker/src/playout/datastore.ts
```diff
@@ -0,0 +1,23 @@
...
+	const timeline = cache.Timeline.doc
+	const datastore = await context.directCollections.TimelineDatastores.findFetch({})
```
nytamin, line 7: add studioId in the selector here

packages/job-worker/src/playout/datastore.ts
```diff
@@ -0,0 +1,23 @@
...
+	if (!timeline) {
+		return
+	}
+
+	const timelineObjs = deserializeTimelineBlob(timeline.timelineBlob)
+
+	const timelineRefs = timelineObjs
+		.filter((o) => o.content.$references) // todo - update timeline types package
+		.flatMap((o) => Object.keys(o.content.$references))
+	const inactiveKeys = datastore.map((o) => o.key).filter((k) => !!timelineRefs.find((r) => r === k)) // todo - protect some values from being deleted?
```
nytamin, line 18: Should this take `mode: 'indefinite'` into account?

nytamin (commented):
There are a few things I'd like to change, but nothing major.

(fyi: I've ignored commenting on things in playout.ts that cause build errors.)

meteor/lib/api/pubsub.ts
```diff
@@ -162,6 +165,10 @@ export interface PubSubTypes {
 	[PubSub.studios]: (selector: MongoQuery<DBStudio>, token?: string) => DBStudio
 	[PubSub.studioOfDevice]: (deviceId: PeripheralDeviceId, token?: string) => DBStudio
 	[PubSub.timeline]: (selector: MongoQuery<TimelineComplete>, token?: string) => TimelineComplete
+	[PubSub.timelineDatastore]: (
+		selector: MongoQuery<DBTimelineDatastoreEntry>,
```
Julusian, line 169: I would like to see this publication be not using a mongo query, and instead be using a studioId, or whatever identifier is relevant

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -61,15 +62,65 @@ import { isPartPlayable } from '@sofie-automation/corelib/dist/dataModel/Part'
...
+			value,
+
+			modified: Date.now(),
+			mode,
+		})
+	}
+
+	async removeTimelineDatastoreValue(key: string): Promise<void> {
+		const studioId = this._context.studioId
+		const id = protectString(`${studioId}_${key}`)
```
Julusian, line 107: It would be good to define this id generation as a function, to ensure it is kept consistent.

Julusian (approved):

PR #768, opened 2022-09-30
feat: split collections to have server and client versions SOFIE-1141

Julusian:

Code quality

We have all the mongodb collections defined in `meteor/lib/collections`. These are then used by both the server and the client.
They expose both async and sync methods, even though we only want to use sync methods on the client (for now), and async methods on the server (to prepare for removal of fibers). in shared code, we need to decide on a strategy.

This splits our wrapping around `Mongo.Collection` into a few more specialised types. These are various combinations of async/sync only, and readonly/writable.

This means we can limit what each collection allows callers to do, doing so differently on each of the client and server. and will let us gradually convert each collection in preperation for Meteor 3.0.

Some client versions of collections have been omitted as they are not used by the client. I have made sure that they are not being published.

For now the client collections reside in `meteor/lib`, so that the shared code in lib can use them. We will likely want to move them eventually, but that will be a much smaller and easier change.  
It currently looks a bit unclear as to whether the sync methods will be usable on the client in later versions of meteor, so in the future these may need to become async too, but this lets us approach that independently to the server side.

Minimongo is fine with these being defined like this, any 'collections' used by custom publications have their definitions scattered somewhere in the ui code for a while.

* **Other information**:

Most of this change is updating imports. The files containing the new definitions are of the most interest:
* https://github.com/nrkno/sofie-core/tree/feat/seperate-server-client-collections/meteor/server/collections
* https://github.com/nrkno/sofie-core/tree/feat/seperate-server-client-collections/meteor/client/collections
* https://github.com/nrkno/sofie-core/tree/feat/seperate-server-client-collections/meteor/lib/collections
These could be broken out to more files if necessary. I am not sure that either will get long enough to justify it.

This will allow us to remove a lot of the `meteor/lib/collections/*` files in a further round of refactoring, as most of them are simply re-exporting types that moved to corelib a while ago as well as defining the collections which are being moved as part of this.

There is some complications here due to some shared code in lib that needs access to collections. For now it has access to some collections in lib, which are re-exported by the client and so are treated as client versions.
Tidying this up needs a proper solution, which can be handled as a follow up PR

**Status**


meteor/server/collections/index.ts
```diff
@@ -0,0 +1,445 @@
...
+import { DBTriggeredActions } from '../../lib/collections/TriggeredActions'
+import { UserActionsLogItem } from '../../lib/collections/UserActionsLog'
+import { DBUser } from '../../lib/collections/Users'
+import { WorkerStatus } from '../../lib/collections/Workers'
+import { registerIndex } from './indices'
+import { getCurrentTime } from '../../lib/lib'
+import { createAsyncMongoCollection, createAsyncOnlyMongoCollection, wrapMongoCollection } from './collection'
+import { ObserveChangesForHash, registerCollection } from './lib'
+
+export const AdLibActions = createAsyncMongoCollection<AdLibAction>(CollectionName.AdLibActions)
```
Julusian, line 54: I feel like this file should be split up into multiple, but im not sure how best to group everything as I think a file per collection will be too many

nytamin (approved):

PR #769, opened 2022-10-04
fix(Prompter): Prompter doesn't scroll to the correct Part when using followtake

jstarpl:

Bug fix

The prompter can scroll to the wrong Part (it scrolls to the Part that _used to_ be the live Part instead of the new live Part.

This happens, because the refresh of the "inner" PrompterView may need to delay update to make sure that all the PieceInstances have streamed in.

The scroll operation scrolls to the specific PartInstance, instead of relying on "live" class being present on the prompter's element.

* **Other information**:

The PrompterView needs to be rewritten to move away from it's legacy architecture.

**Status**

### DIFF
--- a/meteor/client/ui/Prompter/PrompterView.tsx
+++ b/meteor/client/ui/Prompter/PrompterView.tsx
@@ -290,13 +290,6 @@ export class PrompterViewInner extends MeteorReactComponent<Translated<IProps & 
 	componentDidUpdate(prevProps) {
 		this.triggerCheckCurrentTakeMarkers()
 		this.checkScrollToCurrent()
-
-		console.log(
-			'c:',
-			this.props.rundownPlaylist?.currentPartInstanceId,
-			prevProps.rundownPlaylist?.currentPartInstanceId
-		)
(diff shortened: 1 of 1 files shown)


meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -792,15 +801,10 @@ export const Prompter = translateWithTracker<IPrompterProps, {}, IPrompterTracke
 				segment.parts.forEach((part) => {
 					lines.push(
 						<div
-							id={`part_${part.id}`}
+							id={`partInstance_${part.id}`}
```
nytamin, line 804: Minor note: should `part` be renamed to `partInstance` as well?

nytamin (approved):

meteor/client/ui/Prompter/PrompterView.tsx
jstarpl: The whole Prompter API just needs reworking. It's using it's own set of PrompterDataSegment and PrompterDataPart that then borrows the source's Ids, so the naming is way too confusing. I don't want to spend too much time on this, we need to refactor the Prompter to actually have a sound architecture.

PR #772, opened 2022-10-05
feat: differentiate user/default settings datastructure changes

Julusian:

feature

It is impossible to tell what settings are owned by the blueprints and what are owned by the user

This lays the groundwork to use the `ObjectWithOverrides<T>` for various properties that the blueprints can mutate through migrations.
The UI and blueprints are both modifying the defaults portion of the data currently, to preserve existing behaviour.

To facilitate this, there are some new custom publications added, intended for the playout portion of the ui to use. These pre-apply the overrides, and return documents much more similar to before. There are however some subtle differences.
The `ObjectWithOverrides<T>` flow doesn't handle arrays well. Attempting to preserve order of the elements is hard, so we currently treat an array as a single value. This means that any properties which are now wrapped in `ObjectWithOverrides` have been converted to `Record<>`. This continues through to the ui, as it can be of benefit there too.
These new publications are using alternate typings and are being intentionally minimal in their definition, to minimise the data sent to the ui, and minimise the unnecessary reactivity it causes.
This has caused a lot of churn in the ui code, to update the type imports across many files.

The job-worker was much simpler to solve, as we cache the affected documents on the context. It is now pre-apply the overrides before storing the document in this cache. As such, the job-worker is using custom extensions of ShowStyleBase and Studio. This was a fairly straightforward change.

I believe we are not using the affected properties in the meteor server anymore, so very little has changed there. 

The only change to blueprints-integration is changing the `actions` and `triggers` properties on TriggeredActions to be `Record<string, T>`. This is to match how they are represented in the db, and they do not have an id field we can use for translation.
Everywhere else we have kept them as arrays into the blueprints, as they have ids we can use to map into the object form.

* **Other information**:

When we get to updating the ui to be creating the overrides, we should evaluate the 'old' publications. It might be better to remove them and replace them with some more specialised publications, to avoid confusion over two similar publications.

TODO
* [ ] Investigate potential performance penalty of new custom publications

**Status**


Julusian:
@jstarpl I am interested as to what you think about the changes to publications.
In particular the new `UI*` publications, intended for the playout portions of the UI. (Maybe settings should use them too, but as part of the ui changes in the follow up task).

One thing I did do as well, was to make the publications only accept a single id. So that
```
this.subscribe(PubSub.showStyleBases, {
	_id: {
		$in: rundowns.map((rundown) => rundown.showStyleBaseId),
	},
})
```
becomes
```
for (const rundown of rundowns) {
	this.subscribe(PubSub.uiShowStyleBase, rundown.showStyleBaseId)
}
```
Im hoping this isnt a problem spawning off multiple new subscriptions like this. I did it so that the publications can be shared better across subscriptions with overlapping but different sets of ids. And this will also avoid churn when the set of ids changes

jstarpl:
>

jstarpl:
> @jstarpl I am interested as to what you think about the changes to publications. In particular the new `UI*` publications, intended for the playout portions of the UI. (Maybe settings should use them too, but as part of the ui changes in the follow up task).
> Im hoping this isnt a problem spawning off multiple new subscriptions like this. I did it so that the publications can be shared better across subscriptions with overlapping but different sets of ids. And this will also avoid churn when the set of ids changes

Well, the set publications were based on Johan's note that Publications in Meteor are reasonably heavy on the server, so having a single multi-document publication is better than multiple single-document ones. That being said, we only have a couple of showStyles at a time, so I'm not sure if that matters.

jstarpl (approved):
Overall, I seems good, I think I'd like to hear Johans thoughts on the Publications performance

nytamin (approved):
> > @jstarpl I am interested as to what you think about the changes to publications. In particular the new `UI*` publications, intended for the playout portions of the UI. (Maybe settings should use them too, but as part of the ui changes in the follow up task).
> > Im hoping this isnt a problem spawning off multiple new subscriptions like this. I did it so that the publications can be shared better across subscriptions with overlapping but different sets of ids. And this will also avoid churn when the set of ids changes
> 
> Well, the set publications were based on Johan's note that Publications in Meteor are reasonably heavy on the server, so having a single multi-document publication is better than multiple single-document ones. That being said, we only have a couple of showStyles at a time, so I'm not sure if that matters.

I think this is one of the "it depends" -type of situations. If it's a question of having 1000 single-document subscriptions vs 1 multi-document, I think the single subscription is better.
But if it's only a couple of subscriptions like here (and they can be shared with others) I think this is the way to go.
(This is based on anecdotal experience and gut feelings. I don't currently have the data to back this up.)

jstarpl:
> I think this is one of the "it depends" -type of situations. If it's a question of having 1000 single-document subscriptions vs 1 multi-document, I think the single subscription is better. But if it's only a couple of subscriptions like here (and they can be shared with others) I think this is the way to go. (This is based on anecdotal experience and gut feelings. I don't currently have the data to back this up.)

If that's the case, then I think there's no reason not to go with the setup Julian proposed. It's just 2-3 subscriptions max per RundownView, so I expect this to be perfectly fine.

PR #777, opened 2022-10-17
feat: Sofie REST API

thomasslee97:

This PR introduces a versioned Sofie REST API, with the aim of provided an API with some form of stability.
Please note that at this stage this PR is a draft that serves as a place for discussing the development of this feature.


meteor/lib/api/rest/interfaces.ts
```diff
@@ -0,0 +1,41 @@
+export interface RestAPIAdLibActions {}
+export interface RestAPIAdLibPieces {}
+export interface RestAPIBlueprints {}
+export interface RestAPIBucketAdlibActions {}
+export interface RestAPIBucketAdlibs {}
+export interface RestAPIBuckets {}
+export interface RestAPICoreSystem {}
+export interface RestAPIEvaluations {}
+export interface RestAPIExpectedMediaItems {}
```
jstarpl, line 9: I think we should probably keep Expected Media Items out of the stable API, considering that it has been on a path to deprecation for a good while now.

meteor/lib/api/rest/interfaces.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface RestAPIExpectedPackages {}
+export interface RestAPIExpectedPackageWorkStatuses {}
+export interface RestAPIExpectedPlayoutItems {}
+export interface RestAPIExternalMessageQueue {}
+export interface RestAPIIngestDataCache {}
+export interface RestAPIMediaObjects {}
```
jstarpl, line 15: Same as Expected Media Items.

meteor/lib/api/rest.ts
```diff
@@ -0,0 +1,204 @@
...
+		rundownPlaylistId: RundownPlaylistId,
+		adLibId: AdLibActionId | RundownBaselineAdLibActionId | PieceId | BucketAdLibId,
+		connection: Meteor.Connection,
+		triggerMode?: string
+	): Promise<ClientAPI.ClientResponse<object>>
+	/**
+	 * Moves the next point by `delta` places. Negative values are allowed to move "backwards" in the script.
+	 *
+	 * Throws if the target Playlist is not active.
+	 * Throws if there is not next Part set (e.g. rundown is empty)
```
jstarpl, line 83: There can be no next Part set, if this is the end of the Rundown, I suppose it shouldn't throw then, right?

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,660 @@
...
+		return ClientAPI.responseSuccess({ version: CURRENT_SYSTEM_VERSION })
+	}
+	async activate(
+		rundownPlaylistId: RundownPlaylistId,
+		rehearsal: boolean,
+		connection: Meteor.Connection
+	): Promise<ClientAPI.ClientResponse<void>> {
+		return ServerClientAPI.runUserActionInLogForPlaylistOnWorker(
+			{ ...this, connection: connection },
+			REST_API_USER_EVENT,
```
jstarpl, line 54: I think this should contain some information about the request that caused this to execute, so that this provides the same debugging value as the current "user event summary". Request URL maybe?

packages/live-status-gateway/package.json
```diff
@@ -0,0 +1,79 @@
...
+		"ncs",
+		"nrcs",
+		"rundown",
+		"production"
+	],
+	"dependencies": {
+		"@sofie-automation/blueprints-integration": "link:../blueprints-integration",
+		"@sofie-automation/server-core-integration": "link:../server-core-integration",
+		"@sofie-automation/corelib": "link:../corelib",
+		"@sofie-automation/shared-lib": "link:../shared-lib",
```
jstarpl, line 58: It is my understanding that these sorts of links don't work very well. Apparently "*" should work, since these are all present in the workspace?

packages/live-status-gateway/src/liveStatusServer.ts
```diff
@@ -0,0 +1,103 @@
...
+		// const segmentHandler = new SegmentHandler(this._logger, this._coreHandler)
+		// await segmentHandler.init()
+		// const partHandler = new PartHandler(this._logger, this._coreHandler)
+		// await partHandler.init()
+		const partInstancesHandler = new PartInstancesHandler(this._logger, this._coreHandler)
+		await partInstancesHandler.init()
+		const adLibActionsHandler = new AdLibActionsHandler(this._logger, this._coreHandler)
+		await adLibActionsHandler.init()
+		const globalAdLibActionsHandler = new GlobalAdLibActionsHandler(this._logger, this._coreHandler)
+		await globalAdLibActionsHandler.init()
```
jstarpl, line 55: This is omitting AdLibPieces and RundownAdLibPieces. Is that on purpose?

meteor/lib/api/rest.ts
```diff
@@ -0,0 +1,204 @@
...
+	 * @param actionId Action Id string, should match something inside the blueprints.
+	 * @param userData Any value, recommended to be a relatively simple { [key: string]: string }-style object, as the object may need to survive a round of serialization.
+	 * @param connection Connection data including client and header details
+	 */
+	executeAction(
+		rundownPlaylistId: RundownPlaylistId,
+		actionId: string,
+		userData: any,
+		connection: Meteor.Connection
+	): Promise<ClientAPI.ClientResponse<object>>
```
jstarpl, line 61: This feels unnecessary? Since there already is an `executeAdLib` method, that accurately allows executing all AdLibs in the system, why is this one needed? How would a system even know what actionId`s are available?

meteor/lib/api/rest.ts
```diff
@@ -0,0 +1,204 @@
...
+	 * @param adLibId AdLib to execute.
+	 * @param connection Connection data including client and header details
+	 * @param triggerMode A string to specify a particular variation for the AdLibAction, valid actionType strings are to be read from the status API.
+	 */
+	executeAdLib(
+		rundownPlaylistId: RundownPlaylistId,
+		adLibId: AdLibActionId | RundownBaselineAdLibActionId | PieceId | BucketAdLibId,
+		connection: Meteor.Connection,
+		triggerMode?: string
+	): Promise<ClientAPI.ClientResponse<object>>
```
jstarpl, line 78: This doesn't allow triggering some "AdLib-like" items that are present in the GUI, specifically: "Sticky" source-layer adLibs and "Clear" source-layer adLibs.

meteor/lib/api/rest.ts
Julusian: I believe the intention here is to be able to execute a 'known' action without it being an adlib in sofie. Allowing the external system to run a super without needing to ingest it into the rundown or a bucket first.

Maybe the name of this method could be changed to clarify that? Im not sure what to

packages/live-status-gateway/package.json
Julusian: yes, or whatever we do in the other gateways.

We have to do this in some places, as meteor isn't part of the workspace, but meteor wont be importing any code from this package

packages/live-status-gateway/Dockerfile.circle
```diff
@@ -0,0 +1,13 @@
+FROM node:16.17-alpine
```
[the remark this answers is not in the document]
Julusian: we use these for github actions building, they should probably have be renamed to clarify that by now

packages/live-status-gateway/src/liveStatusServer.ts
thomasslee97: Nope, just hasn't been done

meteor/lib/api/rest.ts
thomasslee97: This method should be removed I think, it was implemented to reduce the amount of knowledge sharing via the live status gateway required short-term. But the use of it here is that we have separately codified all of the avaialble adlib actions in the blueprints, so an external system can call `executeAction` in the blueprints with arbitrary data. Highly unsafe, definitely should go, but I think it needs to be replaced with a specifiaction for communicating the editable fields of an adlib action so that the other adlib method can be called with arbitrary (within reason) arguments.

meteor/lib/api/rest.ts
thomasslee97: I wonder how to represent those. Perhaps `clearLayer` and `recallSticky` or similar new methods could be exposed to abstract away the fact that those are adlibs? Becuase isn't the adlib representation of those actions sort of a cheat to get those actions to appear in various UIs?

meteor/lib/api/methods.ts
```diff
@@ -34,6 +35,7 @@ interface IMeteorCall {
 	migration: NewMigrationAPI
 	peripheralDevice: NewPeripheralDeviceAPI
 	playout: NewPlayoutAPI
+	rest: RestAPI
```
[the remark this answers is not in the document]
thomasslee97: Aye, this is the way we were able to put together to allow us to gain a Meteor context via `MeteorCall`, which we currently need to be able to call some methods. I would love to hear alternatives.

packages/openapi/package.json
```diff
@@ -0,0 +1,55 @@
...
+		"/dist",
+		"install_swagger.js"
+	],
+	"dependencies": {
+		"tslib": "^2.4.0"
+	},
+	"devDependencies": {
+		"@openapitools/openapi-generator-cli": "^2.5.2",
+		"eslint-plugin-yml": "^1.2.0",
+		"jest-teamcity-reporter": "0.9.0",
```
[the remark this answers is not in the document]
thomasslee97: Yes, note to self to remove

meteor/lib/api/rest/interfaces.ts
```diff
@@ -0,0 +1,41 @@
+export interface RestAPIAdLibActions {}
```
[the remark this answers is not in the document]
thomasslee97: Yes, they were placeholders for discussion

baltedewit (commented):
This looks like a good addition to the feature set, I don't have much else to add as of now

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,660 @@
...
+const koaRouter = new KoaRouter()
+
+koaRouter.get('/', async (ctx, next) => {
+	ctx.type = 'application/json'
+	ctx.body = ClientAPI.responseSuccess(await MeteorCall.rest.index())
+	ctx.status = 200
+	await next()
+})
+
+koaRouter.post('/activate/:playlistId', async (ctx, next) => {
```
LindvedKrvang, line 379: All these URLs don't seem to follow the REST standard.
In REST you design the API around the "resources" (entities etc) in your app, not the operations of the app. 
So in a REST API this URL would be `/playlists/{playlistId}/activate`.

Also, if you have several ids in your URL you still need to specify the resources of those ids.
Further down you have this URL: `/setNextPart/:playlistId/:partId`. Following REST this should be `/playlists/{playlistId}/parts/{partId}/setNext` (Notice that I don't need to write `setNextPart` because that information is now in the URL.
For the Segment equivalent it would be `/playlists/{playlistId}/segments/{segmentId}/setNext`

https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#organize-the-api-design-around-resources

packages/live-status-gateway/src/collections/partInstances.ts
```diff
@@ -0,0 +1,117 @@
+import { Logger } from 'winston'
+import { CoreHandler } from '../coreHandler'
+import { CollectionBase, Collection, CollectionObserver } from '../wsHandler'
+import { CoreConnection } from '@sofie-automation/server-core-integration'
+import { DBRundownPlaylist } from '@sofie-automation/corelib/dist/dataModel/RundownPlaylist'
+import { DBPartInstance } from '@sofie-automation/corelib/dist/dataModel/PartInstance'
+import { unprotectString } from '@sofie-automation/corelib/dist/protectedString'
+
+export enum PartInstanceName {
+	cur = 'cur',
```
LindvedKrvang, line 10: If this is supposed to be `current` then please call it `current`. 
I don't think you mean an actual Cur
https://en.wikipedia.org/wiki/Cur

packages/live-status-gateway/src/wsHandler.ts
```diff
@@ -0,0 +1,109 @@
+import { StudioId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+import { Observer } from '@sofie-automation/server-core-integration'
+import { Logger } from 'winston'
+import { WebSocket } from 'ws'
+import { CoreHandler } from './coreHandler'
+
+export abstract class WsTopicBase {
```
LindvedKrvang, line 7: I assume that you mean `WebSocket` with `Ws`?
Would be nice if abbreviations weren't used as it forces the reading developer to guess what is meant (and guessing leads to wrong guesses which leads to errors and/or increased development time). 
Here `Ws` could easily be either `WebSocket` or `WebServer`.

Also, what does `Topic` refere to? You have an interface called `WsTopic` but that interface is not used at all in the implementation of `WsTopicBase`, so this is essentially (and should be named) a `BaseWebSocket`

packages/openapi/src/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+	UserActionsApi,
+	Middleware,
+	ResponseContext,
+	ErrorContext,
+	RequestContext,
+	FetchParams,
+} from '../../client/ts'
+
+const httpLogging = false
+const runTests = false
```
LindvedKrvang, line 12: Hopefully I am missing it, but I don't see when this is set true?

meteor/server/api/rest/api.ts
thomasslee97: I see your point, but `setNext` is not an operation on a Part but rather an operation that is applied to a specified Part. The `Part` resource doesn't own the operation, it's a resource consumed by the operation.

Perhaps the URL should just be `/setNextPart` with all Ids moved to the body as the operation arguments?

Julusian: I think that doing things like `/playlists/{playlistId}/activate` would be reasonable, and also `/playlists/{playlistId}/setNextPart`, with the partId in the body. 
And I guess the rest way to clear the nextPart would be to do a DELETE instead of a POST with an id of `null`?

meteor/lib/api/rest.ts
jstarpl: > Highly unsafe, definitely should go, but I think it needs to be replaced with a specifiaction for communicating the editable fields of an adlib action so that the other adlib method can be called with arbitrary (within reason) arguments.

I suppose for the use-case you are describing, some form of identifying AdLibs, on a machine-readable level would be good. Expose AdLib tags maybe? Or externalId?

meteor/server/api/rest/api.ts
thomasslee97: New definitions in https://github.com/nrkno/sofie-core/pull/777/commits/3286b9f3eadaec9dd3c5afc34957a0c300792544 implemented in https://github.com/nrkno/sofie-core/pull/777/commits/66d52dc36c0b2e4b0bcf768261fd35fe9218aa9b

packages/live-status-gateway/src/collections/partInstances.ts
thomasslee97: Renamed in https://github.com/nrkno/sofie-core/pull/777/commits/ce0911fd2b9a01f267c2d186da698ac687c5dcd6

packages/live-status-gateway/src/wsHandler.ts
thomasslee97: Renamed in https://github.com/nrkno/sofie-core/pull/777/commits/fb6235a7b90c2f09f534a1a843aa1d222f6c5466

`WebSocketTopicBase` is extended by topics and provides common functionality via inheritance. `Topic` is an interface that allows us to store/access topics without knowing the class structure.

packages/openapi/src/__tests__/actions.spec.ts
thomasslee97: We don't currently have tests that can run without running up a whole Sofie, something to be resolved soon-ish, in the meantime we're leaving these disabled.

packages/openapi/api/definitions/userActions.yaml
```diff
@@ -0,0 +1,460 @@
...
+            schema:
+              type: object
+              properties:
+                rehearsal:
+                  type: boolean
+      responses:
+        200:
+          $ref: '#/components/responses/postSuccess'
+        412:
+          description: There is already an active Playlist for the studio that the Playlist belongs to.
```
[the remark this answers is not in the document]
thomasslee97: I could do with some direction here. It'd be nice to return `404` if the requested Playlist doesn't exist, and `500` for errors within Sofie, but I'm not sure how to work out why a call throws. One option is to pre-check for errors that should throw `412` or `404` before the call into Sofie itself, but that feels very brittle, slow, and like a lot of code duplication.

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,660 @@
...
+		triggerWriteAccess()
+
+		return ClientAPI.responseSuccess({ version: CURRENT_SYSTEM_VERSION })
+	}
+	async activate(
+		rundownPlaylistId: RundownPlaylistId,
+		rehearsal: boolean,
+		connection: Meteor.Connection
+	): Promise<ClientAPI.ClientResponse<void>> {
+		return ServerClientAPI.runUserActionInLogForPlaylistOnWorker(
```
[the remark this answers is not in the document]
thomasslee97: I don't think this is too much of an issue, it's a few lines of code used to call the same internal API as the useractions API, any "real" code is hidden within that call and shared anyway. Plus, if the shape of the internal API changes, we may want to change this handler to turn the data in the API request into the data the internal APIs want.

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,660 @@
...
+import { RundownBaselineAdLibActions } from '../../../lib/collections/RundownBaselineAdLibActions'
+import { BucketAdLibs } from '../../../lib/collections/BucketAdlibs'
+import { UserError, UserErrorMessage } from '@sofie-automation/corelib/dist/error'
+import { StudioContentWriteAccess } from '../../security/studio'
+import { ServerPlayoutAPI } from '../playout/playout'
+import { TriggerReloadDataResponse } from '../../../lib/api/userActions'
+
+const REST_API_USER_EVENT = 'rest_api'
+
+class ServerRestAPI extends MethodContextAPI implements ReplaceOptionalWithNullInMethodArguments<RestAPI> {
```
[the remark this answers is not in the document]
thomasslee97: I've removed the `MethodContextAPI` class, definitely feels cleaner https://github.com/nrkno/sofie-core/pull/777/commits/409f97d2a314bcfd662d0bb0f4c6229ed75948d2

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,660 @@
...
+	const rehearsal = (ctx.req.body as { rehearsal: boolean }).rehearsal
+	logger.info(`koa POST: activate ${rundownPlaylistId} - ${rehearsal ? 'rehearsal' : 'live'}`)
+
+	try {
+		ctx.body = ClientAPI.responseSuccess(
+			await MeteorCall.rest.activate(rundownPlaylistId, rehearsal, makeConnection(ctx))
+		)
+		ctx.status = 200
+	} catch (e) {
+		const errMsg = UserError.isUserError(e) ? e.message.key : (e as Error).message
```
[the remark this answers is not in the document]
thomasslee97: Translation added, would appreciate a sanity check https://github.com/nrkno/sofie-core/pull/777/commits/e41b00527eb5ea83fd50780c5f5f0e663e61094e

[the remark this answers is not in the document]
thomasslee97: I've added the `sofieAPIRequest` function for registering API calls which wraps the inner handler call in a `try ... catch` and handles errors https://github.com/nrkno/sofie-core/pull/777/commits/a29d006de3b6b2604b7a255b085d7acd2560327e

meteor/lib/api/methods.ts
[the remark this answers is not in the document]
thomasslee97: As per our discussion on Slack, I've removed this API from the APIs exposed over DDP https://github.com/nrkno/sofie-core/pull/777/commits/a29d006de3b6b2604b7a255b085d7acd2560327e

meteor/lib/api/rest.ts
thomasslee97: Aye, I think a combination of exposing `externalId` for actions as well as a structure inside of `userDataManifest.editableFields` that is readable by external systems (perhaps json-schema or adaptive cards) would get us most of the way. We'd then need the `executeAdLib` method to be able to receive data for these editable fields, but we then stand a chance of validating calls.

This sounds like a follow-up discussion/issue, due to potential changes to `userDataManifest.editableFields` having an affect on existing Sofie installations.

packages/live-status-gateway/src/liveStatusServer.ts
thomasslee97: `(Rundown)AdLibPieces` added https://github.com/nrkno/sofie-core/pull/777/commits/e05e7bda47f27a81743e7980dfc91a1fe90f6bcc

packages/live-status-gateway/api/schemas/activePlaylist.yaml
```diff
@@ -0,0 +1,107 @@
...
+            type: boolean
+            default: false
+        required: [id, name]
+        additionalProperties: false
+      - type: 'null'
+    examples:
+      - id: 'H5CBGYjThrMSmaYvRaa5FVKJIzk_'
+        name: 'Intro'
+        autoNext: false
+  adLibAction:
```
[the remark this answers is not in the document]
thomasslee97: Renamed to `adLib` https://github.com/nrkno/sofie-core/pull/777/commits/948beb9bbf26260614584c6e846fb83a895126d3

packages/live-status-gateway/api/schemas/activePlaylist.yaml
```diff
@@ -0,0 +1,107 @@
...
+    properties:
+      id:
+        description: Unique id of the AdLibAction
+        type: string
+      name:
+        description: The user defined AdLibAction name
+        type: string
+      sourceLayer:
+        description: The source layer name for this AdLibAction
+        type: string
```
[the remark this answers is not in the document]
thomasslee97: `outputLayer` added https://github.com/nrkno/sofie-core/pull/777/commits/cb0457ea6cc45902baad3a5251670c748ab0ed34

meteor/server/api/rest/api.ts
thomasslee97: I've made a method that constructs a string with some useful info. Open to suggestions for better formats for this string. https://github.com/evs-broadcast/sofie-core/blob/66d52dc36c0b2e4b0bcf768261fd35fe9218aa9b/meteor/server/api/rest/api.ts#L42

meteor/lib/api/rest.ts
```diff
@@ -0,0 +1,204 @@
...
+	): Promise<ClientAPI.ClientResponse<void>>
+	/**
+	 * Performs a take in the given Playlist.
+	 *
+	 * Throws if spcified Playlist is not active.
+	 * Throws if specified Playlist does not have a next Part.
+	 * @param rundownPlaylistId Target Playlist.
+	 * @param connection Connection data including client and header details
+	 */
+	take(rundownPlaylistId: RundownPlaylistId, connection: Meteor.Connection): Promise<ClientAPI.ClientResponse<void>>
```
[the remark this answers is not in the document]
thomasslee97: Added as an optional body parameter https://github.com/nrkno/sofie-core/pull/777/commits/13491e62b9510724fbb6be4eb9441be79543da23

meteor/lib/api/rest.ts
thomasslee97: I've clarified the wording, I think it's ok to say that this throws if a next Part couldn't be found because in that case we have failed to perform the requested action/the requested action is not valid? https://github.com/evs-broadcast/sofie-core/blob/66d52dc36c0b2e4b0bcf768261fd35fe9218aa9b/packages/openapi/api/definitions/playlists.yaml#L146

thomasslee97:
Hrmm github will only let me request a review from one person at a time, apologies for notification spam. This is ready for another review.

Julusian (approved):
I haven't read everything, but enough to get an overview

Julusian:
> Hrmm github will only let me request a review from one person at a time, apologies for notification spam. This is ready for another review.

Odd, I can't either. I was expecting that would be a limitation of not being a member of the repository...

Oh well, I've requested from Jan and we can keep passing it on

meteor/server/api/rest/api.ts
LindvedKrvang: My main point was that you have made a "REST" API, but you don't follow the conventions of REST. So it's basically just an API. If you want to continue to call it a REST API, you should follow the REST conventions

meteor/lib/api/rest.ts
thomasslee97: Julian made a similar point, my reply here https://github.com/nrkno/sofie-core/pull/777#discussion_r1085159421

meteor/lib/api/rest.ts
thomasslee97: I've added methods for clearing sourcelayers / recalling sticky Pieces https://github.com/nrkno/sofie-core/pull/777/commits/57c07b14883054732eeec54fb137f6b125d652dc

packages/openapi/api/definitions/userActions.yaml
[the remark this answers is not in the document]
thomasslee97: I've attempted to improve the situation https://github.com/nrkno/sofie-core/pull/777/commits/f55bbcc44f9f5ed55d543fb9fe935fa798c72c07

I'm not 100% happy with this solution as it feels like it's starting to reach into the depths of Sofie a bit, but it is repeating an existing pattern present for the `ClientResponseSucces` whereby a status code can be specified.

I had a chat with Julian about this before putting together this solution, we touched on the fact that there's (by my count) four different error types within Sofie:
- User Errors
- Client Response Errors
- Meteor Errors
- JS Errors

Perhaps this is something for further work in the future? This solution at least gives _something_ meaningful without too much code / behaviour change inside of Sofie.

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,1494 @@
...
+	ctx.type = 'application/json'
+	const server = new ServerRestAPI()
+	const response = ClientAPI.responseSuccess(await server.index())
+	ctx.body = JSON.stringify({ status: response.success, result: response.result })
+	ctx.status = response.success
+	await next()
+})
+
+sofieAPIRequest<never, never, string[]>('get', '/playlists', new Map(), async (serverAPI, connection, event, _params, _body) => {
+	logger.info(`koa GET: playlists`)
```
jstarpl, line 997: I guess this is petty, but I don't think this should say `koa`, since that's an implementation detail that will be likely unfamiliar to someone reading these logs.

meteor/server/api/rest/api.ts
```diff
@@ -0,0 +1,1494 @@
...
+		logger.info(`koa DELETE: sourceLayer ${playlistId} ${sourceLayerId}`)
+
+		check(playlistId, String)
+		check(sourceLayerId, String)
+		return await serverAPI.clearSourceLayer(connection, event, playlistId, sourceLayerId)
+	}
+)
+
+sofieAPIRequest<{ playlistId: string; sourceLayerId: string }, never, void>(
+	'post',
```
jstarpl, line 1196: Why is this a `post` when _take_ and _executeAdLib_ is a `put`? All of these perform the same _sort_ of action, I think.

packages/live-status-gateway/src/collections/adLibActions.ts
```diff
@@ -0,0 +1,72 @@
...
+export class AdLibActionsHandler
+	extends CollectionBase<AdLibAction[]>
+	implements Collection<AdLibAction[]>, CollectionObserver<Map<PartInstanceName, DBPartInstance | undefined>>
+{
+	public observerName: string
+	private _core: CoreConnection
+	private _curRundownId: string | undefined
+
+	constructor(logger: Logger, coreHandler: CoreHandler) {
+		super('AdLibActionHandler', 'adLibActions', logger, coreHandler)
```
jstarpl, line 19: Should these be using the `CollectionName` enum from `@sofie-automation/corelib/dist/dataModel/Collections` ?

packages/live-status-gateway/src/collections/adLibActions.ts
```diff
@@ -0,0 +1,72 @@
...
+		const prevRundownId = this._curRundownId
+		this._curRundownId = data ? unprotectString(data.get(PartInstanceName.current)?.rundownId) : undefined
+
+		await new Promise(process.nextTick.bind(this))
+		if (!this._collection) return
+		if (prevRundownId !== this._curRundownId) {
+			if (this._subscriptionId) this._coreHandler.unsubscribe(this._subscriptionId)
+			if (this._dbObserver) this._dbObserver.stop()
+			if (this._curRundownId) {
+				this._subscriptionId = await this._coreHandler.setupSubscription(this._collection, {
```
jstarpl, line 44: I don't think the Publication name and the Collection name are strictly related. I think, by convention we've set up this so that there's a Publication by the name of the Collection, but I'm not sure if this should be reused like that.

jstarpl (approved):

meteor/server/api/rest/api.ts
thomasslee97: This has been updated to `put`

meteor/server/api/rest/api.ts
thomasslee97: I've changed the logs to log e.g. `API GET: ...` so we can search logs for `API [METHOD]`

packages/live-status-gateway/src/collections/adLibActions.ts
thomasslee97: I've moved to using `CollectionName` and having separate `publication` and `collection` for their separate uses. It would be nice to have publication names exposed in a similar way to `CollectionName` (unless I've missed where they're exposed?) but I think exposing them is outside the scope of this PR because it touches a lot of Sofie code
 https://github.com/nrkno/sofie-core/pull/777/commits/6f0742863b0f042fc9f30caea2534bb7fb08d40c

baltedewit (approved):

jstarpl:
@thomasslee97 I see that there are still some CI issues to be sorted out, specifically tests for corelib, job-worker, shared-lib, etc. fail now for Node 14, which I guess is a requirement until we are able to upgrade Node in meteor to something newer. Could you look into making that compatible with Node 14 or figuring out a way where they would be skipped if Node 14 is being used so that they don't cause a problem?

thomasslee97:
> @thomasslee97 I see that there are still some CI issues to be sorted out, specifically tests for corelib, job-worker, shared-lib, etc. fail now for Node 14, which I guess is a requirement until we are able to upgrade Node in meteor to something newer. Could you look into making that compatible with Node 14 or figuring out a way where they would be skipped if Node 14 is being used so that they don't cause a problem?

I don't think there's a good way to make the `openapi` package support versions before node 16, because even with node 16 we're relying on the `--experimental-fetch` flag. Instead I've made the `openapi` package be ignored for node 14. This should be fine for now as the `openapi` package only serves as a way of running tests outside of CI (against a running Sofie) or for consumers of the API to provide types, so I think it's ok to say that any consumer of the `openapi` package must be using at least node 16?

packages/shared-lib/src/peripheralDevice/peripheralDeviceAPI.ts
```diff
@@ -72,6 +72,7 @@ export enum PeripheralDeviceCategory {
 	PLAYOUT = 'playout',
 	MEDIA_MANAGER = 'media_manager',
 	PACKAGE_MANAGER = 'package_manager',
+	API = 'api',
```
Julusian, line 75: I wonder if this should be API or LIVE_STATUS? 
Do we expect/want other gateway types to exist under this category without needing any changes to support them in meteor?

packages/package.json
```diff
@@ -25,7 +26,7 @@
...
+		"license-validate": "yarn sofie-licensecheck --allowPackages \"caniuse-lite@1.0.30001448;mos-gateway@$(node -p \"require('mos-gateway/package.json').version\");playout-gateway@$(node -p \"require('playout-gateway/package.json').version\");sofie-documentation@$(node -p \"require('sofie-documentation/package.json').version\");@sofie-automation/corelib@$(node -p \"require('@sofie-automation/corelib/package.json').version\");@sofie-automation/shared-lib@$(node -p \"require('@sofie-automation/shared-lib/package.json').version\");@sofie-automation/job-worker@$(node -p \"require('@sofie-automation/job-worker/package.json').version\");lunr-languages@1.10.0;live-status-gateway@$(node -p \"require('live-status-gateway/package.json').version\")\"",
```
Julusian, line 29: Not expected to be part of this PR, someone should really change this into a js file at some point. this is a nightmare to read and work with.

packages/openapi/src/httpLogging.ts
```diff
@@ -0,0 +1,48 @@
...
+
+class TestError extends Error {
+	override name: 'TestError' = 'TestError' as const
+	constructor(msg: string) {
+		super(msg)
+	}
+}
+
+export default class Logging implements Middleware {
+	_logging: boolean
```
Julusian, line 12:
```suggestion
	private readonly _logging: boolean
```

meteor/server/api/rest/api.ts
Julusian, line 1: This file is pretty long, it would be nice to split it up a bit if possible

Julusian (approved):

meteor/server/api/rest/api.ts
Julusian: fine with me :)

packages/shared-lib/src/peripheralDevice/peripheralDeviceAPI.ts
thomasslee97: I've renamed this to `LIVE_STATUS`, given that e.g. package manager has its own category I'm happy to have a special category until we find any commonality.

packages/package.json
thomasslee97: I agree, but I'm not going to address it in this PR

nytamin (changes requested):

packages/openapi/api/actions.yaml
```diff
@@ -0,0 +1,84 @@
...
+  license:
+    name: 'MIT License'
+    url: 'http://opensource.org/licenses/MIT'
+servers:
+  - url: http://localhost:3000/api/v1.0
+    description: Initial release of UserActions API - http - for use during development only!
+tags:
+  - name: userActions
+    description: Provides the ability to control the Sofie application
+paths:
```
nytamin, line 15: Regarding all api-endpoints that returns lists of strings, for example `GET /devices`, `GET /playlists`, `GET /studios` etc.

I think they should instead return a list of objects: `{id: string}[]`, this makes it possible to extend the endpoints with more data at a later time without breaking the API. (an example is when extending to `{id: string, name: string}[]`)

packages/openapi/api/actions.yaml
```diff
@@ -0,0 +1,84 @@
...
+    $ref: 'definitions/playlists.yaml#/resources/playlists'
+  /playlists/{playlistId}/activate:
+    $ref: 'definitions/playlists.yaml#/resources/activate'
+  /playlists/{playlistId}/deactivate:
+    $ref: 'definitions/playlists.yaml#/resources/deactivate'
+  /playlists/{playlistId}/executeAdLib:
+    $ref: 'definitions/playlists.yaml#/resources/executeAdLib'
+  /playlists/{playlistId}/moveNextPart:
+    $ref: 'definitions/playlists.yaml#/resources/moveNextPart'
+  /playlists/{playlistId}/moveNextSegment:
```
nytamin, line 45: Using hyphens ("-") and lowercase in the URLs instead of camelCase seems to be recommended over at [restfulapi.net](https://restfulapi.net/resource-naming/)
```
/playlists/{playlistId}/moveNextSegment -> /playlists/{playlistId}/move-next-segment
```
should we change the URLs to reflect that?

nytamin:
I've given this a review now, focusing on documentation and maintainability (I didn't touch on the logic itself).

I think this looks very very good in general! `Requesting changes` pending my notes above regarding the API endpoints.

nytamin:
I also created a [mirror PR](https://github.com/nrkno/sofie-core/pull/913), to get SonarCloud to do a scan.
Please have a look at the short list of issues here: https://sonarcloud.io/summary/new_code?id=nrkno_sofie-core&pullRequest=913

Not all of them need to be fixed, but it often gives hints to potential issues.

packages/openapi/api/actions.yaml
jstarpl: This is a good point. This actually made me think: should we rather wrap the collections into a wrapper object? Something like?
```json
{
"devices": [
{ "id": "123" },
{ "id": "345" }
]
}
```

A wrapper object would allow us to potentially also provide metadata about _the data collection_ in a forwards-compatible way - such as cursors for paging, if the need ever arises.

jstarpl: Scratch what I wrote. I've played with this definition in: [ReDoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/evs-broadcast/sofie-core/draft/sofie-api/packages/openapi/api/actions.yaml#tag/devices/operation/device) (also, is there a reason why this is using Swagger - which does require JRE - as opposed to a nodejs-based tool?), and now I see that there already is a wrapper object around the response, so that could carry metadata. I think @nytamin 's point still stands, though. Returning an Array of Objects would be better than an Array of Strings, for future proofing. One could imagine for example adding a `$href` property to the objects with links to `/devices/123`.

packages/openapi/api/definitions/devices.yaml
```diff
@@ -0,0 +1,162 @@
...
+                      - id
+                      - name
+                      - status
+                      - messages
+                      - deviceType
+                      - connected
+        404:
+          description: The requested device does not exist.
+  deviceAction:
+    put:
```
jstarpl, line 102: Should this be a `PUT` or a `POST`? https://restfulapi.net/rest-api-design-tutorial-with-example/ says that `PUT` should be idempotent while `POST` is not. I guess, generally, executing actions on devices is not?
```suggestion
    post:
```

packages/openapi/api/definitions/playlists.yaml
```diff
@@ -0,0 +1,574 @@
...
+            type: string
+      responses:
+        200:
+          $ref: '#/components/responses/putSuccess'
+        404:
+          $ref: '#/components/responses/playlistNotFound'
+        500:
+          $ref: '#/components/responses/internalServerError'
+  executeAdLib:
+    put:
```
jstarpl, line 92: Same as above: should this be a `post`? Executing adLibs is specifically not idempotent.
```suggestion
    post:
```

packages/openapi/api/definitions/playlists.yaml
```diff
@@ -0,0 +1,574 @@
...
+                  status:
+                    type: number
+                    example: 412
+                  message:
+                    type: string
+                    example: AdLib could not be found!
+        500:
+          $ref: '#/components/responses/internalServerError'
+  moveNextPart:
+    put:
```
jstarpl, line 142: Like above, maybe should be `post`?
```suggestion
    post:
```

packages/openapi/api/definitions/playlists.yaml
```diff
@@ -0,0 +1,574 @@
...
+                  status:
+                    type: number
+                    example: 412
+                  message:
+                    type: string
+                    example: The selected part does not exist
+        500:
+          $ref: '#/components/responses/internalServerError'
+  moveNextSegment:
+    put:
```
jstarpl, line 198:
```suggestion
    post:
```

packages/openapi/api/definitions/playlists.yaml
```diff
@@ -0,0 +1,574 @@
...
+                  status:
+                    type: number
+                    example: 412
+                  message:
+                    type: string
+                    example: The selected part does not exist
+        500:
+          $ref: '#/components/responses/internalServerError'
+  take:
+    put:
```
jstarpl, line 397:
```suggestion
    post:
```

packages/openapi/api/definitions/playlists.yaml
```diff
@@ -0,0 +1,574 @@
...
+                  status:
+                    type: number
+                    example: 412
+                  message:
+                    type: string
+                    example: Rundown must be active!
+        500:
+          $ref: '#/components/responses/internalServerError'
+    sticky:
+      put:
```
jstarpl, line 479:
```suggestion
      post:
```

packages/openapi/api/definitions/studios.yaml
```diff
@@ -0,0 +1,556 @@
...
+                properties:
+                  status:
+                    type: number
+                    example: 200
+        404:
+          $ref: '#/components/responses/studioNotFound'
+        500:
+          $ref: '#/components/responses/internalServerError'
+  studioAction:
+    put:
```
jstarpl, line 322: This also doesn't seem like an idempotent action, so maybe this should be `POST`?
```suggestion
    post:
```

jstarpl (changes requested):

jstarpl:
> I don't think there's a good way to make the `openapi` package support versions before node 16, because even with node 16 we're relying on the `--experimental-fetch` flag. Instead I've made the `openapi` package be ignored for node 14. This should be fine for now as the `openapi` package only serves as a way of running tests outside of CI (against a running Sofie) or for consumers of the API to provide types, so I think it's ok to say that any consumer of the `openapi` package must be using at least node 16?

Yeah, I think it's perfectly fine to just make it be ignored for Node 14.

thomasslee97:
I've addressed most of the above points, I believe all that's left is to provide an example of how to use the Live Status Gateway and then resolving conflics. I'll see how far I get the rest of today, but will aim to get both of these done by the end of Monday.

packages/openapi/install_swagger.js
```diff
@@ -0,0 +1,47 @@
...
+
+	const srcPath =
+		'https://repo1.maven.org/maven2/io/swagger/codegen/v3/swagger-codegen-cli/3.0.34/swagger-codegen-cli-3.0.34.jar'
+	const swaggerFilename = 'swagger-codegen-cli.jar'
+	await fs
+		.access(`jars/${swaggerFilename}`, fs_constants.R_OK)
+		.catch(async () => get(srcPath, `jars/${swaggerFilename}`))
+}
+
+checkInstall()
```
nytamin, line 47:
```suggestion
checkInstall().catch(console.error)
```

nytamin (approved):
I'm happy with this now!

jstarpl (approved):

PR #778, opened 2022-10-17
feat: move system storePath to an environment variable

Julusian:

feature

The storePath used for snapshots is stored in the database.

The storePath is really an environment level configuration. It is chosen at the time of first deploying the docker images (it has to be bound in as a volume), and so does not make sense to have as a dynamic configuration in the database.

It is now an environment variable instead. This lets it be defined by the person/tool running the docker commands, who have already made the decision about where the storepath will reside.  
At startup, if this variable is not defined, it will cause sofie to crash with a message `SOFIE_STORE_PATH must be defined to launch Sofie` 

For development, if the variable is not set, then a path `.meteor/local/sofie-store` will be used instead. This simplifies setting up a development environment, and should keep impact to developers to a minimum.

* **Other information**:

**Status**


nytamin (approved):
Nice change!

~Should we also~ We should also add a check for this in `systemStatus.ts`, so that a system admin is made aware if it is missing from the env-variables.

Julusian:
We don't need a system status check, as it will explicitly crash at startup with a helpful error https://github.com/nrkno/sofie-core/blob/feat/system-store-path-as-env-var/meteor/server/coreSystem.ts#L54 https://github.com/nrkno/sofie-core/blob/feat/system-store-path-as-env-var/meteor/server/coreSystem.ts#L480-L482

This is consistent with how other environment variables are handled (MONGO_URL, ROOT_URL).

I went with a crash at startup as I think that is a better convention to follow for an environment variable that is definitely wrong/missing. This makes it clearer who is responsible for fixing it (startup crash saying something isnt setup = ops, systemstatus = sofie-admin). 
When deploying this as part of r48, is it reasonable to expect an ops person to read through the 15+ systemstatus warnings to check for one which is relevant to them, which they presumably arent expecting so dont know what to look for?

PR #788, opened 2022-10-25
feat: stronger ddp call typings in gateways

Julusian:

feat

This improves the type safety of some gateway code. Method calls are now typed, using the same api interface as is used by core.

Various other tidying was done to make bits more type safe where possible, using types in `shared-lib`.

The `studioOfDevice` publication has been removed, as it was used solely by playout-gateway to verify that the studio existed, before attempting to load the RoutedTimeline and RoutedMappings from the subscribed collections. This check has no purpose, as if the studioId was invalid, then meteor would have served an empty collection.
I have checked, and the tv2 inews gateway does not appear to use this publication.

* **Other information**:

I have given a quick test of both playout and mos gateways and they appear to be happy. A more thorough test is needed

**Status**


packages/mos-gateway/src/mosHandler.ts
```diff
@@ -97,7 +97,7 @@ export class MosHandler {
 
 		const peripheralDevice = await coreHandler.core.getPeripheralDevice()
 
-		this._settings = peripheralDevice.settings || {}
+		this._settings = peripheralDevice.settings as any
```
nytamin, line 100: Could you explain the reasoning for this change? Casting to `any` isn't exactly in the spirit of making stronger typings, I think

Julusian: It's because the core version of these settings
https://github.com/nrkno/sofie-core/blob/release47/packages/shared-lib/src/core/model/peripheralDevice.ts#L28-L32
has nothing in common with the mos-gateway version
https://github.com/nrkno/sofie-core/blob/release47/packages/mos-gateway/src/mosHandler.ts#L36-L51

I don't think the core type should know about all these properties, that is the point of the config manifest after all, so I think there is something larger wrong with how we handle config for these devices that causes us to have weird/mismatched typings. Too much to deal with in the scope of this PR

nytamin (approved):

PR #789, opened 2022-10-26
fix: update ErrorBoundrary message to be more user-friendly

nytamin:
This PR changes the visual appearance of the ErrorBoundrary, with the goal of being friendlier and more descriptive to the End-User.

Before:

image

After:

image

### DIFF
--- a/meteor/client/lib/ErrorBoundary.tsx
+++ b/meteor/client/lib/ErrorBoundary.tsx
@@ -1,194 +1,148 @@
 import * as React from 'react'
 
 interface IState {
-	hasError: boolean
-	error?: Error
-	info?: React.ErrorInfo
-	expandedStack?: boolean
-	expandedComponentStack?: boolean
+	error: {
+		error: Error
+		info: React.ErrorInfo
+	} | null
 }
 
 export class ErrorBoundary extends React.Component<{}, IState> {
-	static style = {
-		box: {
+	static style: { [key: string]: React.CSSProperties } = {
+		container: {
+			position: 'relative',
 			display: 'block',
-			position: 'static',
-			margin: '0',
+			margin: '10px',
 			padding: '10px',
-			fontSize: '10px',
-			lineHeight: '1.2em',
-			fontFamily: 'Roboto, sans-serif',
-			fontWeight: 300,
-			width: '100%',
+			// fontSize: '12px',
+			// lineHeight: '1.2em',
+
+			// fontWeight: 300,
+			textDecoration: 'none',
+			// width: '100%',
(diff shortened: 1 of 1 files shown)


meteor/client/lib/ErrorBoundary.tsx
```diff
@@ -1,194 +1,123 @@
...
+		container: {
+			position: 'relative',
 			display: 'block',
-			position: 'static',
-			margin: '0',
+			margin: '10px',
 			padding: '10px',
-			fontSize: '10px',
-			lineHeight: '1.2em',
-			fontFamily: 'Roboto, sans-serif',
```
jstarpl, line 20: Actually, these `fontFamily` definitions are put here on purpose. We don't know in which context the ErrorBoundry may be inclined to display, therefore we may unknowingly inherit CSS properties that aren't sensinble (say, `lineHeigh` or `height` or `background`) and can make the entire warning message unreadable. I believe we should still keep these and not rely on any inherited properties.

jstarpl (changes requested):

meteor/client/lib/ErrorBoundary.tsx
nytamin: I slightly disagree actually.

I would argue that we _should_ inherit the fontSize-related properties. I think it is reasonable to assume that any inherited values will be chosen to be readable. - From personal experience this was the case previously, where `fontSize: '10px'` made the text quite unreadable on my high-dpi screen, while the other content was OK.

 
We do (kind of) know the context of the ErrorBoundrary - it will only be where we're using the `<ErrorBoundrary>` component.
I think we should consider the case of where the context makes the error unreadable to be an edge-case. But as an remedy for this, we could add a ´className="error-boundrary"´ and fix the edge-case in the css-place that causes the issue. Not great, but doable I think.

jstarpl: Well, but that then means that whoever is using `ErrorBoundary` needs to understand the styling CSS context, without any way to see what it will look like, other than purposefully adding some way of crashing the component for testing, which I don't think is a reasonable expectation. If anything, we are using it 

If the `10px` is a problem for a high DPI screen, then this could simply be changed to, I don't know `1rem`, which will result in almost all cases in the same font size as is _probably_ the one you are seeing with inheritance. I would strongly argue against using any CSS inheritance in this case: this is supposed to be a set-and-forget safety net.

nytamin: All good points! I will make it so 👍

jstarpl (approved):

PR #790, opened 2022-10-27
Feat/settings UI use overrides

Julusian:

feature

In release47, various types in the database were changed to use an `ObjectWithOverrides<T>` helper, so that we could in the future differentiate between user and blueprint owned settings values.  
In the original change, the settings UI was updated to continue modifying the defaults.

This changes the settings ui to modify these `ObjectWithOverrides<T>` by using the 'overrides' functionality.

The overrides we generate are quite simple for now, I expect this to be refined as we start to use it and want it to be more granular.

During this I started to split out some of the control types from EditAttribute, so that it is possible to use these without the many layers of wrapping and abstraction that EditAttribute applies. I don't know if this was a good idea to do?  
For these controls I made some additional wrappings for use in these `ObjectWithOverrides<T>` areas. These feel very repetetive and should be simplified, but I am unsure on how best to achieve that.

* **Other information**:

I have tried to make as many things reusable as possible. Some helpers have been written to aid this.

I am wondering if this might cause some issues for blueprint migrations, as they are only able to observe the `defaults` block. If this is a problem, we could look at flattening the overrides into the defaults before running these migrations. This will make these ui changes feel weird, as running the migrations will cause the overrides to no longer be overrides.  
This would only be temporary until the next task of replacing the blueprints migrations system.

**Status**


jstarpl (changes requested):
A good first version, I think. One thought I have, is that the Default/Reset functionality seems like a wrapper around the actual InputControl - could that be made into it's own component so that we can easily modify how that feature works/looks like?I think maybe the whole Label + Hint + Default + Reset could be actually made into some sort of a wrapper component?

Julusian:
@jstarpl how do you mean?

Something like this?
```
<ControlWithOverrideForObject ... >
  {(value, handleUpdate) => <CheckboxControl value={value} handleUpdate={handleUpdate} />}
</ControlWithOverrideForObject>
```
I think this works fine, apart from handling the difference for checkboxes where we want the label to go after rather than before.

Or should the inside of `CheckboxControlWithOverrideForObject` be more like this?
```
return (
  <label>
    <CheckboxControl ... />
    {label}
    <ControlDefaultsAndReset ... />
  </label>
)
```

jstarpl (changes requested):
I've got some comments, mainly, I think, if `BlueprintConfigManifestSettings gets a 1166-line rewrite, it would be good to use that time to break it up into individual files, to avoid replicating the massive file of `ConfigManifestSettings`.

meteor/client/ui/Settings/BlueprintConfigManifestSettings.tsx
```diff
@@ -0,0 +1,1166 @@
+import ClassNames from 'classnames'
```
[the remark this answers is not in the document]
Julusian: I wouldn't say `ground-up` here, this was a combination of a heavy refactor to functional react, (should be mostly in https://github.com/nrkno/sofie-core/pull/790/commits/e1e758e5c631e4e88a3106423359efcd57f94508), and then additional changes for the new data flow.
From the combination of these and renaming the file, github doesn't understand the relation anymore...

But I have now split it out to multiple files, in logical chunks. Not quite one file per component, but almost

meteor/client/ui/Settings/BlueprintConfigManifest/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+	const deleteRef = useRef<DeleteItemModalRef>(null)
+
+	const addItem = useCallback(() => {
+		if (addRef.current) {
+			addRef.current.show()
+		}
+	}, [])
+	const showDelete = useCallback((item: ConfigManifestEntry) => {
+		if (deleteRef.current) {
+			deleteRef.current.show(item)
```
jstarpl, line 130: This is a bit of an unusual setup, I guess either this could just be component state (i.e. "deleteDialogItem") where there would be no need for this imperative method, or this could be just using `doModalDialog`, which allows an imperative style and accepts callbacks for user decisions.

jstarpl (approved):

meteor/client/ui/Settings/BlueprintConfigManifest/index.tsx
Julusian: I don't know why I did that model this way, I've changed it to doModalDialog.

PR #798, opened 2022-11-04
fix: checkPieceContentStatus messages not being translated

Julusian:

bug fix

Media object status issues are always presented in english, even when the ui is set to another language which has translations.
The translations exist in the system, but they are not used

The translations are now used.

These status messages are passed untranslated or modified from `checkPieceContentStatus` through to the component which will render them, where the translations are then applied.

* **Other information**:

This also includes some small refactoring of `checkPieceContentStatus`  to make it a bit more managable

I also found that there was a function `getMinimumReactivePieceNotesForPart` which was converting the result of `checkPieceContentStatus` into `PartNote`, and passed through the ui with the rest of the notes from the segment. These notes were simply being counted for display in the segment headers, so this has all been simplified to pass simply those counts around and avoid the extra processing being done to these notes which are never used.

**Status**


Julusian:
There are a few `// TODO - translate` inside of `checkPieceContentStatus` now, for lines line `addFrameWarning(mediaObject.mediainfo.blacks, 'black')`, where we used to translate black before passing it in. 
I don't know how best to get that translated

meteor/client/ui/SegmentContainer/withResolvedSegment.ts
```diff
@@ -100,10 +100,15 @@ export interface IProps {
...
+export interface SegmentNoteCounts {
+	criticial: number
+	warning: number
+}
+
 export interface ITrackedProps {
 	segmentui: SegmentUi | undefined
 	parts: Array<PartUi>
-	segmentNotes: Array<SegmentNote>
+	segmentNotes: SegmentNoteCounts
```
baltedewit, line 111: I think it would be more explanatory if this interface property was also renamed

meteor/client/ui/SegmentList/SegmentListHeader.tsx
```diff
@@ -45,7 +44,7 @@ export function SegmentListHeader({
 	playlist: RundownPlaylist
 	studio: UIStudio
 	parts: Array<PartUi>
-	segmentNotes: Array<SegmentNote>
+	segmentNotes: SegmentNoteCounts
```
baltedewit, line 47: Same rename here as above

meteor/lib/mediaObjects.ts
```diff
@@ -169,557 +169,599 @@ export interface ScanInfoForPackage {
...
+														type,
+														count: frames,
+													}
+												),
+											})
+										}
+									}
+								}
+								if (!piece.content.ignoreBlackFrames && mediaObject.mediainfo.blacks?.length) {
+									addFrameWarning(mediaObject.mediainfo.blacks, 'black') // TODO - translate
```
baltedewit, line 414: To fix the translations you could wrap the calls to generateTranslation as `generateFrameWarningTranslation(type, message, args)` and move the type from the args into the string. But since we are moving away from mediaObjects slowly I'm ok with leaving it untranslated, maybe check that with Rannveig though?

baltedewit (approved):

meteor/lib/mediaObjects.ts
Julusian: I don't think the i18n tooling will handle that, which will result in no translations being produced for them.
The same flaw exists above/below for expected-packages.

I have an idea on how to solve it. by passing an object of the 'complete' strings into this function instead

baltedewit: Your updated solution looks good 👍 (and along the lines of what I was thinking of)

PR #804, opened 2022-11-18
Bug fix: change how to wait for other Fibers in collections.*Async

nytamin:
Bug fix

If for example `collection.findFetchAsync` throws instantly (like if providing an invalid argument) `makeReady` rejects right away. Since there is a `sleep(0)` in `collection.findFetchAsync`, the promise is therefore unhandled, thus causing Core to crash.

I've moved the sleep into `makePromise` and ensured that the resulting promise is handled from the start.

* **Other information**:

**Status**

### DIFF
--- a/meteor/lib/__tests__/lib.test.ts
+++ b/meteor/lib/__tests__/lib.test.ts
@@ -1,3 +1,5 @@
+import '../../__mocks__/_extendJest'
+
 import { Meteor } from 'meteor/meteor'
 import { Mongo } from 'meteor/mongo'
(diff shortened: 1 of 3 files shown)


Julusian (approved):
This sounds like it could be a larger problem for us based on how you were triggering this crash.. 

There are many places where we do something similar of create a promise which we will await after doing a little more stuff, and it sounds like all of those could also trigger a crash.

Hopefully there is an eslint rule we can enable to tell about these promises that we aren't awaiting/catching immediately, so that we can find all these places

PR #807, opened 2022-11-21
Use default values from config manifest if no value has been configured

LindvedKrvang:
Bug/Missing feature

In the config manifest there is an option to specify a default value for each configuration entry. This default value seems not to be used.

When sending device options to each TSR device the default value from the config manifest will now be applied if there hasn't been configured any value. 
The configuration field in the UI displays a text "Defaults to '{defaultValue}' if left empty" if there is a default value specified in the config manifest.
The EditAttributeText now trims its value so strings like " " is no longer possible.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/lib/EditAttribute.tsx
+++ b/meteor/client/lib/EditAttribute.tsx
@@ -229,7 +229,11 @@ const EditAttributeText = wrapEditAttribute(
 			this.handleEdit(event.target.value)
 		}
 		handleBlur(event) {
-			this.handleUpdate(event.target.value)
+			let value: string = event.target.value
+			if (value) {
+				value = value.trim()
+			}
+			this.handleUpdate(value)
 		}
 		handleEscape(event) {
 			const e = event as KeyboardEvent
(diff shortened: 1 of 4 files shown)


baltedewit (changes requested):
I would strongly advise to iterate over the `PLAYOUT_SUBDEVICE_CONFIG` in `configManifest.ts` once at startup to create an object `defaultOptions` of the following type `{ [deviceType: DeviceType]: Record<string, defaultOption> }`. Then in `populateDefaultValuesIfMissing` you can simply return `{ ...defaultOptions[deviceOptions.type], ...deviceOptions  }`

LindvedKrvang:
> I would strongly advise to iterate over the `PLAYOUT_SUBDEVICE_CONFIG` in `configManifest.ts` once at startup to create an object `defaultOptions` of the following type `{ [deviceType: DeviceType]: Record<string, defaultOption> }`. Then in `populateDefaultValuesIfMissing` you can simply return `{ ...defaultOptions[deviceOptions.type], ...deviceOptions }`

Updated

I had to modify your suggestion for `populateDefaultValuesIfMissing` a little since we need to populate on `deviceOptions.options`.
Also, empty strings counted as a value so had to do an ugly filter of them. Any suggestions to make that filter less ugly is much appreciated

LindvedKrvang:
Closing this draft again since the changes will be merged to the TV2 core

PR #808, opened 2022-11-21
feat: Blueprints migrations replacement

Julusian:

feature

Builds on #790 and #772 

The current blueprint migrations work, but are error prone. It is very easy to end up with fighting over values, and very hard to get an overview of what has changed.

This adds a new flow to the blueprints, intended to replace the current migrations system. The migrations are being left in place untouched for now, but it is intended that it will be removed in a couple of releases time.

The new flow works as a two stage process. 

1) `validateConfig`. This is a step where the blueprints are given the full blob of blueprintConfig, and are expected to return an array of messages for the user. This should do some sanity checks on the data, reporting any errors or potential misconfigurations that there are in the config block. The result of this can be ignored, the user does not have to resolve these messages if they do not wish to.

2) `applyConfig` This step takes the full blob of blueprintConfig and generates the new 'default' values for the `ObjectWithOverrides` blocks that this step owns. The results will be saved to the 'defaults' properties with very little modification by sofie.

Neither of these are provided with the previous values, or given any knowledge on what was an override in the blueprintConfig. It is intended that the output of both of these steps are predictable and stable.  
The intention is that this should be a very predictable flow. The same config after 5 release of sofie and matching blueprints changes will give identical data to a fresh installation with the same blueprints and config.  
For NRK, this will help us keep everything in sync, as changes made in live systems should be propogated back into the blueprints to be auto generated in a future version.

--------

To achieve this, the methods described above have been added to both the showstyle and studio blueprints (with different return types).

Additionally, a concept of 'config presets' has been introduced for blueprints. Instead of having the config be part of the configManifest for the blueprint, it can instead provide multiple 'config presets'. The settings for both ShowStyle and Studio have a new dropdown available to select the preset to use.  
The intention is to allow for different variations of the config to be exposed. This is particularly useful for two studios which share a blueprint but want some configuration defaults to be different.
ShowStyleVariants also have these 'config presets' which exist as a child of the config preset of their ShowStyleBase.  
This new flow is not capable of automatically creating ShowStyleVariants (yet?), this must be done manually, but is as simple as creating it, filling in the name and choosing the config-preset, so hopefully isnt a burden to those managing deployments.

image

And a new section has been added to the 'Database Upgrade' page of the settings ui (it probably needs some styling/layout/thought work)

image

The result of `validateConfig`:
image

When sofie has determined that there is a change that requires `applyConfig` to be re-run, it can provide details on what changed:
image

* **Other information**:

Upgrading blueprints to this new flow can be slow work. For NRK, there were multiple different configs that needed consolidating, with differing migrations and new properties had to be added to control what was created for each studio.

**Status**


Julusian:
Some of the systemStatus tests are failing, because getting them to pass requires setting up the mocks to be fully upgraded, and it is not simple to do that. I could either spend the time getting everything properly configured for this/every test (which will become some hundreds of lines of code) or mock the `getUpgradeSystemStatusMessages()` method. I don't know which would be best

baltedewit (commented):
API changes for the blueprint manifest and contexts look good to me.

I didn't look at the UI code _too_ closely.

PR #810, opened 2022-11-29
TV2 Sync: Merge current develop into R49

jstarpl:

This PR merges changes from the current _develop_ branch at TV2 Denmark into the current NRK Release

* **Other information**:

**Status**


meteor/client/ui/Shelf/Keyboard/KeyboardPreview.tsx
```diff
@@ -27,22 +27,6 @@ import { translateMessage } from '@sofie-automation/corelib/dist/TranslatableMes
 
 const _isMacLike = navigator.platform.match(/(Mac|iPhone|iPod|iPad)/i) ? true : false
 
-declare global {
```
LindvedKrvang, line 30: You don't want the files in metoer/ui/Shelf/Keyboard
In the gitignore there should be some comments that need to be uncommented so you don't get these file.

meteor/client/ui/globals/keyboardGlobals.d.ts
```diff
@@ -0,0 +1,14 @@
+declare type KeyboardLayoutMap = Map<string, string>
```
LindvedKrvang, line 1: You do want this file though

meteor/yarn.lock
```diff
@@ -9832,10 +9832,10 @@ timecode@0.0.4:
   resolved "https://registry.yarnpkg.com/timecode/-/timecode-0.0.4.tgz#cdea9d2352009700e50da657cb7fb7c1720f5704"
   integrity sha512-xFVZlWnqls5eVphtUJo0UuXeJa0grBmxESSn5QPQB7CMsaVgTnLdP9PTZqb4KyFmLdrE+PUu3eEt42zHsMxCtw==
 
-"timeline-state-resolver-types@npm:@tv2media/timeline-state-resolver-types@2.2.2":
-  version "2.2.2"
-  resolved "https://registry.yarnpkg.com/@tv2media/timeline-state-resolver-types/-/timeline-state-resolver-types-2.2.2.tgz#10c11881673efc1bfa1ef6ba88dee7364dd2f42a"
-  integrity sha512-/sxyHUr+wmIp9FiI+CGKnr+isc0p3vFJOzZZW8Pq8iqM3QPo+kR26U7hybIlO+tO32PN95eUANx4fFjZBkKaiA==
+"timeline-state-resolver-types@npm:@tv2media/timeline-state-resolver-types@2.2.3":
```
LindvedKrvang, line 9835: I don't think you want our TSR repo

packages/blueprints-integration/package.json
```diff
@@ -39,7 +39,7 @@
 	],
 	"dependencies": {
 		"@sofie-automation/shared-lib": "link:../shared-lib",
-		"timeline-state-resolver-types": "npm:@tv2media/timeline-state-resolver-types@2.2.3",
+		"timeline-state-resolver-types": "npm:@tv2media/timeline-state-resolver-types@2.3.1",
```
LindvedKrvang, line 42: I think this should be pointing at your TSR repo

packages/playout-gateway/package.json
```diff
@@ -62,7 +62,7 @@
 		"debug": "^4.3.3",
 		"fast-clone": "^1.5.13",
 		"influx": "^5.9.3",
-		"timeline-state-resolver": "npm:@tv2media/timeline-state-resolver@2.2.2",
+		"timeline-state-resolver": "npm:@tv2media/timeline-state-resolver@2.3.1",
```
LindvedKrvang, line 65: I think you want to be pointing at your TSR repo

meteor/client/ui/Settings/ShowStyle/VariantSettings.tsx
```diff
@@ -103,21 +291,82 @@ export const ShowStyleVariantsSettings = withTranslation()(
 			})
 		}
 
-		renderShowStyleVariants() {
+		VariantItem = ({ index, showStyleVariant, moveVariantHandler }) => {
```
jstarpl, line 294: This is a unusual design pattern - Is there a particular reason why this needs to be a member of `ShowStyleVariantsSettings` class? It seems like it could just accept some additional props and a callback and it wouldn't have to be a member. Also, extracting this function from a class member into a separate file & component would help with reducing the size of the `VariantSettings` file.

LindvedKrvang: @RasmusAlbrektsen

Julusian: I agree that this mixing of style (class vs function react component) is unexpected. In my opinion a file should only contain either class or functional components.
In addition, this has lead to some of the newly added callbacks which are passed into child components to be written using `useCallback` and some to be members of the parent class

RasmusAlbrektsen: This can largely be attributed to my unfamiliarity to React and how the flows work here. The intention was to make it easy to later retract from the file and divide into separate files to easier make it generic for a later task down the road. But again, it is mostly because i had issues with figuring out how to make it generic at this point in time. 
I will gladly take any advice on how this could be done better, i'm still very much learning when it comes to how the UI is put together!

meteor/client/ui/Settings/ShowStyle/VariantSettings.tsx
```diff
@@ -169,26 +420,90 @@ export const ShowStyleVariantsSettings = withTranslation()(
...
-					<table className="table expando settings-studio-showStyleVariants-table">
-						<tbody>{this.renderShowStyleVariants()}</tbody>
-					</table>
-					<div className="mod mhs">
-						<button className="btn btn-primary" onClick={this.onAddShowStyleVariant}>
-							<FontAwesomeIcon icon={faPlus} />
-						</button>
-					</div>
+					<h2 className="mhn">{t('Show Style Variants')}</h2>
+					<DndProvider backend={HTML5Backend}>
```
Julusian, line 473: This `DndProvider` is unnecessary and could cause issues. One is already registered at the root level of the application.

meteor/server/api/showStyles.ts
```diff
@@ -75,50 +75,94 @@ export async function insertShowStyleBaseInner(organizationId: OrganizationId |
...
+export async function createDefaultShowStyleVariant(
+	context: MethodContext | Credentials,
+	showStyleBaseId: ShowStyleBaseId,
 	name?: string
 ): Promise<ShowStyleVariantId> {
-	return ShowStyleVariants.insertAsync({
+	await assertShowStyleBaseAccess(context, showStyleBaseId)
+
+	const highestRank = ShowStyleVariants.find(
+		{},
```
nytamin, line 106: As a general rule; never do full-db-searches, this is a classic source for performance problems!
This should be limited to `{ showStyleBaseId }`.

meteor/server/api/showStyles.ts
```diff
@@ -75,50 +75,94 @@ export async function insertShowStyleBaseInner(organizationId: OrganizationId |
...
 	const access = await ShowStyleContentWriteAccess.anyContent(context, showStyleBaseId)
 	const showStyleBase = access.showStyleBase
 	if (!showStyleBase) throw new Meteor.Error(404, `showStyleBase "${showStyleBaseId}" not found`)
-
-	return insertShowStyleVariantInner(showStyleBase, name)
 }
-export async function insertShowStyleVariantInner(
-	showStyleBase: ShowStyleBaseLight,
+
+export async function createDefaultShowStyleVariant(
```
nytamin, line 98: This might be personal opinion / nag, but what was wrong with the old name `insertShowStyleVariant`?

The new name differs much more from it's neighbor `insertShowStyleBase`, which gives me the impression that it does something considerably different, which it's not.

So the functionality of the function basically hasn't changed, so why change the name?

meteor/server/api/showStyles.ts
```diff
@@ -75,50 +75,94 @@ export async function insertShowStyleBaseInner(organizationId: OrganizationId |
 		_rundownVersionHash: '',
 	}
 	ShowStyleBases.insert(showStyleBase)
-	await insertShowStyleVariantInner(showStyleBase, 'Default')
+
+	const showStyleVariant: ShowStyleVariant = {
```
nytamin, line 79: This could be DRY:ed by instead calling `createDefaultShowStyleVariant(context, showStyleBase._id, 'Default')`.

Also, the default rank is set to 500 here, but is 0 in `createDefaultShowStyleVariant`.

meteor/server/api/showStyles.ts
```diff
@@ -75,50 +75,94 @@ export async function insertShowStyleBaseInner(organizationId: OrganizationId |
...
 	await Promise.allSettled([
-		ShowStyleBases.removeAsync(showStyleBase._id),
+		ShowStyleBases.removeAsync(showStyleBaseId),
 		ShowStyleVariants.removeAsync({
-			showStyleBaseId: showStyleBase._id,
+			showStyleBaseId: showStyleBaseId,
 		}),
 		RundownLayouts.removeAsync({
-			showStyleBaseId: showStyleBase._id,
+			showStyleBaseId: showStyleBaseId,
```
nytamin, line 161: minor note: `{ showStyleBaseId: showStyleBaseId }` can be shortened to `{ showStyleBaseId }`

meteor/server/api/showStyles.ts
```diff
@@ -132,18 +176,48 @@ export async function removeShowStyleVariant(
 	await ShowStyleVariants.removeAsync(showStyleVariant._id)
 }
 
+export async function reorderAllShowStyleVariants(
+	context: MethodContext,
+	showStyleBaseId: ShowStyleBaseId,
+	orderedVariants: ShowStyleVariant[]
```
nytamin, line 182: No need to send the full data set over the wire here, sinve only the `._id` is used.
This should be refactored to `ShowStyleVariantId[]` instead.

meteor/server/api/showStyles.ts
```diff
@@ -132,18 +176,48 @@ export async function removeShowStyleVariant(
...
+): Promise<void> {
+	await assertShowStyleBaseAccess(context, showStyleBaseId)
+	await reassignShowStyleVariantIndexes(orderedVariants)
```
nytamin, line 185: There are a few problems that should be addressed here:

1. Technically, we haven't checked that the ShowStyleVariants actually exist inside the showStyleBaseId here. We should probably to that.
2. What about if not all of the variants in a ShowStyleBase are provided? This needs to be ensured to avoid settings strange `_rank`s.

meteor/server/api/showStyles.ts
```diff
@@ -75,50 +75,94 @@ export async function insertShowStyleBaseInner(organizationId: OrganizationId |
...
+	showStyleVariant: ShowStyleVariant
+): Promise<ShowStyleVariantId> {
+	await assertShowStyleBaseAccess(context, showStyleVariant.showStyleBaseId)
+
+	return insertShowStyleVariant(showStyleVariant)
 }
+
+export async function importShowStyleVariantAsNew(
+	context: MethodContext | Credentials,
+	showStyleVariant: ShowStyleVariant
```
nytamin, line 139: Minor note: Since the `_id` is overwritten anyway, `Omit<ShowStyleVariant, '_id'>` is a slightly better argument.

meteor/client/ui/Settings/ShowStyleBaseSettings.tsx
```diff
@@ -78,9 +78,16 @@ export default translateWithTracker<IProps, IState, ITrackedProps>((props: IProp
...
-			? ShowStyleVariants.find({
-					showStyleBaseId: showStyleBase._id,
-			  }).fetch()
+			? ShowStyleVariants.find(
+					{
+						showStyleBaseId: showStyleBase._id,
+					},
+					{
+						sort: {
+							_rank: 1,
```
nytamin, line 87: Since it currently is possible in `importShowStyleVariant` to add variants with duplicate `_rank`-values, we need to add a secondary argument for sorting:
```
sort: {
  _rank: 1,
  _id: 1,
}
```

meteor/server/migration/X_X_X.ts
```diff
@@ -12,5 +13,26 @@ import { CURRENT_SYSTEM_VERSION } from './currentSystemVersion'
...
+		canBeRunAutomatically: true,
+		validate: () => {
+			return (
+				ShowStyleVariants.find({
+					_rank: { $exists: false },
+				}).count() > 0
+			)
+		},
+		migrate: () => {
+			ShowStyleVariants.find({
```
nytamin, line 27: It would have been nice with a comment here that we just assign any random rank for the variants, since they didn't have any previous order.

packages/corelib/src/dataModel/ShowStyleVariant.ts
```diff
@@ -4,6 +4,9 @@ import { ShowStyleVariantId, ShowStyleBaseId } from './Ids'
 
 export interface DBShowStyleVariant {
 	_id: ShowStyleVariantId
+
+	_rank: number
```
nytamin, line 8: Add a description of this property:
```
/** A number used to sort the variants within their ShowStyleBase. */
```

nytamin (changes requested):

meteor/server/api/showStyles.ts
Julusian: Adding to that, it is loading the full document of every showstyle variant, when it could be doing a findOne, and only fetching the `_rank` field of the matching document:
```
ShowStyleVariants.findOneAsnc({
  showStyleBaseId
}, {
  fields: {
    _rank: 1
  },
  sort: {
    _rank: -1,
  },
})
```

meteor/server/api/showStyles.ts
jstarpl: I agree, I don't see the point in the name change, other than a collision with a private `insertShowStyleVariant` method, that didn't do anything beyond calling `ShowStyleVariants.insertAsync()`, which seems superfluous?

PR #811, opened 2022-12-01
fix: in the event of multiple tags they should all have the correct name

siljekristensen:
Fixes the issue where if there are multiple tags, only the first image name is "fully qualified" and can be pushed to the correct repository.

* **Other information**:

**Status**

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -107,13 +107,25 @@ jobs:
         run: |
           echo "Enable build and push: ${{ env.SECRET_ACCESS != '' }}"
           echo "enable=${{ env.SECRET_ACCESS != '' }}" >> $GITHUB_OUTPUT
-      - name: Get the Docker tag
-        id: docker-tag
+      - name: Get the Docker tag for GHCR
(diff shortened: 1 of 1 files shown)


.github/workflows/node.yaml
```diff
@@ -107,13 +107,25 @@ jobs:
         run: |
           echo "Enable build and push: ${{ env.SECRET_ACCESS != '' }}"
           echo "enable=${{ env.SECRET_ACCESS != '' }}" >> $GITHUB_OUTPUT
-      - name: Get the Docker tag
+      - name: Get the Docker tag for GHCR
         id: docker-tag
```
Julusian, line 111: should this be called `ghcr-tag` to clarify its purpose?

.github/workflows/node.yaml
```diff
@@ -225,13 +237,25 @@ jobs:
...
```
Julusian, line 241: should this be called `ghcr-tag` to clarify its purpose?

siljekristensen: Sure!

PR #814, opened 2022-12-05
Ci/update gha workflows

PeterC89:
Updates GHA workflows to remove depreciated calls

Workflows run successfully

Workflows continue to run successfully with the same behaviour

* **Other information**:

**Status**

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -82,7 +82,11 @@ jobs:
         id: image-tags
         run: |
           IMAGES=
-          DOCKER_TAG=${{ github.ref_name }}
+          if [[ ${{ github.ref_name }} == "master" || ${{ github.ref_name }} == "main" ]]
+            DOCKER_TAG=latest
+          else
+            DOCKER_TAG=${{ github.ref_name }}
+          fi
           # check if a release branch, or master, or a tag
           if [[ $DOCKER_TAG =~ ^release([0-9]+)$ || $DOCKER_TAG == "latest" || "${{ github.ref }}" == refs/tags/* ]]
           then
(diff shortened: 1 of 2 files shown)


.github/workflows/node.yaml
```diff
@@ -175,17 +174,16 @@ jobs:
...
       - name: Determine images to publish
         id: image-tags
         # TODO - image needs changing...
         run: |
           IMAGES=
-          DOCKER_TAG=${{ steps.docker-tag.outputs.tag }}
+          if [[ ${{ github.ref_name }} == "master" || ${{ github.ref_name }} == "main" ]]
+            DOCKER_TAG=latest
+          else
+            DOCKER_TAG=${{ github.ref_name }}
```
baltedewit, line 185: Are special characters allowed in the docker tag or should it have the same filter applied as elsewhere?

baltedewit (approved):

Julusian:
Silje has already been reworking the workflows in this repository, so I don't think there will be anything left to do here
https://github.com/nrkno/sofie-core/pull/803 https://github.com/nrkno/sofie-core/pull/811

.github/workflows/node.yaml
PeterC89: It appears they are not, but it didn't have it before so I didn't add it, probably should!
The docker website says:
`A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.`

I wonder if we should add a check for length as well then...

siljekristensen (approved):

siljekristensen:
> > Silje has already been reworking the workflows in this repository, so I don't think there will be anything left to do here #803 #811
> 
> Ah whoops, I thought all that work had been merged, I'll check!

I hadn't touched this workflow, so these changes are needed :)

jstarpl (approved):

PR #820, opened 2022-12-14
fix: update mos-connection dependency

nytamin:
This PR changes the dependency of `mos-connection` to `@mos-connection/connector`, following the mono-repoification in [this PR](https://github.com/nrkno/sofie-mos-connection/pull/62).

* **Other information**:

**Status**


packages/mos-gateway/src/coreHandler.ts
```diff
@@ -465,18 +474,21 @@ export class CoreMosDeviceHandler {
 	 * @param o the object to convert
 	 */
 	private fixMosData(o: any): any {
-		if (_.isObject(o) && (o instanceof MosTime || o instanceof MosDuration || o instanceof MosString128)) {
```
Julusian, line 468: Didnt we discus this fixMosData function and decide to remove it?

packages/mos-gateway/src/mosHandler.ts
```diff
@@ -205,6 +216,8 @@ export class MosHandler {
 		if (!this._settings.mosId) throw Error('mosId missing in settings!')
 		connectionConfig.mosID = this._settings.mosId
 
+		connectionConfig.strict = this._settings.strict
```
Julusian, line 219: should this also recreate `this.mosTypes`?

packages/shared-lib/src/peripheralDevice/mos.ts
```diff
@@ -1,15 +1,15 @@
...
-export type IMOSROFullStory = any
-export type IMOSROReadyToAir = any
-export type IMOSROStory = any
-export type IMOSRunningOrder = any
-export type IMOSRunningOrderBase = any
-export type IMOSRunningOrderStatus = any
-export type IMOSStoryAction = any
-export type IMOSStoryStatus = any
-export type MosString128 = any
+// export type IMOSItem = any
```
Julusian, line 3: delete this file instead?

packages/corelib/package.json
```diff
@@ -44,7 +44,7 @@
 		"fast-clone": "^1.5.13",
 		"i18next": "^21.9.1",
 		"influx": "^5.9.3",
-		"mos-connection": "^2.0.1",
+		"@mos-connection/model": "3.0.0-nightly-feat-mono-repo-20221216-111645-0e02d70.0",
```
Julusian, line 47: do we need this dependency here? it is also included in shared-lib, so perhaps can be accessed via there?

Julusian (approved):
I think that the mos types should also be re-exported by blueprints-integration (like we do for tsr types already). Its already going to be there as a dependency, so it may as well be exposed (blueprints-integration -> shared-lib -> @mos-connection/model).

packages/mos-gateway/src/mosHandler.ts
nytamin: not needed here, since `this._settings` hasn't changed

packages/corelib/package.json
nytamin: Hm, but that would be an "extraneous import" wouldn't it?

Julusian: I mean make shared-lib re-export it (like https://github.com/nrkno/sofie-core/blob/release49/packages/shared-lib/src/tsr.ts).

We do have the ci job which will make sure all versions of `@mos-connection/model` being used are the same, which should stop that for corelib. But if the blueprints have to define `@mos-connection/model` explicitly, then there will be a high risk of it using two different versions

packages/mos-gateway/src/coreHandler.ts
nytamin: Note to self: we did discuss it, but (I) decided to (try to) not break-everything in this PR

PR #821, opened 2022-12-14
fix(EditAtribute): Number(0) isn't shown, because it's replaced with String("")

jstarpl:

Bug fix

It's impossible to see the value `0`, since it's being replaced by `""`.

Instead of using `||` as fallback, the Int and Float inputs use a `??` operator, which allows the `0` value to show.

* **Other information**:

(cherry picked from commit 6616b7be6586654b8c9a913682f1b372cbbabd0c)

**Status**


Julusian (approved):
LGTM

PR #823, opened 2022-12-15
feat: TSR device debug state

thomasslee97:
Adds the ability for Core to fetch a representation of the state of a Playout Gateway sub-device.

**What is the current behavior?**
New feature.

**What is the new behavior?**
To aid the the process of debugging new TSR integrations, as well as changes to blueprints, it can be useful to know what state a TSR integration is currently in. The existing timeline view in Core shows the state that is intended, but a TSR integration may need to translate a timeline state into a different representation in order to produce commands to achieve the desired state.

This PR adds the ability for a developer to request a view of this alternate representation from the Playout Gateway, allowing a developer to check whether a TSR device is correctly understanding a timeline, and helping developers to understand what happens to a timeline after it leaves Core.

For example:

debug-state-status-page

The exact format and contents of this "debug state information" is up to the individual integration, however, for an object in this format:

```
{
    "pgm_switcher": {
        "input": 1,
        "transition": "mix"
    },
    "pvw_switcher": {
        "input": 2,
        "transition": "cut"
    }
}
```

The UI will display the object in the following format:

| | |
|---|---|
|pgm_switcher.input|1|
|pgm_switcher.transition|"mix"|
|pvw_switcher.input|2|
|pvw_switcher.transition|"mix"|

It is also up to each integration how/when this state is emitted. It could be emitted periodically or after every update to the timeline for example.

Debug states are emitted from all device integrations that wish to make this information available, no changes are required to existing integrations until the reporting of this information is desired. The emitted states are then collected by the TSR conductor and are held in memory - one state held per device.

Core can then request these states from Playout Gateway, which will retrieve the states from TSR. The state is only stored for a given device if two conditions are met:
- `Activate State Debugging` is checked in the Playout Gateway settings.
- `Activate debug logging for device` is checked for the device in question.

Likewise, a call to Playout Gateway to fetch this information is only made if the user is on either the System Status page or the Playout Gateway settings page, and the aforementioned settings are checked.

As such, impact on production is minimal as additional calls will not be made unless explicitly opted into. Further, it is recommended that the `debugState` device setting is checked before emitting any state in integrations, so in a production environment a single additional `if` statement will be hit and no debug state will be emitted - so no additional memory will be used to hold debug state information and no additional work will be performed to generate this information.

debug-state-playout-gateway

The implementation follows closely the pattern used in the Package Manager for getting the status of expectations for the status page.

**Other information**:

Related PR against TSR: https://github.com/nrkno/sofie-timeline-state-resolver/pull/234

**Status**


jstarpl:
The TSR side of this PR has already been merged: https://github.com/nrkno/sofie-timeline-state-resolver/pull/234

packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -655,14 +657,31 @@ export class TSRHandler {
...
+			}
+
+			const debug: boolean = this.getDeviceDebugState(options)
+			if (device.debugState !== debug) {
+				this.logger.info(`Setting debugState of device ${device.deviceId} to ${debug}`)
+				debugLoggingPs.push(device.setDebugState(debug))
+			}
+		}
 		await Promise.all(debugLoggingPs)
+		await Promise.all(ps)
```
baltedewit, line 674: We already await this on line 615 so it seems this line should not be necessary?

baltedewit (approved):
I'm happy with the changes in the playout gateway. Core looks good to me as well, but I'm not the expert there.

meteor/client/ui/Status/SystemStatus.tsx
```diff
@@ -182,6 +186,48 @@ export const DeviceItem = reacti18next.withTranslation()(
 			}
 		}
 
+		private getDebugStateTableBody(debugState: object) {
```
Julusian, line 189: This might be better as being its own component, to make it more reusable. It would also allow for memoizing this computation

Julusian (commented):
There is a push to convert some of the ui polling into custom publications. Im not sure if that should be done here as it is a debug tool and isnt part of the rundown view, so is unlikely to be open in more than one place and is ok to be slower.

meteor/lib/api/userActions.ts
```diff
@@ -394,6 +399,8 @@ export enum UserActionAPIMethods {
 	'switchRouteSet' = 'userAction.switchRouteSet',
 
 	'disablePeripheralSubDevice' = 'userAction.system.disablePeripheralSubDevice',
+
+	'getDebugStates' = 'userAction.playout.getDebugStates',
```
jstarpl, line 403: I don't think this should be `userAction.playout.getDebugStates`. This is not a "playout" action, as far as the user is concerned. Maybe this should be a `userAction.debug.*`?

jstarpl (changes requested):

packages/playout-gateway/src/tsrHandler.ts
thomasslee97: Good spot, removed

thomasslee97:
I have moved the `getDebugStates` function under the `SystemStatusAPI` namespace given that's where the methods are being used. Perhaps better?
I've also created a component for drawing the table.

jstarpl (approved):
Thanks, I think this looks a lot cleaner!

meteor/client/ui/Status/DebugState.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react'
+import { Translated } from '../../lib/ReactMeteorData/react-meteor-data'
+import * as reacti18next from 'react-i18next'
+
+interface IDebugStateTableProps {
+	debugState: object
+}
+interface IDebugStateTableState {}
+
+export const DebugStateTable = reacti18next.withTranslation()(
```
Julusian, line 10: I would prefer this to be a functional/hooks component, but that doesnt need to be done

Julusian (approved):
looks good :)

meteor/client/ui/Status/DebugState.tsx
jstarpl: Yeah, me too, but I don't think that's a reason to block this.

Julusian:
@thomasslee97 could you fix the merge conflict? once that is done I shall merge it

PR #827, opened 2023-01-03
fix: refactor playout-gateway types into shared-lib

Julusian:

Feature

Some types are copied between playout-gateway and corelib/meteor, as there is nowhere that they can be placed due to the dependency on tsr/tsr-types.

The tsr-types dependency has been moved from blueprints-integration to shared-lib. While this does mean that some gateways (ingest/package-manager) will have the dependency unnecessarily, it is a small no-dependency library so has little impact on size.
This lets us move some more types into shared-lib, for better reuse across the sofie components.

* **Other information**:

While some changes were made to blueprints-integration typings, this has no impact to blueprints, as it is simply changing where tsr-types were imported from, and is keeping the existing re-export.

**Status**


nytamin (approved):
LGTM!

PR #829, opened 2023-01-09
fix: Queued parts at the end of a playlist

thomasslee97:
Bug fix

When queuing a Part at the end of a Playlist, timing information for this Part is not reported. This results in the UI drawing the queued Part and the Playhead incorrectly, in some cases resulting in a Part that is not visible to the user.

If a Part is marked as an `adlib-part` orphan and its parent Segment is found but it cannot otherwise be placed in the list of Parts, the Part is added to the end of the list.

**Other information**:
If R49 is closed for contributions then I will rebase onto R50 when a branch is available.

**Status**


jstarpl (approved):
Looks good to me. Since it's a bugfix, and very limited in scope at that, I don't see a reason not to merge it.

PR #832, opened 2023-01-11
Fix: No retry when calling storeRundownSnapshot

nytamin:
Bug fix

There was an issue where the client retried sending the `storeRundownSnapshot`-command many times, due to an initial network issue (and after that Core crashed, so the client sent even more retries..)

No retries are done when calling `storeRundownSnapshot`

* **Other information**:


jstarpl:
I wonder if the solution to this is to limit this at the client, or if it makes more sense to maybe add some sort of a semi-one-time-use token (like the current timestamp) to the snapshot method and make the subsequent calls invalid and return an error?

nytamin:
> I wonder if the solution to this is to limit this at the client, or if it makes more sense to maybe add some sort of a semi-one-time-use token (like the current timestamp) to the snapshot method and make the subsequent calls invalid and return an error?

You have a point, [the Meteor Guide states that](https://guide.meteor.com/methods.html#retries)
>in the case where an extra Method call could have negative consequences it is worth putting in extra effort to ensure that Methods are idempotent - that is, calling them multiple times doesn’t result in additional changes to the database.

However I think that if we where to follow that, we should add a one-time-use token for all UserActions, to ensure that they all are only executed once. - For example, if a Take is done while a client is excperiencing connection-troubles, several Takes might happen on the server..

jstarpl:
> > in the case where an extra Method call could have negative consequences it is worth putting in extra effort to ensure that Methods are idempotent - that is, calling them multiple times doesn’t result in additional changes to the database.
> 
> However I think that if we where to follow that, we should add a one-time-use token for all UserActions, to ensure that they all are only executed once. - For example, if a Take is done while a client is excperiencing connection-troubles, several Takes might happen on the server..

Well, takes actually carry the currentPartInstanceId for that purpose, so all the subsequent calls will fail. I suppose the adLib and moveNext methods are vulnerable to that, since they don't carry any context information.

Julusian:
This has been long merged into r44 and has been merged into newer releases from there

PR #833, opened 2023-01-13
Fix: Hack for slowness in release 44

nytamin:
This PR contains a hack to fix an issue we've seen where Core (paired with Package Manager and usage of Buckets) becomes very slow.

Currently, there is one adlib per showstyle-variant created per adlib, so if there are 20 variants, and 20 items in a bucket, there will be 400 adlibs (and 400 expectedPackages) created.

We believe that the 400 expectedPackages causes PM to send 800 PackageInfo updates (1 scan + 1 deepScan per package) to Core, and all of those are subscribed to by the web clients, causing both clients and server to become slow.

This PR adds a hack that only generates one (1) expectedPackage per _unique_ adlib-in-bucket. 

One consequence of this hack is that zebra-stripes are likely to show up in the GUI for the items in a bucket, that is something we'll have to address in a proper long-term fix later.

### DIFF
--- a/meteor/server/api/ingest/packageInfo.ts
+++ b/meteor/server/api/ingest/packageInfo.ts
@@ -7,7 +7,12 @@ import { runIngestOperation } from './lib'
 import { IngestJobs } from '@sofie-automation/corelib/dist/worker/ingest'
 
 export function onUpdatedPackageInfo(packageId: ExpectedPackageId, _doc: PackageInfoDB | null) {
-	logger.info(`PackageInfo updated "${packageId}"`)
+	let logStr = ''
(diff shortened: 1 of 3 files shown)


jstarpl (changes requested):
I think that if we push this hack to production, it needs to also disable the zebra stripes in the GUI for buckets. Otherwise, we'll mess our users' minds too much.

nytamin:
Closing this for now, as we don't need the hack. We solved the issueusing the `allvariants` property instead: https://github.com/nrkno/sofie-core/blob/release44/packages/blueprints-integration/src/action.ts#L91

PR #834, opened 2023-01-17
feat: Expose peripheralDeviceActions to blueprints

PeterC89:
peripheralDeviceActions API exposed to blueprints

Blueprints can't use actions

Blueprints can use actions

* **Other information**:

**Status**

### DIFF
--- a/packages/blueprints-integration/src/context.ts
+++ b/packages/blueprints-integration/src/context.ts
@@ -15,7 +15,9 @@ import {
 	IBlueprintSegmentRundown,
 } from './rundown'
 import { BlueprintMappings } from './studio'
-import { OnGenerateTimelineObj } from './timeline'
+import { OnGenerateTimelineObj, TSR } from './timeline'
+import { PeripheralDevicePublic } from '@sofie-automation/shared-lib/dist/core/model/peripheralDevice'
+import { PeripheralDeviceId } from '@sofie-automation/shared-lib/dist/core/model/Ids'
 
 /** Common */
 
@@ -229,6 +231,14 @@ export interface IActionExecutionContext
 	// updateAction(newManifest: Pick<IBlueprintAdLibActionManifest, 'description' | 'payload'>): void // only updates itself. to allow for the next one to do something different
 	// executePeripheralDeviceAction(deviceId: string, functionName: string, args: any[]): Promise<any>
 	// openUIDialogue(message: string) // ?????
+	/** Returns a list of the PeripheralDevices */
+	listPeripheralDevices(): Promise<PeripheralDevicePublic[]>
+	/** Execute an action on a certain PeripheralDevice */
+	executeTSRAction(
+		deviceId: PeripheralDeviceId,
+		actionId: string,
+		payload: Record<string, any>
+	): Promise<TSR.ActionExecutionResult>
 }
 
 /** Actions */
--- a/packages/job-worker/src/blueprints/context/adlibActions.ts
+++ b/packages/job-worker/src/blueprints/context/adlibActions.ts
@@ -13,11 +13,16 @@ import {
 	SomeContent,
 	Time,
 	WithTimeline,
+	TSR,
 } from '@sofie-automation/blueprints-integration'
-import { PartInstanceId, RundownPlaylistActivationId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+import {
+	PartInstanceId,
+	PeripheralDeviceId,
+	RundownPlaylistActivationId,
+} from '@sofie-automation/corelib/dist/dataModel/Ids'
 import { DBRundown } from '@sofie-automation/corelib/dist/dataModel/Rundown'
 import { UserError, UserErrorMessage } from '@sofie-automation/corelib/dist/error'
-import { assertNever, getRandomId, omit } from '@sofie-automation/corelib/dist/lib'
+import { assertNever, Complete, getRandomId, omit } from '@sofie-automation/corelib/dist/lib'
 import { logger } from '../../logging'
 import { ReadonlyDeep } from 'type-fest'
 import { CacheForPlayout, getRundownIDsFromCache } from '../../playout/cache'
@@ -63,6 +68,10 @@ import _ = require('underscore')
 import { ProcessedShowStyleConfig } from '../config'
 import { DatastorePersistenceMode } from '@sofie-automation/shared-lib/dist/core/model/TimelineDatastore'
 import { getDatastoreId } from '../../playout/datastore'
+import { executePeripheralDeviceAction } from '../../peripheralDevice'
+import { PeripheralDevicePublicWithActions } from '@sofie-automation/shared-lib/dist/core/model/peripheralDevice'
+import { PeripheralDevice, PeripheralDeviceType } from '@sofie-automation/corelib/dist/dataModel/PeripheralDevice'
+import { literal } from '@sofie-automation/shared-lib/dist/lib/lib'
 
 export enum ActionPartChange {
 	NONE = 0,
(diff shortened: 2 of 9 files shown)


packages/corelib/src/worker/studio.ts
```diff
@@ -33,6 +33,7 @@ export enum StudioJobs {
 	SetNextPart = 'setNextPart',
 	SetNextSegment = 'setNextSegment',
 	ExecuteAction = 'executeAction',
+	ExecuteTSRAction = 'executeTSRAction',
```
Julusian, line 36: I don't see this as being used anywhere

packages/blueprints-integration/src/context.ts
```diff
@@ -229,6 +229,11 @@ export interface IActionExecutionContext
 	// updateAction(newManifest: Pick<IBlueprintAdLibActionManifest, 'description' | 'payload'>): void // only updates itself. to allow for the next one to do something different
 	// executePeripheralDeviceAction(deviceId: string, functionName: string, args: any[]): Promise<any>
 	// openUIDialogue(message: string) // ?????
+	executeTSRAction(
```
Julusian, line 232: It would be good to have a documentation comment describing this method

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +638,21 @@ export class ActionExecutionContext
...
 	}
+
+	async executeTSRAction(
+		deviceId: string,
+		actionId: string,
+		payload: Record<string, any>
+	): Promise<TSR.ActionExecutionResult> {
+		const device = await this._context.directCollections.PeripheralDevices.findOne({
+			_id: new RegExp(`${deviceId}$`),
+		})
```
jstarpl, line 649: I don't think this is a good idea? If there is a device called `caspar01` and and another one called `backscreen_caspar01`, which one will be selected is undefined. I think this either needs to be an exact match (and the bluprints _just have to know_ what is the correct ID) or we need to have some sort of a middle step where the playoutGateway/mappings deviceId gets translated to a peripheralDeviceId.

jstarpl (changes requested):

packages/job-worker/src/blueprints/context/adlibActions.ts
jstarpl: Maybe collaborate with @baltedewit on how to do the matching?

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +638,21 @@ export class ActionExecutionContext
...
 		return getMediaObjectDuration(this._context, mediaId, this.studioId)
 	}
+
+	async executeTSRAction(
+		deviceId: string,
+		actionId: string,
+		payload: Record<string, any>
+	): Promise<TSR.ActionExecutionResult> {
+		const device = await this._context.directCollections.PeripheralDevices.findOne({
+			_id: new RegExp(`${deviceId}$`),
```
Julusian, line 648: I'm not liking this query. if this was called with a `deviceId` of `"0"`, then it would end up with almost any of the configured devices.
In addition, in an environment where there are playout-gateways for other studios (or ones not attached to a studio), it could pick one of those too.

I don't see a studioId property that you could filter on (that is only on the parent devices), but we do have the parent devices already in memory.
So you could do 
```
const parentIds = this._cache.PeripheralDevices.findAll((d) => d.type === PeripheralDeviceType.PLAYOUT).map(
			(d) => d._id
		)
```
and then add `parentDeviceId: { $in: parentIds },` to the query. That will ensure the selected device belongs to the current studio.

As for improving the regex, it might be possible to determine the full name of it. I'm not certain if this will work reliably enough though. It looks like the parent device is always called `${parentId}PlayoutCoreParent`, and the subdevices are always called `${parentId}Playout${childId}`.
I am not 100% sure this will always work as I have a playout-gateway called `99778B8822A2G448H`, but I think that is because it was from pre-r45 which would sometimes allow a gateway to be called a random id.

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +638,21 @@ export class ActionExecutionContext
...
+		})
+		if (!device) {
```
Julusian, line 650: As you are only intesrested in the `_id` of the of the device, it would be good to use a `projection` on the query, to avoid other properties (such as the large configmanifest) being parsed and loaded into memory

packages/job-worker/src/peripheralDevice.ts
```diff
@@ -8,8 +8,12 @@ export async function executePeripheralDeviceFunction(
 	context: JobContext,
 	deviceId: PeripheralDeviceId,
 	timeoutTime0: number | null,
-	functionName: string,
-	...args: any[]
+	action: {
```
Julusian, line 11: I think this signature makes more sense than the spread operator it used to be.

Perhaps this could be written as `action: Pick<PeripheralDeviceCommand, 'functionName' | 'args' | 'actionId' | 'payload'>`.
That would help ensure the types stay more in sync.

But also, `functionName` and `args` feel like they should be required properties here.
Combining that into the above can be done, but starts to feel messy action: SetRequired<Pick<PeripheralDeviceCommand, 'functionName' | 'args' | 'actionId' | 'payload'>, 'functionName | 'args'>`.

Julusian (changes requested):

packages/job-worker/src/blueprints/context/adlibActions.ts
baltedewit: I agree that we want this to be a strict match. Did you use a RegExp because it expects a Protected String otherwise? We have convenience methods for fixing that.

In addition I think we should add a timeout and perhaps instead of throwing an error we should return an execution result?

packages/job-worker/src/peripheralDevice.ts
baltedewit: if we want to have required properties I think we should separate these into separate methods, because the consumer will use _either_ functionName _or_ actionId.

Julusian (changes requested):
Perhaps we should look at this together tomorrow?

PeterC89:
> Perhaps we should look at this together tomorrow?

Yes sounds good!

nytamin (approved):

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +645,40 @@ export class ActionExecutionContext
...
+				_id: d._id,
+				name: d.name,
+				studioId: d.studioId,
+				category: d.category,
+				type: d.type,
+				subType: d.subType,
+				parentDeviceId: d.parentDeviceId,
+				created: d.created,
+				status: d.status,
+				settings: d.settings,
```
jstarpl, line 672: Wouldn't it make sense to expose `deviceName` from `PeripheralDevice` (https://github.com/nrkno/sofie-core/blob/031e3e3da71a9dc0eb6c4b15215eb375fc5e08ec/packages/corelib/src/dataModel/PeripheralDevice.ts#L31) and the available TSR Actions (from `configManifest`)? I think it would greatly improve on the problem highlighed by @Julusian above.

jstarpl (changes requested):

packages/job-worker/src/blueprints/context/adlibActions.ts
PeterC89: Had a go at this, it's a little messy so suggestions on cleaning that up welcome!

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +645,40 @@ export class ActionExecutionContext
...
+	async listPeripheralDevices(): Promise<PeripheralDevicePublic[]> {
+		const selector: Partial<PeripheralDevice> = {}
+
+		if (this._context.studio.organizationId) {
+			selector.organizationId = this._context.studio.organizationId
+		} else {
+			selector.organizationId = null
+		}
+
+		const devices = await this._context.directCollections.PeripheralDevices.findFetch(selector)
```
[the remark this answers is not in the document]
PeterC89: Think the query is improved now?

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +645,40 @@ export class ActionExecutionContext
 	async hackGetMediaObjectDuration(mediaId: string): Promise<number | undefined> {
 		return getMediaObjectDuration(this._context, mediaId, this.studioId)
 	}
+
+	async listPeripheralDevices(): Promise<PeripheralDevicePublic[]> {
```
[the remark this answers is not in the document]
PeterC89: Both the name and deviceName are now exposed to the blueprints so I guess they could get the first part of the string which should match right?
Seems a little messy, is there a neater way of doing that?

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +646,49 @@ export class ActionExecutionContext
...
+
+		const devices = await this._context.directCollections.PeripheralDevices.findFetch({
+			parentDeviceId: {
+				$in: parentDeviceIds,
+			},
+		})
+
+		return devices.map((d) => {
+			// Only expose a subset of the PeripheralDevice to the blueprints
+			return literal<Complete<PeripheralDevicePublic & { actions: SubdeviceAction[] | undefined }>>({
```
Julusian, line 667: should this `PeripheralDevicePublic & { actions: SubdeviceAction[] | undefined }` type be named and user in the interface definition?

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -636,4 +646,49 @@ export class ActionExecutionContext
...
+				studioId: d.studioId,
+				category: d.category,
+				type: d.type,
+				subType: d.subType,
+				parentDeviceId: d.parentDeviceId,
+				created: d.created,
+				status: d.status,
+				settings: d.settings,
+				actions:
+					d.configManifest.subdeviceManifest &&
```
Julusian, line 680: could this be simplified to `actions: d.configManifest.subdeviceManifest?.[d.subType]?.actions,` ?

Julusian (approved):

packages/job-worker/src/blueprints/context/adlibActions.ts
PeterC89: TypeScript kept complaining when I tried this before but having tried it again it's now happy, I wonder if something wasn't built right

PR #837, opened 2023-01-20
feat!: json schemas for device config and mappings

Julusian:

feature for R50

Depends on https://github.com/nrkno/sofie-timeline-state-resolver/pull/237

Gateways use a custom format for defining the config ui.
For TSR, this information is defined in playout-gateway, causing differences to occur between the manifest and the types in TSR.

Building on #809, this changes gateways to use a json-schema to define their configuration. It is expected that typescript interfaces should be generated from this schema to ensure that it stays in sync too.

* **Other information**:

Support for the old ConfigManifest has been removed for the gateways, but the blueprint variant is still in use for blueprint based things currently. That should be expected to change in a future PR.

Ill admit that the `SchemaForm*` types are a little strange in how they are built on `SchemaFormWithOverrides`. After doing the initial step of device config with a basic `SchemaForm` type, I found that for Mappings which use the `ObjectWithOverrides` structure, it was necessary for a variation of the SchemaForm which supported this (including a variation of the table type). While implementing this I realised that a `SchemaFormWithOverrides` with no default values would behave the same as the basic `SchemaForm` type, so to avoid duplication and similar types which would drift apart in functionality over time I wrote the `SchemaFormForCollection` as a way of achieving this. `SchemaFormForCollection` turned out simpler than I hoped, so I am pretty happy with the result.

**Status**


meteor/client/lib/forms/schemaFormTable.tsx
```diff
@@ -0,0 +1,325 @@
...
+		// Send it onwards
+		overrideHelper.setItemValue(item.id, attr, newObj)
+	}, [schema.items, overrideHelper, rows, item.id])
+
+	const tableOverrideHelper = useMemo(
+		() => new OverrideOpHelperTable(overrideHelper, item.id, rows, attr),
+		[overrideHelper, item.id, rows, attr]
+	)
+
+	const title = schema['ui:title']
```
jstarpl, line 47: I think it would be good to define some sort of an `enum` or object with these special extensions to JSONSchema.

packages/shared-lib/src/core/deviceConfigManifest.ts
```diff
@@ -11,179 +11,59 @@
...
 
+import { TSRActionSchema } from 'timeline-state-resolver-types'
 import { TranslationsBundle } from '../lib/translations'
 
 export interface DeviceConfigManifest {
 	/**
 	 * A description of the config fields
 	 */
-	deviceConfig: ConfigManifestEntry[]
+	deviceConfigSchema: string
```
jstarpl, line 21: Shouldn't this be `JSONBlob<JSONSchema>`?

packages/shared-lib/src/core/deviceConfigManifest.ts
```diff
@@ -11,179 +11,59 @@
...
+		 * Name to display to users for this device type.
+		 * This will be translated if supplied by the device.
+		 */
+		displayName: string
+		/**
+		 * Stringified JSON schema to use for the settings for this device
+		 */
+		configSchema: string
+		/**
+		 * Playout layer-mappings available for this device type
```
jstarpl, line 53: I think a little bit more explanation of how this object works would be good? This is a mapping of what strings onto what?

meteor/client/lib/forms/schemaFormWithOverrides.tsx
```diff
@@ -0,0 +1,281 @@
...
+			if (props.schema.enum) {
+				return <EnumFormWithOverrides {...childProps} />
+			} else {
+				return <IntegerFormWithOverrides {...childProps} />
+			}
+		case TypeName.Number:
+			return <NumberFormWithOverrides {...childProps} />
+		case TypeName.Boolean:
+			return <BooleanFormWithOverrides {...childProps} />
+		case TypeName.String:
```
jstarpl, line 98: The original Config Manifest system supported a multi-line string. I think this too should support that - maybe through a `"ui:multiline": Boolean` key, for example?

meteor/client/lib/forms/schemaFormWithOverrides.tsx
```diff
@@ -0,0 +1,281 @@
...
+				return <ObjectFormWithOverrides {...props} />
+			}
+		case TypeName.Integer:
+			if (props.schema.enum) {
+				return <EnumFormWithOverrides {...childProps} />
+			} else {
+				return <IntegerFormWithOverrides {...childProps} />
+			}
+		case TypeName.Number:
+			return <NumberFormWithOverrides {...childProps} />
```
jstarpl, line 95: I'd like these to respect JSON Schema `multipleOf`, `minimum` and `maximum`, and pass them down to the HTML input elements, but I guess it's not necessarily an MVP thing, since the Config Manifest system did not have that capability either.

jstarpl (changes requested):
A couple of notes on feature parity with the Config Manifest system.

meteor/client/lib/forms/schemaFormWithOverrides.tsx
Julusian: The multiline support we had was producing an array of strings and is now handled at https://github.com/nrkno/sofie-core/blob/feat/gateways-json-schema/meteor/client/lib/forms/schemaFormWithOverrides.tsx#L115-L116

Example use: https://github.com/nrkno/sofie-timeline-state-resolver/blob/feat/config-schema/packages/timeline-state-resolver/src/integrations/vizMSE/%24schemas/options.json#L71-L77

A `ui:multiline` property would make sense, but as it would be new functionality should come later?

baltedewit (approved):

meteor/client/lib/forms/schemaFormWithOverrides.tsx
jstarpl: Ah, okay, I thought that was just Strings with newlines. If that's the case, then yeah, I think we are just aiming at feature parity.

meteor/client/lib/forms/schemaFormWithOverrides.tsx
Julusian: I have implemented minimum and maximum.
I think `multipleOf` requires some more implementation as its doesn't quite line up with the behaviour of step, so shall leave that for later.

packages/shared-lib/src/core/deviceConfigManifest.ts
Julusian: Good idea, I forget we have that.

It does make a bit of a mess in the gateways to translate things across, but I think it is worth it

packages/shared-lib/src/core/deviceConfigManifest.ts
Julusian: This property has changed type to `playoutMappings?: Record<string, JSONBlob<JSONSchema>>`, which hopefully clarifies this?

packages/scripts/schema-types.mjs
```diff
@@ -0,0 +1,57 @@
+import { compileFromFile } from 'json-schema-to-typescript'
+import * as fs from 'fs/promises'
+
+/** ********************************************************
+ *
+ * This script goes through the json-schemas of all devices (located under /$schemas )
+ * and auto-generates types for those schemas
+ *
+ * @todo: auto generate the interfaces.ts in /src as well
```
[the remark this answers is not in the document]
Julusian: yep, I forgot to remove it after copying this file from TSR..

jstarpl:
One note I'd like to make on this PR is that it definitely breaks compatibility with any Peripheral Devices, so I would like to hold off on merging this in until we have a wider discussion on how to handle that, by for example bumping Core (and Playout Gateway, MOS Gateway, and Package Manager and Input Gateway) to 2.0.0

meteor/client/lib/forms/schemaFormWithOverrides.tsx
```diff
@@ -0,0 +1,279 @@
...
+		default:
+			return <>{t('Unsupported array type "{{ type }}"', { type: props.schema.items?.type })}</>
+	}
+}
+
+const ObjectFormWithOverrides = (props: SchemaFormWithOverridesProps) => {
+	return (
+		<>
+			{' '}
+			{Object.entries(props.schema.properties || {}).map(([index, schema]) => {
```
jstarpl, line 126: Shouldn't this also take `additionalProperties` into consideration? I'm looking into how this affects Input Gateway, and have one instance where for one of the SubDevice types, I want to have a configuration of an object type, with user-configurable keys that then contains an object definition for that key that varies on a `type` discriminator property. In that case I used a `ConfigManifestEntryType.TABLE` with a `defaultType`, `typeField` and some `config` type definitions. I don't think this is possible here?

packages/shared-lib/src/core/deviceConfigManifest.ts
jstarpl: Yeah, I think that's enough clarification.

meteor/client/lib/forms/schemaFormWithOverrides.tsx
Julusian: I dont think `additionalProperties` is something the UI should be looking at. https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties
The best we would be able to do if using it is allow for adding custom fields, using `additionalProperties` and `patternProperties` to limit the names/types of those. But they wont have a description or anything, so that wont be a good replication of what you were doing in input gateway.

I forgot to check input-gateway, so didn't realise that complex use of nested tables was being used. It could either be reimplemented using [`oneOf`](https://json-schema.org/understanding-json-schema/reference/combining.html?highlight=oneof#oneof), or that table could be split up into multiple tables to not need the complex types

jstarpl: Yeah, I guess we can live with a multiple tables, I don't expect us to configure the MIDI integration all that often. Although I think we should leave some _TODO_ here, since this is quite a significant limitation, I think.

PR #843, opened 2023-01-27
chore: docs about HOLD and ab playback

Julusian:


packages/documentation/docs/for-developers/for-blueprint-developers/HOLD.md
```diff
@@ -0,0 +1,40 @@
+# HOLD
```
jstarpl, line 1: I feel like this should be just ordinary "Hold". I'm not sure if we should use the all-caps spelling. Maybe @hummelstrand has some opinions?

packages/documentation/docs/for-developers/for-blueprint-developers/HOLD.md
```diff
@@ -0,0 +1,40 @@
...
+
+It is not possible to run any adlibs while a hold is active, or to change the nexted part. Once it is in progress, it is not possible to abort or cancel the HOLD and it must be run to completion. If the second part has an autonext and that gets reached before the HOLD is completed, the HOLD will be treated as completed and the autonext will execute as normal.
+
+When the part to be held is playing, with the correct part as next, the flow for the users is:
+
+- Activate HOLD (By hotkey or other source)
+- Perform a take into the HOLD
+- Perform a take to complete the HOLD
+
+Before the first take in the HOLD, it can be cancelled in the same way it was activated.
```
jstarpl, line 20: I think a sentence describing what is happening to the parts, i.e. (which one is next, which one is current and when etc.) would be good here.

packages/documentation/docs/for-developers/for-blueprint-developers/ab-playback.md
```diff
@@ -0,0 +1,172 @@
...
+- Users get confused if the player of a clip changes without a reason
+- Reloading an already loaded clip can be costly, so should be avoided when possible
+- Adlibbing a clip, or changing what Part is nexted can result in needing to move what player a clip has assigned
+- Postroll or preroll is often needed
+- Some studios can have less players available than ideal. (eg, going back to back between two clips, and a clip is playing on the studio monitor)
+
+## Defining
+
+:::info
+This method of defining sessions is completely optional, it could be done in many other ways depending on how you wish to implement it.
```
jstarpl, line 24: I think "completely optional" may be a bit misleading. The point here is that this is just one of many possible implementations.

```suggestion
This method of defining sessions is just a suggestion, it could be done in many other ways depending on how you wish to implement it.
```

jstarpl (approved):
An excellent addition to our docs. I think I'll try to set up some sort of an illustration of a J-cut, to drive the point home with the use case.

PR #848, opened 2023-02-01
feat: new bucket design

PeterC89:
New design of buckets to improve usability

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/styles/shelf/dashboard.scss
+++ b/meteor/client/styles/shelf/dashboard.scss
@@ -616,4 +616,93 @@ $dashboard-button-height: 5.625em;
 			}
 		}
 	}
+	
+	&.dashboard-panel__panel--bucket {
+		$dashboard-button-width: 10.625em;
+		$dashboard-button-height: 8.90625em;
+		$dashboard-button-label-height: 2.7em;
+		$dashboard-button-thumbnail-height: $dashboard-button-height - $dashboard-button-label-height;
+
+		.dashboard-panel__panel__group {
+	
+			.dashboard-panel__panel__group__liveline {
+				transform: translate(calc(-1 * #{$dashboard-button-width}), 0);
+				&::after {
+					transform: translate(calc(#{$dashboard-button-width} + -1px), 0);
+				}
+	
+				&::before {
+					transform: translate(calc(#{$dashboard-button-width} + -1px), 0);
+				}
+			}
+		}
+
+		.dashboard-panel__panel__button {
+			border-radius: 0px;
+			width: $dashboard-button-width;
+			height: $dashboard-button-height;
+
+			&.source-missing, &.source-missing, &.unknown-state {
+				&::before {
+					z-index: 2;
+				}
+			}
+
+			> .dashboard-panel__panel__button__content {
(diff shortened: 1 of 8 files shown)


hummelstrand (approved):

meteor/client/styles/shelf/dashboard.scss
```diff
@@ -390,31 +389,58 @@ $dashboard-button-height: 5.625em;
...
+					&.lower-third {
+						background: var(--segment-layer-background-lower-third);
+					}
+
+					&.remote {
+						background: var(--segment-layer-background-remote);
+					}
+
+					&.vt {
+						background: var(--segment-layer-background-vt);
```
jstarpl, line 427: This can be done easier/more reliably using `item-type-colors()` from https://github.com/nrkno/sofie-core/blob/031e3e3da71a9dc0eb6c4b15215eb375fc5e08ec/meteor/client/styles/_itemTypeColors.scss#L69-L79

jstarpl (changes requested):
I think the way it is implemented right now, this will also affect the look of all Dashboard Layouts, not just the Buckets, which just use the same components to display their contents. You can differentiate if you are in a Bucket Dashboard-panel or a regular panel by selecting on the `dashboard-panel__panel--bucket` class - https://github.com/nrkno/sofie-core/blob/031e3e3da71a9dc0eb6c4b15215eb375fc5e08ec/meteor/client/ui/Shelf/BucketPanel.tsx#L818

PeterC89:
> I think the way it is implemented right now, this will also affect the look of all Dashboard Layouts, not just the Buckets, which just use the same components to display their contents. You can differentiate if you are in a Bucket Dashboard-panel or a regular panel by selecting on the `dashboard-panel__panel--bucket` class -
> 
> https://github.com/nrkno/sofie-core/blob/031e3e3da71a9dc0eb6c4b15215eb375fc5e08ec/meteor/client/ui/Shelf/BucketPanel.tsx#L818

Should now be fixed in a hopefully not too messy way!

meteor/client/styles/shelf/dashboard.scss
```diff
@@ -616,4 +616,95 @@ $dashboard-button-height: 5.625em;
...
+				> .dashboard-panel__panel__button__label-container {
+					background-color: #000;
+					height: $dashboard-button-label-height;
+					display: flex;
+					position: absolute;
+					top: $dashboard-button-thumbnail-height;
+					right: unset;
+					bottom: unset;
+					left: unset;
+					padding: unset;
```
jstarpl, line 664: This seems weird? If the label is supposed to be attached to the bottom of the button, why position it from the top? Wouldn't it make more sense to just set `top: auto; bottom: 0`? Also, I think the label was supposed to span the entire width of the button, so I think this should be `left: 0; right: 0` as well, so that it looks like this?

obraz

If not, and the label is supposed to be a left-aligned flag, the `min-width` on `.dashboard-panel__panel__button__label` should be set to `fit-content` so that it doesn't look like this:

obraz

but instead looks like this:

obraz

In which case we would need a right padding value, I think.

jstarpl (approved):

PR #850, opened 2023-02-08
chore: Replace hardcoded translation script with generic

nytamin:
build-script.

The language scripts are hardcoded in meteor/package.json

All languages defined in meteor/i18n/*.po are compiled into json-files.
(no functionality changed)

* **Other information**:

**Status**


meteor/scripts/i18n-compile-json.js
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	console.log(`Found languages: ${languages.join(', ')}`)
+
+	for (const lng of languages) {
+		await runCmd(
+			`i18next-conv -l ${lng} -s i18n/${lng}.po -t public/locales/${lng}/translations.json --skipUntranslated`
+		)
+	}
+})().catch((e) => console.error(`ERROR: ${e}`))
```
Julusian, line 31: If this was an mjs file you wouldnt need this `;(async () => { ... })()` wrapper as top level await is possible.
You would have to change the `require()` to `import` too, but that is probably preferable as thats the syntax we use for typescript anyway

Julusian (approved):

meteor/scripts/i18n-compile-json.js
jstarpl, line 31: Shouldn't this throw, so that the exit code is non-zero? Or wait for all commands to "settle" and then exit with a non-zero code?

jstarpl (approved):
Just a workflow question on whether this script should exit with a non-zero code if one of the conversions fails.

PR #858, opened 2023-02-22
Fix: bug on package manager status page

nytamin:
This PR performs a quick-fix for a bug on package-manager's status page.

The console errores with "insert failed" when calling `PeripheralDeviceAPI.executeFunctionWithCustomTimeout()`

This PR aims to restore a previously working function, however we should probably make a later change so that we won't allow inserting `PeripheralDeviceCommands` client-side (and move `PeripheralDeviceAPI.executeFunctionWithCustomTimeout()` to server-side only.


Julusian (approved):
This should probably change `remove` to `return true` too, as `executeFunctionWithCustomTimeout` will also call that.

Also wow, this has been linger in the code for way too long without being noticed..

jstarpl (changes requested):
I'm not quite sure if we need this hack? This seems to be as simple as replacing `PeripheralDeviceAPI.executeFunction(this.props.deviceId, 'getExpetationManagerStatus')` with `MeteorCall.client.callPeripheralDeviceFunction('auto-refresh', this.props.deviceId, 1000, 'getExpetationManagerStatus')`, isn't it?

jstarpl (approved):

PR #859, opened 2023-02-23
fix: Package Manager expanded status without needing client write-access to PeripheralDeviceCommands

jstarpl:

Bug fix

Rewrites the special Package Manager interface as a functional component and removes the requirement for write access to the PeripheralDeviceCommands collection.

* **Other information**:

**Status**


nytamin (approved):
I found a few issues in the implementation. I pushed a fix for those and tested locally, and it seems to work now.

PR #860, opened 2023-02-24
Chore: Allow node 14+ in published libs

nytamin:
This PR changes the allowed Node-versions in package.json files to allow all Node-versions above 14.19 (and not only allow 14 and 16).

This applies only to the libraries that are published and used by external apps.

My reasoning for this is that I think that we should not hold back other libraries by disallowing the use of newer versions of Node

### DIFF
--- a/packages/server-core-integration/package.json
+++ b/packages/server-core-integration/package.json
@@ -48,7 +48,7 @@
 		"license-validate": "yarn sofie-licensecheck"
 	},
 	"engines": {
-		"node": "^14.19 || ^16.14"
+		"node": ">=14.19"
 	},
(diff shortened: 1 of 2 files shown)


jstarpl (approved):
Does anyone remember why we needed `^16.14` specifically? Generally, I agree that there's no point in limiting consumers to a ceiling.

Julusian:
> Does anyone remember why we needed `^16.14` specifically? Generally, I agree that there's no point in limiting consumers to a ceiling.

Perhaps it was inherited from a dependency? I can't think what that would be though

Julusian (approved):

PR #863, opened 2023-02-27
fix: infinite piece timing consistency (SOFIE-2117)

jstarpl:

Bug fix

The timing of the timeline objects within an Infinite Piece can be inconsistent if using a preroll. This is because their starting times are incorrectly capped at 0, where they should be allowed to pre-roll.

The `infiniteGroup.enable.start` is offset by the preroll amount, thus resulting in the same "ultimate" value of start time.

* **Other information**:

**Status**


Julusian (approved):
I would like to give it a bit of a test, but I don't think I can do that this week. Im happy for it to bemerged and I can test it next week

jstarpl:
> I would like to give it a bit of a test, but I don't think I can do that this week. Im happy for it to bemerged and I can test it next week

There are some failing tests in the suite for job-worker there that I need to understand first: https://github.com/nrkno/sofie-core/actions/runs/4282232508/jobs/7457095348

nytamin (approved):

PR #868, opened 2023-03-03
fix: JobQueueWithClasses

nytamin:
Fixes to JobQueueWithClasses

A few bugs in JobQueueWithClasses (see commits)

Fixed bugs, added unit test

* **Other information**:

**Status**

### DIFF
--- a/packages/shared-lib/src/lib/JobQueueWithClasses.ts
+++ b/packages/shared-lib/src/lib/JobQueueWithClasses.ts
@@ -71,7 +71,12 @@ export class JobQueueWithClasses {
 				resolve,
 				reject,
 			})
-			if (this.#autoStart && this.#paused) this.start()
+			if (this.#autoStart) {
+				// debounce so fn() isn't executed synchronously with .add():
+				setImmediate(() => {
+					this.start()
+				})
+			}
 		})
 	}
 	/** Clear queue */
--- a/packages/shared-lib/src/lib/JobQueueWithClasses.ts
+++ b/packages/shared-lib/src/lib/JobQueueWithClasses.ts
@@ -26,6 +26,8 @@ export class JobQueueWithClasses {
 	#executionWrapper?: WrapperFunction<any>
 	#resolutionWrapper?: DeferFunction
 
+	#waitForDonePromises: { resolve: () => void }[] = []
+
 	constructor(opts?: Options) {
 		this.#autoStart = opts?.autoStart ?? true
 		this.#executionWrapper = opts?.executionWrapper
@@ -96,7 +98,13 @@ export class JobQueueWithClasses {
 			// eslint-disable-next-line no-constant-condition
 			while (true) {
 				const firstIn = this.#queue.shift()
-				if (!firstIn) break
+				if (!firstIn) {
+					// No more jobs on queue.
+					this.#waitForDonePromises.forEach((p) => {
+						p.resolve()
+					})
+					break
+				}
 				try {
 					await firstIn.fn()
 					firstIn.resolve()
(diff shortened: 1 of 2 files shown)


packages/shared-lib/src/lib/JobQueueWithClasses.ts
```diff
@@ -102,6 +117,18 @@ export class JobQueueWithClasses {
 			console.error(error)
 		})
 	}
+	/** Returns the count of waiting jobs in the queue (ie not started yet) */
+	getWaiting(): number {
+		return this.#queue.length
+	}
+	/** Returns a Promise that resolves when the queue is eventually empty */
+	async waitForDone(): Promise<void> {
+		if (!this.#queue.length) return Promise.resolve()
```
Julusian, line 126: Is this correct? If there is a job currently executing, it looks like this will resolve before it finishes.

packages/shared-lib/src/lib/JobQueueWithClasses.ts
```diff
@@ -102,6 +117,18 @@ export class JobQueueWithClasses {
...
+
+		return new Promise((resolve) => {
+			this.#waitForDonePromises.push({ resolve })
```
Julusian, line 129: Does `#waitForDonePromises` benefit from being an array instead of a single promise that all callers are provided?

Julusian (approved):

packages/shared-lib/src/lib/JobQueueWithClasses.ts
```diff
@@ -89,7 +98,13 @@ export class JobQueueWithClasses {
 			// eslint-disable-next-line no-constant-condition
 			while (true) {
 				const firstIn = this.#queue.shift()
-				if (!firstIn) break
+				if (!firstIn) {
+					// No more jobs on queue.
+					this.#waitForDonePromises.forEach((p) => {
+						p.resolve()
+					})
```
jstarpl, line 105:
```suggestion
					})
					this.#waitForDonePromises.length = 0
```

Should we forget about these promises now, so that this doesn't leak?

packages/shared-lib/src/lib/__tests__/JobQueueWithClasses.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+	test('queue jobs synchronously', async () => {
+		const queue = new JobQueueWithClasses()
+
+		const results: string[] = []
+
+		queue.add(async () => {
+			results.push('start_0')
+			await sleep(10)
+			results.push('end_0')
+		})
```
jstarpl, line 14: I guess it would be good to `.catch()` all of these, so that the linting doesn't complain.

jstarpl (changes requested):

packages/shared-lib/src/lib/JobQueueWithClasses.ts
```diff
@@ -93,41 +94,53 @@ export class JobQueueWithClasses {
...
+				console.error(error)
+			})
 	}
 	/** Returns the count of waiting jobs in the queue (ie not started yet) */
 	getWaiting(): number {
 		return this.#queue.length
 	}
-	/** Returns a Promise that resolves when the queue is eventually empty */
+	/** Returns a Promise that resolves when the queue is eventually empty and all jobs are done */
 	async waitForDone(): Promise<void> {
```
Julusian, line 132: I think this `waitForDone` implementation has a race condition.

If you call `add()`, don't call `start()` (because autostart was set to true), then immediately call `waitForDone()`, `this.#jobIsRunning` will be false. as `start()` hasnt been run yet (its inside of a `setImmediate`), resulting in the call to `waitForDone()` returning `Promise.resolve()` instead of creating the correct promise. 
I think that dropping the ` && !this.#jobIsRunning` from L133, I don't see the purpose in that being part of the guard

PR #884, opened 2023-03-15
Feat: remove makeready and use TSR actions

baltedewit:
This PR removes the use of makeReady and standDown during rundown activation and deactivation and instead exposes methods to the blueprints during activiation / deactivation to execute TSR Actions inside of the playout-gateway


packages/job-worker/src/blueprints/context/RundownActivationContext.ts
```diff
@@ -0,0 +1,78 @@
...
+			showStyleCompound,
+			context.getShowStyleBlueprintConfig(showStyleCompound),
+			rundown
+		)
+
+		this._context = context
+		this._cache = cache
+	}
+
+	async listPeripheralDevices(): Promise<PeripheralDevicePublicWithActions[]> {
```
Julusian, line 37: This code looks duplicated from the other context? It would be good to refactor it to be shared

Julusian (approved):

PR #889, opened 2023-03-24
Fix: out of sync playhead countdown

thomasslee97:

Bug fix.

When taking from one Part to another, the timing data for the previous Part lingers in the context provided by the `RundownTimingProvider` and is provided to the `CurrentPartRemaining` component. This results in the coundown continuing to display the countdown for the previous Part.

To fix this issue, two measures have been taken:
1. The `RundownTimingProvider` now reports what the `currentPartInstanceId` was when it performed the timing calculation as part of the `RundownTimingContext` object that it provides. This allows the `CurrentPartRemaining` component to not render the countdown timer if the timing data that it has received is stale. This does result in the timer being briefly hidden if up-to-date timing data does not arrive quickly.
2. To that end, the `RundownTimingProvider` has been updated to force all of the timing events (LR, HR, Synced) to be emitted whenever the current or next part instances change. This results in a much shorter period of time where there is stale data, and reduces the number of frames for which the coundown is hidden while it catches up with reality. I would like to stress that it is a matter of frames - during a screen capture I was unable to see any time where the coundown ended up hidden even when stepping through frame-by-frame. This is more of a safety guard to prevent bad data from being presented to the user.

**Other information**:
A change that has been made here is that the `syncedEventTimeNow` value in `RundownTimingProvider` has been removed so that `calmedDownNow` is used everywhere. This is because there was a disagreement between `syncedEventTimeNow` and `calmedDownNow` as to which second we were in, which caused a coundown of: `00:10`, `00:11`, `00:10`, `00:09`, for example. I am unsure if this is the right thing to do.

**Status**

### DIFF
--- a/meteor/client/ui/RundownView/RundownTiming/CurrentPartRemaining.tsx
+++ b/meteor/client/ui/RundownView/RundownTiming/CurrentPartRemaining.tsx
@@ -32,11 +32,12 @@ export const CurrentPartRemaining = withTiming<IPartRemainingProps, {}>({
 			if (!this.props.timingDurations || !this.props.timingDurations.currentTime) return null
 			if (this.props.timingDurations.currentPartInstanceId !== this.props.currentPartInstanceId) return null
 			const displayTimecode = this.props.timingDurations.remainingTimeOnCurrentPart
+			if (displayTimecode === undefined) return null
 			return (
 				<span
 					className={ClassNames(
 						this.props.className,
-						Math.floor((displayTimecode || 0) / 1000) > 0 ? this.props.heavyClassName : undefined
+						Math.floor(displayTimecode / 1000) > 0 ? this.props.heavyClassName : undefined
 					)}
 					role="timer"
 				>
--- a/meteor/client/ui/RundownView/RundownTiming/CurrentPartRemaining.tsx
+++ b/meteor/client/ui/RundownView/RundownTiming/CurrentPartRemaining.tsx
@@ -28,7 +28,9 @@ export const CurrentPartRemaining = withTiming<IPartRemainingProps, {}>({
 	dataResolution: TimingDataResolution.Synced,
 })(
 	class CurrentPartRemaining extends React.Component<WithTiming<IPartRemainingProps>> {
-		render() {
+		render(): JSX.Element | null {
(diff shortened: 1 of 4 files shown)


meteor/client/ui/RundownView/RundownTiming/CurrentPartRemaining.tsx
```diff
@@ -29,6 +29,8 @@ export const CurrentPartRemaining = withTiming<IPartRemainingProps, {}>({
 })(
 	class CurrentPartRemaining extends React.Component<WithTiming<IPartRemainingProps>> {
 		render(): JSX.Element {
+			if (!this.props.timingDurations || !this.props.timingDurations.currentTime) return <div></div>
+			if (this.props.timingDurations.currentPartInstanceId !== this.props.currentPartInstanceId) return <div></div>
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Won't the same be achievable without returning null if an empty fragment is returned?

meteor/client/ui/RundownView/RundownTiming/RundownTimingProvider.tsx
```diff
@@ -207,12 +207,11 @@ export const RundownTimingProvider = withTracker<
 				this.dispatchLREvent(calmedDownNow)
 			}
 
-			const syncedEventTimeNow = Math.floor(now / 1000) * 1000
```
ianshade, line 210: I believe `syncedEventTimeNow` makes other calculations rounded perfectly to the whole second represented by the synced event. Without it some values may start jumping unexpectedly e.g. you'll get timers that count like 00:01, 00:02, 00:02, 00:04

ianshade: Additionally I believe this made the synced timers tick as closely to the system clock as possible, which I don't think is the case anymore.

thomasslee97: Curiously, that sort of sequence of counting was what I was seeing before removing `syncedEventTimeNow`

ianshade: Was that happening before you made any changes attempting to fix this bug?

PR #890, opened 2023-03-24
fix: Piece preroll in Part preroll calculations

thomasslee97:

Bug fix

Piece preroll is not accounted for when calculating Part preroll inside `rundownTiming`.

Pieces are provided to `updateDuration` so that Piece preroll can be accounted for.

**Other information**:
Most of the changes are to facilitate calls to `calculatePartInstanceExpectedDurationWithPreroll` which are made by a wide range of components, so a set of Pieces needs to make its way through various components to their children.
This change does add an additional call to a collection to the UI, specifically a query of the `Pieces` collection is performed by the `RundownTimingProvider`.

**Status**


jstarpl (approved):
Overall, this feels alright, but I'm beginning to wonder if we should start taking PartInstances and PieceInstances into account now, which would complicate this even further.

jstarpl (approved):

meteor/lib/collections/rundownPlaylistUtil.ts
```diff
@@ -313,6 +314,18 @@ export class RundownPlaylistCollectionUtil {
 		const instances = RundownPlaylistCollectionUtil.getActivePartInstances(playlist, selector, options)
 		return normalizeArrayFunc(instances, (i) => unprotectString(i.part._id))
 	}
+	static getPiecesForParts(parts: Array<PartId>): Map<PartId, Piece[]> {
+		const allPieces = Pieces.find({ startPartId: { $in: parts } }).fetch()
+		const piecesMap = new Map<PartId, Piece[]>()
+
+		for (const piece of allPieces) {
```
Julusian, line 321: There is a helper in corelib which could be used here `groupByToMap`

packages/corelib/src/playout/timings.ts
```diff
@@ -157,7 +157,10 @@ export function getPartTimingsOrDefaults(
 }
 
 function calculateExpectedDurationWithPreroll(rawDuration: number, timings: PartCalculatedTimings): number {
-	return Math.max(0, rawDuration + timings.toPartDelay - timings.fromPartRemaining)
+	// toPartDelay is removed from fromPartRemaining here, otherwise the values cancel out and the 'from' Part will
+	//  count overtime when prerolling into the 'to' Part.
+	// As a result, the 'to' Part will have a countdown which includes its own preroll.
+	return Math.max(0, rawDuration + timings.toPartDelay - (timings.fromPartRemaining - timings.toPartDelay))
```
Julusian, line 163: This change seems odd.. It was already using `timings.toPartDelay`, and is now taking the value twice.
That could be reformatted as `rawDuration + (2 * timings.toPartDelay) - timings.fromPartRemaining`

meteor/client/ui/SegmentTimeline/Parts/FlattenedSourceLayers.tsx
```diff
@@ -58,6 +58,7 @@ export function FlattenedSourceLayers(props: IFlattenedSourceLayerProps): JSX.El
 									onDoubleClick={props.onPieceDoubleClick}
 									mediaPreviewUrl={props.mediaPreviewUrl}
 									piece={piece}
+									pieces={layer.pieces.map((p) => p.instance.piece)}
```
Julusian, line 61: Using `map` here will cause all the `SourceLayerItemContainer` child components to be rerendered whenever this `FlattenedSourceLayers` component does. It would be good to use some `useMemo` to avoid this

Julusian (changes requested):
I don't understand the change in `calculateExpectedDurationWithPreroll` and would like to before I ok this

meteor/client/ui/SegmentTimeline/Parts/FlattenedSourceLayers.tsx
thomasslee97: Ok, I've moved that to the top level to construct a map of Pieces https://github.com/nrkno/sofie-core/pull/890/commits/2e3d8ea1562b7418f76dbaf47c95cf281f14c1d4
The `?? []` feels unnecessary but I don't like the idea of `piecesForLayers.get(...)!` much better

meteor/client/ui/SegmentContainer/withResolvedSegment.ts
```diff
@@ -207,6 +210,7 @@ export function withResolvedSegment<T extends IProps, IState = {}>(
 			)
 
 			const rundownOrder = RundownPlaylistCollectionUtil.getRundownOrderedIDs(props.playlist)
+			const pieces = RundownPlaylistCollectionUtil.getPiecesForParts(orderedAllPartIds)
```
jstarpl, line 213: I think it would be good to put this into something like a `memoizedIsolatedAutorun`. Also, would it be possible to limit the scope of reactivity on `getPiecesForParts` like it's done on `getSegmentsAndPartsSync` @ L183 above? Ultimately, we are passing a `CalculateTimingsPiece` anyway, so no point having all of the other properties there in the first place. I think the less properties we depend on, the better?

jstarpl (changes requested):

meteor/client/ui/SegmentContainer/withResolvedSegment.ts
thomasslee97: I've made the requested changes

packages/corelib/src/playout/timings.ts
Julusian: Sorry, could you change it back to your formatting of the line, that made the intent clearer. I was simply trying to show how odd the calculation looks.
But ok, the change sounds reasonable and this value is only used by the UI, so I am happy to go with it

PR #891, opened 2023-03-29
fix: "waitForPromise can't be used client-side" is periodically printed in the Browser Console

jstarpl:

Bug fix

A log line for an error "waitForPromise can't be used client-side" is logged on an interval.

`waitForPromise` should not be run when running Client-side.

* **Other information**:

**Status**


Julusian (approved):
I don't think the `Meteor.isClient` check is necessary here, we don't need to wait for the result in the same fiber, so the catch method should be fine on the server too

PR #892, opened 2023-03-31
refactor: simplify media status

jstarpl:

Refactoring

We are pulling a lot of very reactive data to render previews, thumbnails, etc. We also have a lot of duplicated logic between server & client for the Piece statuses.

We use the `uiPieceContentStatuses` publication for Media Statuses on Pieces and we use a new `packageContainerPackageStatusesSimple` publication for low reactivity PackageContainerPackageStatuses, without the `modified` field, which we don't need in most cases anyway.

* **Other information**:

**Status**


jstarpl:
I've attempted a re-write, with the logic still staying client-side, but using hooks, however that proved to have worse performance than the HOC we currently have, so it needs a different approach. I'm going to salvage the publication removing the `modified` field and use it in a new PR.

PR #904, opened 2023-04-21
Fix: Cronjob cleanup

nytamin:
Fix, Improve database cleanup during nightly cronjob.

We have an ongoing issue where database collections grow continously (when not having run migrations for some time).
This is due to the nightly cronjob was missing a few cleanup procedures (such as externalMessageQueue).

The nightly cronjob now executes the same function as the migrations cleanup function.
Also the cleanup of ExternalMessageQueue now takes the `expires` property into account.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/cleanup.ts
+++ b/meteor/server/api/cleanup.ts
@@ -52,6 +52,10 @@ import {
 	StudioId,
 } from '@sofie-automation/corelib/dist/dataModel/Ids'
 
+/**
+ * If actuallyCleanup=true, cleans up old data. Otherwise just checks what old data there is
+ * @returns A string if there is an issue preventing cleanup. CollectionCleanupResult otherwise
+ */
 export async function cleanupOldDataInner(actuallyCleanup: boolean = false): Promise<CollectionCleanupResult | string> {
 	if (actuallyCleanup) {
 		const notAllowedReason = await isAllowedToRunCleanup()
--- a/meteor/server/api/cleanup.ts
+++ b/meteor/server/api/cleanup.ts
@@ -1,4 +1,4 @@
-import { ProtectedString, getCurrentTime, getCollectionKey } from '../../lib/lib'
+import { ProtectedString, getCurrentTime, getCollectionKey, waitForPromise } from '../../lib/lib'
 import { CollectionCleanupResult } from '../../lib/api/system'
 import { MongoQuery } from '../../lib/typings/meteor'
 import { AdLibActions } from '../../lib/collections/AdLibActions'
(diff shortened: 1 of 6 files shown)


meteor/server/cronjobs.ts
```diff
@@ -46,79 +39,21 @@ export function nightlyCronjobInner(): void {
...
-		lowPrioFcn(() => {
-			IngestDataCache.remove(roc._id)
-		})
-		rundownCacheCount++
-	})
-	if (rundownCacheCount) logger.info('Cronjob: Will remove cached data from ' + rundownCacheCount + ' rundowns')
-
-	const cleanLimitTime = getCurrentTime() - 1000 * 3600 * 24 * 50 // 50 days ago
-
-	// Remove old PartInstances and PieceInstances:
```
Julusian, line 65: I don't see this cleanup being done anymore

meteor/server/api/cleanup.ts
```diff
@@ -70,134 +82,211 @@ export async function cleanupOldDataInner(actuallyCleanup: boolean = false): Pro
...
+		ownedByDeviceId(PeripheralDeviceCommands)
 	}
-	// Buckets
+
+	// Documents owned by Parts:
 	{
-		ownedByStudioId(Buckets)
+		const partIds = getAllIdsInCollection(Parts, removedParts)
+		removeByQuery(PartInstances, {
+			$or: [{ rundownId: { $nin: rundownIds } }, { 'part._id': { $nin: partIds } }],
```
nytamin, line 284: @Julusian this removal is intended to replace the one you mentioned in the other comment.
However, I'm a tad unsure if this query is safe, let's discuss it a bit?
The original query was
```javascript
{
	reset: true,
	'timings.plannedStoppedPlayback': { $lt: cleanLimitTime },
	'part._id': { $nin: existingPartIds },
}
```
do you think we should apply something similar here?
so
```javascript
{
	reset: true,
	'timings.plannedStoppedPlayback': { $lt: getCurrentTime() - Settings.maximumDataAge },
	$or: [{ rundownId: { $nin: rundownIds } }, { 'part._id': { $nin: partIds } }],
}
```

meteor/server/cronjobs.ts
nytamin: I commented on where PartInstances are handled, let's continue that talk there.
I also added a cleanup of PieceInstances when their PartInstance has been removed [in cleanup.ts:292](https://github.com/nrkno/sofie-core/pull/904/files#diff-975acd74a53b0f96f509165792e86422ec38d8142023efe9d9ca757d8c6381ffR292).

nytamin:
I've updated the PR to target R49, could you give it another read please, @Julusian ?

meteor/server/api/cleanup.ts
Julusian: I think this is going to be tricky to combine the two previous behaviours.
The key thing to note is that the `PartInstances` and `PieceInstances` replace the old `AsRunLog` that we used to have.
Which suggests that maybe what we are doing elsewhere is bad, as we remove all the PartInstances when removing a Rundown.

The cleanup was removing any where their Rundown was missing.
The cronjob was removing any that had finished playback and the part they were of was missing. (Intended to catch the TV2DK case of reusing a single rundown for months)

So I think a better representation would be:
```js
{
	$or: [
	{
		// where the parent Rundown is missing
		rundownId: { $nin: rundownIds } 
	},
	{
		// Remove any from long running rundowns where they have long since expired
		reset: true,
		'timings.plannedStoppedPlayback': { $lt: getCurrentTime() - Settings.maximumDataAge },
		'part._id': { $nin: partIds },
	}]
}

```

Similar to your proposal, but better matches what we are doing elsewhere.  
I do feel like yours would be better so that there is a longer living AsRunLog, but that isnt what we are doing elsewhere in Sofie

nytamin: Good points! Let's go with your suggestion

PR #905, opened 2023-04-25
fix: Package Manager status page crash

jstarpl:

Bug fix

When you open the Package Manager peripheral device Status page, Core can crash with an uncaught exception, if the device is not assigned to a studio.

The thrown error will be returned back to the client instead of crashing Core.

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/api/client.ts
+++ b/meteor/server/api/client.ts
@@ -334,48 +334,52 @@ export namespace ServerClientAPI {
 			)
 		}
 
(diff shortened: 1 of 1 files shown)


Julusian (commented):
Those `try {} catch {}` blocks won't be changing anything. We are already in async functions, so any thrown error results in a promise rejection.

Changing that `void` to `await` would be the fix. Makes the function wait for the promise to resolve instead of letting it run in parallel (defeating the purpose of the security check), resulting in the uncaughtPromiseRejection when it fails

PR #907, opened 2023-04-26
feat: move gateway settings onto Studio SOFIE-1330

Julusian:

Feature/refactoring

Depends on #906 

Settings for PeripheralDevice subdevices live in the PeripheralDevice documents and are configured on each gateway in the ui.
The new blueprints upgrade and config flows are unable to define any subdevices, which they were previously capable of with the old migrations system.

The settings for these subdevices has been moved off the gateways and onto the studio document, using the `ObjectWithOverrides` that we use elsewhere. This lets studio blueprints define defaults, and for someone to configure them in the UI to fill in the remaining properties.

The 'Attached Subdevices' settings section for the Studio has been expanded to house these subdevices, grouped by the gateway category they belong to.

Screenshot from 2023-04-26 15-21-30

SubDevices can be added and removed (same as we allow for Mappings). Because they reside on the document now, there is a dropdown to select which PeripheralDevice they belong to. `Unassigned` is an allowed option, and is how they will be set when created by blueprints.  
Due to the ui being built by config manifests from the gateway, when set to unassigned the settings cannot be viewed or edited.
Once assigned to a gateway, they look the same as before

image

The PeripheralDevice page has been simplified as a result:
image

There is a breaking change in the apis used by PeripheralDevices. Instead of subscribing and using the `peripheralDevices` collection and publication, they must now use `peripheralDeviceForDevice`. The old publication is used by the UI, and needs to be in the raw form, but the gateways need a publication which provides the settings of the subdevices.  
This new publication is using a new simpler type (`PeripheralDeviceForDevice`) that strips out more unused properties, the `getPeripheralDevice` ddp method has been updated to also use this new type.

* **Other information**:

Providing these subdevices from blueprints is optional, which allows existing blueprints to work unmodified. The existing config gets copied over to be pure overrides, so will not be lost when rerunning the applyConfig.

The settings of the device, outside of the subdevices still resides on each PeripheralDevice. These are less likely to be wanted to be set by blueprints, and were outside of the scope of this task.

Some refactoring of `server-core-integration` was done as part of this, to help minimise the amount of unused data being written into mongo for each of the subdevices. This could be pulled out into its own PR if needed, 

**Status**


meteor/client/ui/Settings/SettingsMenu.tsx
```diff
@@ -253,7 +253,7 @@ function SettingsMenuStudio({ studio }: SettingsMenuStudioProps) {
 	const childLinks = React.useMemo(
 		() => [
 			{ label: t('Generic Properties'), subPath: `generic` },
-			{ label: t('Attached Devices'), subPath: `devices` },
+			{ label: t('Peripheral Devices'), subPath: `devices` },
```
jstarpl, line 256: Why this name change?

packages/corelib/src/dataModel/Studio.ts
```diff
@@ -84,17 +84,59 @@ export interface DBStudio {
...
 	previewContainerIds: string[]
 	thumbnailContainerIds: string[]
 
-	/** Details on the last blueprint used to generate the defaults values for this */
-	lastBlueprintConfig: LastBlueprintConfig | undefined
+	playoutDevices: ObjectWithOverrides<Record<string, StudioPlayoutDevice>>
+
+	ingestDevices: ObjectWithOverrides<Record<string, StudioIngestDevice>>
+
+	inputDevices: ObjectWithOverrides<Record<string, StudioInputDevice>>
```
jstarpl, line 107: I'm a little bit unsure about this architecture. One of the advantages of the "generic" Peripheral Device structure, was that it was very easy to just add a new peripheral device and have it working. With this setup, devices need to "neatly" slot into one-and-only-one of these categories, which may turn out to be problematic. Could this be refactored somehow, so that we don't have to categorize the devices so strongly? (Yes, I know that PeripheralDevices have types and subTypes, but that system seems like it could be easily refactored away to something like an "capabilities" array, it's not a core part of the system).

jstarpl (changes requested):
I'm a little bit unsure about this sorting of peripheralDevices into these bins. Is this necessary?

packages/corelib/src/dataModel/Studio.ts
jstarpl: Ah, OK, I completely misread this. I can now see, that as far as the previous structure is concerned, these are technically "sub-devices".

meteor/__mocks__/defaultCollectionObjects.ts
```diff
@@ -105,9 +105,14 @@ export function defaultStudio(_id: StudioId): DBStudio {
 		_rundownVersionHash: '',
 		routeSets: {},
 		routeSetExclusivityGroups: {},
-		packageContainers: {},
-		previewContainerIds: [],
-		thumbnailContainerIds: [],
+		peripheralDeviceSettings: {
+			packageContainers: {},
+			previewContainerIds: [],
+			thumbnailContainerIds: [],
```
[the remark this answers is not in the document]
Julusian: I don't have a strong opinion on whether these should be moved or not. I think it was suggested by @nytamin, perhaps he can argue the case?

packages/corelib/src/dataModel/Studio.ts
Julusian: I've added some comments, which explicitly states them as sub-devices.

meteor/__mocks__/defaultCollectionObjects.ts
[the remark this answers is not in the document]
nytamin: I don't have strong opinions on it either, Jan's suggestion sounds reasonable to me!

PR #908, opened 2023-04-26
reafactor(GUI): App as functional component

jstarpl:

Refactoring

Sofie has a lot of very simple, legacy components implemented as Class components, even though they can easily be converted to a Functional component.

Implement App and a couple of top-level View components as Functions.

* **Other information**:

**Status**

### DIFF
--- a/meteor/client/ui/Account/NotLoggedIn/LoginPage.tsx
+++ b/meteor/client/ui/Account/NotLoggedIn/LoginPage.tsx
@@ -10,8 +10,8 @@ import { getUser, User } from '../../../../lib/collections/Users'
 import { NotLoggedInContainer } from './lib'
 import { stringifyError } from '@sofie-automation/corelib/dist/lib'
 
-interface ILoginProps extends RouteComponentProps<{ token: string }> {
-	requestedRoute: string
+interface ILoginProps extends RouteComponentProps<{ token?: string }> {
+	requestedRoute?: string
 }
 
 interface ILoginPageProps extends ILoginProps {
--- a/meteor/client/ui/ActiveRundownView.tsx
+++ b/meteor/client/ui/ActiveRundownView.tsx
@@ -1,159 +1,97 @@
-import * as React from 'react'
+import React, { useEffect } from 'react'
 import * as _ from 'underscore'
-import { Route, Switch } from 'react-router-dom'
-import { translateWithTracker, Translated } from '../lib/ReactMeteorData/ReactMeteorData'
-import { RundownPlaylist } from '../../lib/collections/RundownPlaylists'
+import { NavLink, Route, Switch, useRouteMatch } from 'react-router-dom'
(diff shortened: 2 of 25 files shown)


meteor/client/ui/ActiveRundownView.tsx
```diff
@@ -1,159 +1,97 @@
...
-				this.subscribe(PubSub.uiStudio, this.props.studioId)
-			}
+	const studioReady = useSubscription(PubSub.uiStudio, studioId)
+	const playlistReady = useSubscription(
+		PubSub.rundownPlaylists,
+		_.extend(
+			{
+				activationId: { $exists: true },
+			},
+			studioId
```
Julusian, line 26: the typings say that `studioId` will also be defined now, so this ternary and `_.extend` look unnecessary

meteor/client/ui/App.tsx
```diff
@@ -374,9 +365,162 @@ export const App = translateWithTracker(() => {
...
+				testing: getAllowTesting(),
+				service: getAllowService(),
+			})
+		}
+	}, [location.search, user, subsReady])
+
+	return roles
+}
+
+function useFeatureFlags() {
```
Julusian, line 473: Perhaps the return of this could be wrapped in a context which other places in the code can use to retrieve the values?
Same for `useRoles`.
That will allow us to start referencing this config in a reactive way, preparing for when we eventually want it to be fully reactive. This will be more important for the `useRoles`, as that depends on `user`

Julusian (approved):

meteor/client/ui/Account/NotLoggedIn/LoginPage.tsx
```diff
@@ -10,8 +10,8 @@ import { getUser, User } from '../../../../lib/collections/Users'
 import { NotLoggedInContainer } from './lib'
 import { stringifyError } from '@sofie-automation/corelib/dist/lib'
 
-interface ILoginProps extends RouteComponentProps<{ token: string }> {
-	requestedRoute: string
+interface ILoginProps extends RouteComponentProps<{ token?: string }> {
```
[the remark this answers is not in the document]
jstarpl: I think it should technically be optional. This view behaves differently depending on if a token is provided (which means it's in a "verify-email" mode) and if it's not provided (regular login). So I think it makes sense to say it's optional, because it's a mode-switch.

meteor/client/ui/App.tsx
jstarpl: I think it can, but I didn't want to inflate the scope of this PR too much. All I wanted to achieve here was to refactor these components as funcional components and modernize the Routes in the Router so that we can at some point do an easier switch to react-router-dom v6. Also, I'm not sure if we want to commit to using contexts, especially if we were to switch to a new state management, like MobX, or something. I think I'd rather keep it minimal for now, and do it when we make a decision on how to progress post-Meteor.

PR #910, opened 2023-04-27
fix: Replaced "the the" with "the"

hummelstrand:
Replaced what must have been a result of a bad search and replace.

-

-

* **Other information**:
-
**Status**


meteor/public/locales/nb/translations.json
```diff
@@ -934,8 +934,8 @@
     "Stop": "Stopp",
     "Clip \"{{fileName}}\" can't be played because it doesn't exist on the playout system": "Klippet \"{{fileName}}\" kan ikke spilles av fordi det ikke finnes på utspillingssystemet",
     "{{sourceLayer}} is not yet ready on the playout system": "{{sourceLayer}} er ennå ikke klar til å spilles ut fra avviklingsserver",
-    "{{sourceLayer}} is transferring to the the playout system": "{{sourceLayer}} overføres til avviklingsserver",
-    "{{sourceLayer}} is transferring to the the playout system and cannot be played yet": "{{sourceLayer}} overføres til avviklingsserver og kan ikke spilles av ennå",
+    "{{sourceLayer}} is transferring to  playout system": "{{sourceLayer}} overføres til avviklingsserver",
```
jstarpl, line 937: I think this needs to be the same as line 936 in `nn/translations.json` and also needs to match line 55 in `lib/mediaObjects.ts`, otherwise it will just not work.

packages/shared-lib/src/core/deviceConfigManifest.ts
```diff
@@ -154,7 +154,7 @@ export interface TableConfigManifestEntry extends ConfigManifestEntryBase {
 	defaultType: string
 	/** Used when the .config indexes are different from the type enum */
 	deviceTypesMapping?: any
-	/** The name of the the property used to decide the type of the entry */
+	/** The name the property used to decide the type of the entry */
```
jstarpl, line 157: I think this was supposed to be `The name of the property`

jstarpl (changes requested):
Some of the translations do not match the source strings, which will break the translations

packages/shared-lib/src/core/deviceConfigManifest.ts
hummelstrand: Correct, sloppy of me! Fixing.

meteor/public/locales/nb/translations.json
hummelstrand: OK, VisualCode auto-save problem. I must have committed the file before it auto-saved.

jstarpl (approved):

PR #912, opened 2023-04-27
chore: use yarn 3 SOFIE-2251

Julusian:

Dependency update

We are using yarn v1.22 in all of our repositories. This entered maintenance mode in early 2020, but it worked fine so we had no reason to move away from it.

Recently we have been seeing some weird errors where the answer given online is that it is fixed in newer versions of yarn.

The version of yarn we are using has been updated to v3. Developers do not need to install the new version, it is detected automatically by yarn v1 and used instead.

Developers running complex yarn commands themselves may need to change what they are running to continue working, but all of the scripts provided in the repository should remain working.

* **Other information**:

I am not liking the lack of console output during a `yarn install` at the root of the project. I can't find a solution to this other than to split the install into two scripts that must both be run. (You can't run a yarn script until the initial install has been done. And yarn is hiding the console output of postinstall scripts)

We have a bit of an architecture issue with https://github.com/nrkno/sofie-code-standard-preset in that it includes many tools that we invoke from our projects. This isn't really supported in yarn3, so we have to hack a way to be able to use those bins. This is not ideal and we should reevaluate this at some point. I suspect the answer will be that we need to stop defining our dependencies like that and to install them explicitly in each project, and live with the maintenance challenges that will reintroduce.

**Status**


packages/blueprints-integration/package.json
```diff
@@ -15,18 +15,18 @@
...
 	"scripts": {
-		"build": "rimraf dist && yarn build:main",
-		"build:main": "tsc -p tsconfig.build.json",
-		"lint:raw": "eslint --ext .ts --ext .js --ignore-pattern dist",
-		"lint": "yarn lint:raw .",
-		"unit": "jest",
-		"test": "yarn lint && yarn unit",
-		"watch": "jest --watch",
-		"cov": "jest --coverage; open-cli coverage/lcov-report/index.html",
+		"build": "run -T rimraf dist && run build:main",
```
nytamin, line 18: I'm curious, what does `run -T` do?

Julusian: `run ...` is a shorthand for `yarn run ...`.
`-T` means top-level. As we don't have `rimraf` defined as a dependency or dev-dependency of `blueprints-integration`, doing a `run rimraf` doesn't find the tool to execute. But the `-T` flag tells it to find the bin/script at the top level of the workspace, where we have got it defined as a dev-dependency.
It was either this, or add rimraf, tsc, and others as a dev-dependency of every package in the workspace. Luckily I discovered this flag before doing that

PR #914, opened 2023-05-02
fix: leaking `debounce-fn`s when queuing `EventsJobs.PartInstanceTimings` jobs SOFIE-2307

Julusian:

Bug fix

The studio-worker in Sofie is slowly leaking memory.

It looks to be the caching of these debounce-fns, one for every PartInstance taken, is leaking the `context` object for the job, which in turn is causing the `CacheForPlayout` for that job to be leaked.

By using a simpler method of for the debounce, where it can clean up after itself we can avoid this leak

* **Other information**:

**Status**

### DIFF
--- a/packages/job-worker/package.json
+++ b/packages/job-worker/package.json
@@ -45,7 +45,6 @@
 		"@sofie-automation/corelib": "link:../corelib",
 		"@sofie-automation/shared-lib": "link:../shared-lib",
 		"amqplib": "^0.10.3",
-		"debounce-fn": "^4.0.0",
 		"deepmerge": "^4.2.2",
 		"elastic-apm-node": "^3.38.0",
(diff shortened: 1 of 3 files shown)


jstarpl (approved):
Looks alright, though maybe we have the time to figure out how to refactor the job-worker/context so that we're not using the context in a timeout? I guess this is a unique case?

Julusian:
Yeah I believe this is the only case of using the context longer than it is supposed to be here.
The queueing other jobs portion it uses could be pulled out to avoid the lifetime concerns, but really this worker is not supposed to use any kind of setTimeout that isnt awaited. I dont have a good suggestion on improving that without going the full way of avoiding the in-memory debounce and using a proper work queue

PR #921, opened 2023-05-11
fix: small parts after #903

jstarpl:

Bug fix

There are some problems with rendering Small Parts in the Timeline view after #903 was merged.

Before:

What the bug looks like

As it's visible, the Small Parts overlay is not correctly aligned (it's not in the correct position of the timeline), the flags are incorrectly placed, etc. Also, the "Parts duration" counter incorrectly shows the "display duration" instead of the actual "expected duration" sum of the Parts.

There should be no problems with Small Parts. Also, the "relative" rendering mode for Parts was removed, because the only place that used it was the Small Parts component, it just made all the rendering more complicated, and the Small Parts could be implemented differently anyway.

After:

What it's supposed to look like

* **Other information**:

**Status**


jstarpl:
@alvancamp I've added screenshots for the before/after. This may not be an issue, if your particular rundown does not have Segments with very long and very short Parts at the same time.

PR #927, opened 2023-05-16
feat: use mongodb transactions in job-worker SOFIE-1534

Julusian:

Feature

When we do our writing to mongodb, we are issuing multiple mongodb bulkWrites in parallel, one for each collection with changes. If one of these fails, the others will be run to completion causing an inconsistent state in the database.

All of this bulk writing is done inside of transactions, so if one of the writes fails, the whole set will be rolled back. This mostly affects how the caches are saved, but there were some other places where some cleanup was done by issues stray writes during the `deferAfterSave` hook. A new hook `deferDuringSave` has been added which allows for accessing the save transaction for any additional updates. Additionally the `defer` hook was renamed to `deferBeforeSave` for better clarify of when it is run.

To ensure that transactions are consider in all places, the `transaction` parameter to the mongodb write methods is typed `IMongoTransaction | null` and must be provided. This ensures that the transaction is thought about, but allows for one not be to be used when it is not necessary. In all places where null is provided, this is justified with a comment.  
The read methods can take the transaction parameter as it can be useful, but it is not required.

To minimise the pain of this in unit tests, they now have access to `mockCollections` which provides direct access (typed to the class, not the interface) of the mock collections, so that they are not required to pass this parameter. This was done in one commit, and updating the tests to use it results in ~800 lines of added and removed code.

* **Other information**:

**Status**


packages/job-worker/src/ingest/bucketAdlibs.ts
```diff
@@ -134,155 +153,161 @@ export async function handleBucketItemImport(context: JobContext, data: BucketIt
...
-			},
-			context,
-			showStyleCompound,
-			watchedPackages
-		)
+	await context.directCollections.runInTransaction(async (transaction) => {
+		const ps: Promise<any>[] = []
+		let isFirstShowStyleVariant = true
+		for (const showStyleVariant of showStyleVariants) {
+			const showStyleCompound = createShowStyleCompound(showStyleBase, showStyleVariant)
```
jstarpl, line 160: This loop has now become very deeply nested. Maybe this could be extracted to it's own function that deals with an individual `showStyleVariant`, for clarity?

packages/job-worker/src/ingest/bucketAdlibs.ts
```diff
@@ -134,155 +153,161 @@ export async function handleBucketItemImport(context: JobContext, data: BucketIt
...
-			if (newRank === undefined) {
-				const [highestAdlib, highestAction] = await Promise.all([
-					context.directCollections.BucketAdLibPieces.findFetch(
-						{
-							bucketId: data.bucketId,
-						},
-						{
-							sort: {
-								_rank: -1,
+			if (rawAdlib) {
```
jstarpl, line 198: If this loop body were extracted to a separate function, it could also help with this huge condition block, and just have it abort early.

packages/job-worker/src/ingest/commit.ts
```diff
@@ -138,31 +139,39 @@ export async function CommitIngestOperation(
...
+									playlistId: protectString('__TMP__'),
+								},
+							},
+							transaction
 						)
 
-						if (updatedOldPlaylist) {
-							// ensure the 'old' playout is updated to remove any references to the rundown
-							await updatePlayoutAfterChangingRundownInPlaylist(
+						if (oldPlaylist) {
```
jstarpl, line 154: I think this is another candidate for returning early

packages/job-worker/src/playout/adlibAction.ts
```diff
@@ -58,34 +58,37 @@ export async function handleExecuteAdlibAction(
...
+						await executeDataStoreAction(actionContext, data.actionId, data.userData, data.triggerMode)
+					} catch (err) {
+						logger.error(`Error in showStyleBlueprint.executeDatastoreAction: ${stringifyError(err)}`)
+						throw err
+					}
+				})
 			}
+		} catch (err) {
+			logger.error(`Error in showStyleBlueprint.executeDatastoreAction: ${stringifyError(err)}`)
 		}
```
jstarpl, line 92: This bloc too has become very nested. Maybe we could break it out into individual methods for `executeDataStoreAction` and `executeAction` so that they could return early and help with this nesting?

packages/job-worker/src/playout/lock.ts
```diff
@@ -93,11 +95,15 @@ export async function runWithPlaylistCache<TRes>(
...
-		await fullCache.saveAllToDatabase()
+		await fullCache.saveAllToDatabase(existingTransaction)
 
 		return res
 	} catch (err) {
-		fullCache.discardChanges()
+		try {
+			fullCache.discardChanges()
+		} catch (e) {
+			// Ignore
```
jstarpl, line 105: Should this at the very least log some information that this threw an error? It shouldn't throw, generally, right?

packages/job-worker/src/rundownPlaylists.ts
```diff
@@ -331,17 +339,37 @@ export async function handleMoveRundownIntoPlaylist(
...
-						await updatePlayoutAfterChangingRundownInPlaylist(context, newPlaylist, oldPlaylistLock, null)
-					}
+						// Regenerate the playlist
+						const newPlaylist = await regeneratePlaylistAndRundownOrder(
+							context,
+							oldPlaylistLock,
+							transaction,
+							oldPlaylist
+						)
+						if (newPlaylist) {
```
jstarpl, line 362: This one could also return early

jstarpl (approved):
A very good effort for data integrity. I've noticed that the additional wrapper of transactions has the side-effect of already long and deeply nested methods being even longer and more nested. Some of these seem to now be very easy to simplify with early returns, others maybe could be broken down, for clarity?

packages/job-worker/src/ingest/bucketAdlibs.ts
Julusian: I dont think it is that simple to refactor this to achieve that. 
Each run of the loop modifies the `newRank` variable and adds to the `ps` array. The first run will set `onlyGenerateOneItem` and `isFirstShowStyleVariant`. And each run will modify one of `adlibIdsToRemove` or `actionIdsToRemove` depending what it was.
So to achieve that will take either a lot of parameters that may be mutated, or some complex return types that will need to be broken out and processed. Perhaps this is a good candidate for using a class, to allow for this shared state?

I have moved this method into its own file and broken out a couple of pretty isolated methods, and some other tidying. Going much further will mean largely rewriting it.

packages/job-worker/src/playout/lock.ts
Julusian: I hit this because `fcn()` a few lines above threw an error. And then this threw an error which hid the real error from me.
Reading into it, I agree that discardChanges shouldnt be throwing. But also, the discardChanges call here is rather pointless. The discard is reverting its internal state back to before the job was run, and then verifying that it did that correctly. As far as I can tell, this is simply costing us time in doing some clones, just before we discard the cache.
I'll review all the places we call discardChanges, and replace/remove the ones that seem odd like this.

Julusian:
Some of the refactoring isn't as far as you would like, but I'm going to merge this now to avoid this PR becoming even larger (the changes yesterday pushed it from +2200 to +2900), with diffs that make it hard to see what changed. If we want to continue refining this refactoring I am happy to start a new PR that will be easier to see the changes.

PR #928, opened 2023-05-16
Allow blueprint debugging

ianshade:
minor refactor allowing blueprints to be debugged while running within Core

breakpoints fail to set when attaching the debugger

breakpoints are set, inline source map is found and used

* **Other information**:

  Three issues spotted so far, some should be addressed, some at least acknowledged in the readme before this is ready
  - ~~When a breakpoint is triggered in the blueprints or in a worker, a timeout in threadedclass kills the worker.~~
  - A specific order may be required before the debugger registers the files and sets the breakpoints. It usually works best when I attach the debugger first, and then upload the blueprints. Maybe we could solve it by persisting the VMScript in memory?
  - When code changes are made and as a result the blueprints are built and re-uploaded while the debugger is attached, an already set breakpoint may jump to a different line.

**Status**


Julusian:
Nice, this was easier than I expected. 
For some reason I had trouble debugging a worker thread when I tried to not too long ago, but perhaps that was because I was using the chrome debugger not vscode?

One problem is that when paused on breakpoints, the timeouts in threadedclass can trigger. It would be good to disable those timout checks when run in inspect mode

ianshade:
@Julusian I updated the description with three issues that I've seen so far, including the one you've mentioned.
I have not tried the chrome debugger, but I'm suspecting it could work if we went the route from the first commit and provided the absolute paths.

jstarpl:
It seems like this Pull Request is done - can it be taken out of the Draft status? I would like to get it merged in before we freeze Release 50.

Julusian (approved):
I'm happy with this.

> A specific order may be required before the debugger registers the files and sets the breakpoints. It usually works best when I attach the debugger first, and then upload the blueprints. Maybe we could solve it by persisting the VMScript in memory?

I found that it seems to 'just work', but starting with the second occurance. Either at startup, or after reuploading blueprints, the second run of the blueprints seems to hit the breakpoints.  
I think this is fine, and unless an easy fix isn't worth digging into

The proxy object issue is a bit more annoying, apparently in `onTimelineGenerate` by timeline has a length of 0. I do think we should try to improve upon this, otherwise it makes for very hard understanding of what is happening in certain places

Julusian:
@ianshade for the NRK blueprints, I made the a simple config change to the webpack config, utilising builtin webpack variables to detect for being run in watch mode. I think this will be sufficient for including sourcemaps when appropriate
image

PR #931, opened 2023-05-19
feat: update meteor to 2.12 SOFIE-2368

Julusian:

Dependency updates

This updates Meteor from 2.7.3 to 2.12. In addition it makes the following key changes:
* Update mongodb client lib from 4.x to 5.x. (more on this below)
* Update typescript to 4.9 (packages was already using this version, meteor now matches)
* This changes how meteor types are imported, to use the recommended 'zodern:types' instead of pulling them from npm

Updating mongodb would have been nice to do in its own PR, but it is hard to do one without the other. The version of nodejs needed by 5.x of the lib is not satisfied by meteor 2.7.3, making it not possible to do that change first. The mongodb typings in 4.x of the lib cause typescript to crash. So doing either one without the other leaves Sofie in a 'broken' state.  
The key difference in this mongodb change is that the queries are no longer strongly typed. Mongodb disabled this due to the performance impact it has on the typescript compiler. It is however possible to opt-in to the stricter queries on a per query basis, so we could keep it enabled for collections/queries which would benefit from it and don't introduce much compile time cost.

* **Other information**:

Everyone will need to do a `db.adminCommand({ setFeatureCompatibilityVersion: 
\"5.0\" })` in their mongodb before updating to this version or their mongo database will be incompatible. After running this, their mongo database will not be compatible with release41 or older.

This version of meteor uses mongodb 6.0 in development, ideally 6.0 should be used in production deployments too.

**Status**


meteor/client/ui/Account/NotLoggedIn/SignupPage.tsx
```diff
@@ -93,11 +93,6 @@ export const SignupPage = translateWithTracker((props: ISignupPageProps) => {
 				email: this.state.email,
 				password: this.state.password,
 				profile: { name: this.state.name },
-				createOrganization: {
```
nytamin, line 96: Why remove this? Seems unrelated to this PR and also breaks the functionality in meteor\server\api\user.ts:109

meteor/client/lib/__tests__/rundown.test.ts
```diff
@@ -245,7 +245,7 @@ describe('client/lib/rundown', () => {
 				const outputLayerIds = Object.keys(showStyleBase.outputLayers)
 
 				const playlistActivationId = protectString('mock_activation_0')
-				mockRundownPlaylistsCollection.update(playlistId, {
+				mockRundownPlaylistsCollection.update(unprotectString(playlistId), {
```
nytamin, line 248: This seems like a step backwards in type safety, is it possible to keep the protectedString here?

nytamin (commented):
Code looks good in general (see comments).

A couple of notes:
* Since this PR breaks the db, I think this should not target R50, but a later release so there can be a coordinated deploy.
* In the changelog, we should add the update instructions, and link to https://guide.meteor.com/2.11-migration.html

meteor/client/ui/Account/NotLoggedIn/SignupPage.tsx
Julusian: I did not see the usage of this property, I think its presence got hidden by uses of a function with the same name in the same file. And this property was not defined in the interface used here, so was being flagged as a type error.

Instead I've added this property to the interface, so it is no longer a type error

meteor/client/lib/__tests__/rundown.test.ts
Julusian: Not easily. this `mockRundownPlaylistsCollection` is of type `Meteor.Collection`, so we would either need to return a wrapped type for these mock collections to be used here, or override the typings of the collection. As this is test code, and only a couple of places, I am not sure it is worth the cost to attempt either route.

Fun fact, the old query is not at all type safe. I can change it to `mockRundownPlaylistsCollection.update(showStyleBase._id, {` and typescript does not care. I think because its using the loose mongo queries. Only now it is being strict and rejecting the `ProtectedString` as something that isnt compatible with a `MongoQuery<T>`

PR #922, opened 2023-05-12
Rework Settings gui (SOFIE-2300)

nytamin:
_Work in progress_

This PR aims at change how the settings pages are made, in order to provide several quality-of-life issues, such as:

* A unified “filterable” settings tree component that we could reuse.
* Use VS Code Settings page as a reference
* Have the settings be divided into nested sections.
* Lists should be sortable
* Have an auto-focused “Search settings” text box

PR #933, opened 2023-05-23
feat: improve the visual layout of all settings pages

alvancamp:

Bug fix/feature, it could be argued either way

The layout of the settings pages is not so good.

The layout of the settings pages is better.

* **Other information**:

Before:
image

After:
image

Too many things have changed to screenshot them all, but I visually checked every part of every single page and they all looked good to me.

**Status**

### DIFF
--- a/meteor/client/lib/Components/LabelAndOverrides.tsx
+++ b/meteor/client/lib/Components/LabelAndOverrides.tsx
@@ -16,11 +16,13 @@ export interface LabelAndOverridesProps<T extends object, TValue> {
 	opPrefix: string
 	overrideHelper: OverrideOpHelperForItemContents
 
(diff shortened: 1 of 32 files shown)


nytamin:
Just a heads-up; we're currently looking into a full rewrite of the settings-pages (an initial draft here: https://github.com/nrkno/sofie-core/pull/922).
I think that we should be able to merge this in in the meantime, and I'll make sure to take these fixes into account in the rewrite.

jstarpl (approved):

Julusian:
I've just encountered this edge case in the new stylings:
image

It looks to be caused by that really long default value. Which is having the side effects of making the input boxes taller, and the reset button has been pushed outside of the component and is no longer visible

PR #939, opened 2023-05-31
Package statuses when Package Manager is offline (SOFIE-1513)

nytamin:

Refactor + change in display of package work statuses.

When PM is offline, Package-statuses are unaffected.

When PM is offline, Package Containers and Work Statuses display differently.

* **Other information**:

Screenshot:

image

**Status**

### DIFF
--- a/meteor/client/ui/Status/package-status/PackageContainerStatus.tsx
+++ b/meteor/client/ui/Status/package-status/PackageContainerStatus.tsx
@@ -43,21 +43,21 @@ export const PackageContainerStatus: React.FC<{
 				<td className="indent"></td>
 				<td>{packageContainerStatus.containerId}</td>
 				<td>
-					<StatusCodePill
-						connected={device?.connected || false}
-						statusCode={packageContainerStatus.status.status}
-						messages={statusMessages}
-					/>
+					<StatusCodePill connected={device?.connected || false} statusCode={packageContainerStatus.status.status} />
 				</td>
 				<td>
-					<Tooltip overlay={packageContainerStatus.status.statusReason.tech} placement="top">
-						<span>{packageContainerStatus.status.statusReason.user}</span>
+					<Tooltip overlay={statusMessages || packageContainerStatus.status.statusReason.tech} placement="top">
+						<span>{statusMessages || packageContainerStatus.status.statusReason.user}</span>
 					</Tooltip>
 				</td>
 				<td>
-					<button className="btn btn-secondary mls" onClick={(e) => restartPackageContainer(e)}>
-						{t('Restart')}
-					</button>
+					{device?.connected && (
+						<>
+							<button className="btn btn-secondary mls" onClick={(e) => restartPackageContainer(e)}>
+								{t('Restart Container')}
+							</button>
+						</>
+					)}
 				</td>
 			</tr>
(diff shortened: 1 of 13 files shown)


meteor/client/ui/Status/package-status/PackageStatus.tsx
```diff
@@ -128,108 +63,127 @@ export const PackageStatus = withTranslation()(
...
+		[props.statuses]
+	)
+	const requiredModifiedHash = useMemo(() => getModifiedHash(true), [props.statuses])
+	const allModifiedHash = useMemo(() => getModifiedHash(false), [props.statuses])
 
-			const labelRequiredProgress =
-				this.requiredProgress < 1 ? `${Math.floor(this.requiredProgress * 100)}%` : t('Ready')
-			const labelAllProgress = this.allProgress < 1 ? `${Math.floor(this.allProgress * 100)}%` : t('Done')
+	const requiredProgressLastChanged = useMemo(() => Date.now(), [requiredProgress, requiredModifiedHash])
+	const allProgressLastChanged = useMemo(() => Date.now(), [allProgress, allModifiedHash])
```
jstarpl, line 89: In the old React docs, this used to be pointed out as an antipattern, since React does not guarantee that the useMemo computation will always be skipped on a render if the deps don't change. However, I don't see that same notice in the new docs, this isn't exactly critical, and It is much more readable than a `useState` and a `useEffect`, though they too could be combined and hidden in a new `useLastChangedTimestamp()` hook.

meteor/client/ui/Status/package-status/index.tsx
```diff
@@ -43,6 +50,16 @@ export const ExpectedPackagesStatus: React.FC<{}> = function ExpectedPackagesSta
 	const expectedPackages = useTracker(() => ExpectedPackages.find({}).fetch(), [], [])
 	const packageContainerStatuses = useTracker(() => PackageContainerStatuses.find().fetch(), [], [])
 
+	const devices = new Set<PeripheralDeviceId>()
+	packageContainerStatuses.forEach((pcs) => devices.add(pcs.deviceId))
+	expectedPackageWorkStatuses.forEach((epws) => devices.add(epws.deviceId))
+	const deviceIds = Array.from(devices)
```
jstarpl, line 56: I feel like this should be memoized?

meteor/client/ui/Status/package-status/PackageContainerStatus.tsx
```diff
@@ -27,13 +26,28 @@ export const PackageContainerStatus: React.FC<IPackageContainerStatusProps> = fu
 		)
 	}
 
+	const statusMessages: string[] = []
```
jstarpl, line 29: I guess this should be `offlineReasonMessages` or something? Because L40 directly assumes `.length > 0` means "offline"?

meteor/client/ui/Status/package-status/package-status.scss
```diff
@@ -32,6 +32,13 @@
 				}
 			}
 
+			&.offline {
+				> td {
+					color: #666;
+					background-color: #fbb;
+				}
+			}
+
```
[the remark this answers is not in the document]
nytamin: indeed! That was intended as a placeholder, it looks much better now.

PR #942, opened 2023-06-02
feat: move `checkPieceContentStatus` calls to server in custom publication SOFIE-1281

Julusian:

feature

The client is subscribing to various database collections, to perform the checks on the client side. 

These checks are now performed on the server, with the result being fed into a custom publication that the client can consume.
This will reduce the cost of media status on the client, as it doesnt have to compile the data itself, and will help ensure that portions arent overly reactive to changes in the status.

The existing `uiPieceContentStatuses` publication and collection has been extended to also contain statuses for AdlibPiece, AdlibAction, RundownBaselineAdlibItem, RundownBaselineAdlibAction and any non-reset PieceInstances.
A second publication and collection `uiBucketContentStatuses` has been added for buckets. This is partly due to ownership id differences, and also due to needing very different queries to subscribe to the correct data.

The result of this is that the `PackageInfos` collection is no longer used by the client.  
`MediaObjects` is still used by the `ClipTrimPanel`, which sounds like an unused panel.
`PackageContainerPackageStatuses` is still being used to determine the preview url of clips. A follow up task shall be written to refactor that, possibly into the same media status publications.

The `withMediaObjectStatus` HOC has been reworked to use these new publications, modifying any uses of this HOC felt like too much for this PR.

* **Other information**:

I don't have a package manager setup, but I have tested this by manually inserting and editing documents in the database. 

There is quite a bit of repetition in the `regenerateItems.ts` files, but I don't know how to simplify it without using helper functions which take functions as parameters, which I feel will get confusing to follow.

**Status**


nytamin (approved):
I didn't find anything in particular that stood out, looks good 👍

PR #945, opened 2023-06-07
feat: Segment time-anchoring and Break Segments.

alvancamp:

Feature

There's a concept of "breaks" in Sofie, but they only exist _between_ rundowns.

This adds a break concept _within_ a rundown. The idea is that a break is a special segment with the `isBreak` flag set, along with a `breakStartTime` and/or a `breakEndTime`.  Both Forward and Back timing modes should be supported, though because of the dizzying nature of timing, more testing is needed to ensure correctness. This PR is being opened somewhat early so that a formal, logged discussion around it may begin and that incremental feedback may be addressed.

Why add these intra-rundown breaks? The idea is that we want to count down to the next break, not to the end of the rundown. This is how several studios operate. The most important thing is hitting the _next_ wall time.

Also added is a "WALL TIME" display to break segments. It's subtle but it's there. The segment header for non-breaks is unchanged.

image

* **Other information**:

Time makes my head hurt.

**Status**


meteor/client/lib/rundownTiming.ts
```diff
@@ -732,6 +769,10 @@ export interface RundownTimingContext {
 	breakIsLastRundown?: boolean
 	/** Was this time context calculated during a high-resolution tick */
 	isLowResolution: boolean
+	/** The absolute start time of the next break, if any. */
+	nextBreakStartTime?: number
```
jstarpl, line 773: This should not be called `nextBreakStartTime`, since this can be used for things other than breaks. `nextRundownAnchor` maybe?

meteor/client/lib/rundownTiming.ts
```diff
@@ -773,13 +814,25 @@ export function getPlaylistTimingDiff(
...
+		}
+
 		frontAnchor = Math.max(currentTime, playlist.startedPlayback ?? Math.max(timing.expectedStart, currentTime))
 	} else if (PlaylistTiming.isPlaylistTimingBackTime(timing)) {
-		backAnchor = timing.expectedEnd
+		if (timingContext.nextBreakStartTime || timingContext.nextBreakEndTime) {
+			backAnchor = timingContext.nextBreakStartTime ?? timingContext.nextBreakEndTime ?? timing.expectedEnd
+		} else {
+			backAnchor = timing.expectedEnd
+		}
```
jstarpl, line 835: This `if` doesn't do anything that the nullish-coallesing operator doesn't do already.

meteor/client/lib/rundownTiming.ts
```diff
@@ -523,6 +547,17 @@ export class RundownTimingCalculator {
...
+						const backTime = getBreakStartTime(partInstancesMap, segments, this.linearParts[i][0])
+						if (backTime) {
+							nextBreakStartTime = backTime
+						}
+
+						const cumeTime = getBreakEndTime(partInstancesMap, segments, this.linearParts[i][0])
+						if (cumeTime) {
+							nextBreakEndTime = cumeTime
+						}
+					}
```
jstarpl, line 560: Is this top-level `if` block correct? Shouldn't it break at any of these two being set, as opposed to just one?

meteor/client/lib/rundownTiming.ts
```diff
@@ -732,6 +769,10 @@ export interface RundownTimingContext {
 	breakIsLastRundown?: boolean
 	/** Was this time context calculated during a high-resolution tick */
 	isLowResolution: boolean
+	/** The absolute start time of the next break, if any. */
+	nextBreakStartTime?: number
+	/** The absolute end time of the next break, if any. */
+	nextBreakEndTime?: number
```
jstarpl, line 775: Does it make any sense for the calculator to expose both of these? Maybe it would make more sense to expose a single `nextRundownAnchror`?

packages/blueprints-integration/src/rundown.ts
```diff
@@ -166,6 +166,15 @@ export interface IBlueprintSegment<TMetadata = unknown> {
...
 	displayAs?: SegmentDisplayMode
+
+	/** Used to alter the behavior of the Diff timer to make it relative to this segment's hard backtime. If true, `breakStartTime` and/or `breakEndTime` must also be set. */
+	isBreak?: boolean
+
+	/** Does nothing without `isBreak`. */
+	breakStartTime?: number
+
+	/** Does nothing without `isBreak`. */
+	breakEndTime?: number
```
jstarpl, line 177: I think this functionality can be used for so much more than breaks, so I'd rather these properties be called `expectedStart` and `expectedEnd` as is convention in all other timing-related properties in Sofie, since this is what they describe: the "expected time something _should_ happen". It then makes the clear distinction between plan and execution.

jstarpl: Also, I don't believe the comment about `isBreak` needing to be `true` is correct, based on the implementation, and rightfully so. If the user wants to anchor this Segment in time, they can use either of these properties to signify so. I don't see the need for another property. The comments for the two properties could be also made a bit more clear, like: `Expected end time of this Segment`.

meteor/client/ui/SegmentList/SegmentListHeader.tsx
```diff
@@ -102,15 +103,23 @@ export function SegmentListHeader({
...
-							fixed={fixedSegmentDuration}
-						/>
-					)}
+					{playlist &&
+						parts &&
+						parts.length > 0 &&
+						(segment.isBreak ? (
+							<BreakWallTime
+								segment={segment}
+								label={<span className="segment-timeline__duration__label">{t('Wall Time')}</span>}
```
jstarpl, line 112: I don't think "Wall time" is very descriptive. Should be something clearer, maybe "Expected start" or "Expected end"? I suppose if the anchor is `expectedStart`, the user may want to see the Segment Duration still, and this should replace the "On Air In/On Air At" countdown"?

meteor/client/ui/RundownView/RundownTiming/BreakWallTime.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * A presentational component that will render a counter that will show the wall time that a break starts at.
+ * (Or ends at, if there's no start time but an end time is provided.)
+ * @function BreakWallTime
+ * @extends React.Component<IBreakWallTimeProps>
+ */
+export const BreakWallTime = function BreakWallTime(props: IBreakWallTimeProps): JSX.Element | null {
+	const { t } = useTranslation()
+	const value = props.segment.segmentTiming?.expectedStart ?? props.segment.segmentTiming?.expectedEnd
+	const passedExpectedStart =
+		props.segment.segmentTiming?.expectedStart && Date.now() > props.segment.segmentTiming?.expectedStart
```
jstarpl: This should use `getCurrentTime()` as that's our only source of "current time" that is synced across Sofie clients.

Also, I don't think this will currently react to the time flowing, so I suspect you want to use `useCurrentTime()`.

meteor/client/ui/RundownView/RundownTiming/BreakWallTime.tsx
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	const date = new Date(value)
+
+	return (
+		<>
+			<span className={props.labelClassName}>{t(passedExpectedStart ? 'Planned End' : 'Planned Start')}</span>
+			<span className={classNames(props.className)} role="timer">
+				{RundownUtils.padZeros(date.getHours(), 2)}:{RundownUtils.padZeros(date.getMinutes(), 2)}:
+				{RundownUtils.padZeros(date.getSeconds(), 2)}
```
jstarpl: `RundownHeader` currently uses `<Moment interval={0} format="HH:mm:ss" date={expectedEnd} />` to do this sort of thing, I propose we do the same here. NRK has a task to remove all of `Moment` and use something else™️ soon.

meteor/client/ui/SegmentList/SegmentListHeader.tsx
```diff
@@ -121,31 +122,37 @@ export function SegmentListHeader({
...
+										<span className="segment-timeline__timeUntil__label">{t('On Air In')}</span>
+									)
+								}
+							/>
+						)}
+						{studio.settings.preserveUnsyncedPlayingSegmentContents && segment.orphaned && (
+							<span className="segment-timeline__unsynced">{t('Unsynced')}</span>
+						)}
+					</div>
+				)}
```
jstarpl: Looking at this repetitive code, I suspect we should refactor this to a separate component that could be shared across the headers, but I don't think the place to do this is in this PR

meteor/client/ui/SegmentStoryboard/SegmentStoryboard.tsx
```diff
@@ -35,6 +35,7 @@ import { UIStudio } from '../../../lib/api/studios'
 import { PartId, SegmentId } from '@sofie-automation/corelib/dist/dataModel/Ids'
 import { RundownHoldState } from '@sofie-automation/corelib/dist/dataModel/RundownPlaylist'
 import { CalculateTimingsPiece } from '@sofie-automation/corelib/dist/playout/timings'
+import { BreakWallTime } from '../RundownView/RundownTiming/BreakWallTime'
```
jstarpl: Can we rename this component? Something like `SegmentTimeAnchorTime` maybe?

meteor/client/ui/RundownView/RundownTiming/BreakWallTime.tsx
```diff
@@ -0,0 +1,40 @@
...
+	labelClassName?: string
+}
+
+/**
+ * A presentational component that will render a counter that will show the wall time that a break starts at.
+ * (Or ends at, if there's no start time but an end time is provided.)
+ * @function BreakWallTime
+ * @extends React.Component<IBreakWallTimeProps>
+ */
+export const BreakWallTime = function BreakWallTime(props: IBreakWallTimeProps): JSX.Element | null {
```
jstarpl: I think this component should be renamed to reflect that it's not related to Breaks.

jstarpl (changes requested):

jstarpl:
@alvancamp TypeScript is failing because some component props are now missing: https://github.com/nrkno/sofie-core/actions/runs/5258125649/jobs/9501920871?pr=945

Can you fix this, and then I think this is ready to be merged.

PR #957, opened 2023-06-15
fix: timeline and storyboard view break

imaretic:
Bug fix.

Currently, Timeline and Storyboard segment views on the Rundown page won't break in case the viewport size is too small (narrow).

With these changes, both Timeline and Storyboard segments will break the same as the List segment breaks.

* **Other information**:

**Status**


jstarpl (approved):
I'm not sure if this helps much, since I feel that those views break down on such small screens. That being said, if this helps someone, I see no reason not to accept it.

PR #961, opened 2023-06-16
fix!: mark `priority` as a required property in timeline objects, to prevent footguns

alvancamp:

A refactor that guards blueprint authors.

`priority` is optional in timeline objects, leading to footguns where forgetting to set it means that the timeline object can be clobbered by lookaheads, which have a priority of `0.1`.

`priority` is now mandatory and has an explanation in its comment, which will render via Intellisense in the blueprint author's IDE. It suggests setting this value to `1` as a baseline.

* **Other information**:

I've personally introduced several hard-to-track bugs into the blueprints I work on because of this, so there's definite real-world value to this change.

**Status**


jstarpl (approved):

Julusian:
@alvancamp there are a couple of tests still failing, could you take a look?

packages/shared-lib/src/core/model/Timeline.ts
```diff
@@ -43,6 +43,16 @@ export interface TimelineObjectCoreExt<
...
+	/**
+	 * In the underyling timeline library, these are optional. However,
+	 * because of the way that the lookahead system must set a priority,
+	 * and the way in which this can trip up blueprints authors and clobber their intended objects,
+	 * we make them mandatory here.
+	 *
+	 * As of June 16th, 2023, lookaheads have a priority of 0.1. Therefore, to avoid your objects
+	 * being clobbered by lookaheads, you should set a priority higher than this, such as 1.
+	 */
+	priority: number
```
Julusian, line 55: I dont think this needs to be quite this long for a code comment. I would go with something like:

```
Priority of the object. When multiple overlap in the resolved timeline, highest priority wins.
Lookahead objects have a value of 0.1 (or slightly lower). Your baseline should be using a priority of 0, with other objects using 1 or higher. 
```

PR #962, opened 2023-06-16
feat: split package-manager publication into multiple more optimal publications SOFIE-1926

Julusian:

feature

Sofie provides a single publication to package-manager, containing 4 objects of the data it needs.  
Due to how meteor and ddp only being able to do shallow diffs, this means that whenever a single expected package changes, the blob of all of them is sent to package-manager. In a small rundown, this can be over 100kb of json.

Additionally, when updating it was relying less on caches than our other (newer) publications, and more on reloading large amounts of documents from the database.

This single publication has been split out into 3 more targetted publications. Each of these has a more limited reactivity scope, with better caching and tracking of what triggered the invalidation.  
This means that when a single ExpectedPackage changes, the publication knows it only needs to reload and recalculate that single package.  
As the publication is producing smaller documents, one per ExpectedPackage, the data emitted to the package-manager will be kept much smaller than before.

* **Other information**:

Through some of the refactoring, this also reduces the number of duplicate packages being sent to package-manager

**Status**


meteor/server/publications/packageManager/expectedPackages/contentObserver.ts
```diff
@@ -0,0 +1,118 @@
...
+	createReactiveContentCache,
+	rundownPlaylistFieldSpecifier,
+	pieceInstanceFieldsSpecifier,
+} from './contentCache'
+import { ExpectedPackages, PieceInstances, RundownPlaylists } from '../../../collections'
+import { equivalentArrays, waitForPromise } from '../../../../lib/lib'
+import { ReactiveMongoObserverGroup, ReactiveMongoObserverGroupHandle } from '../../lib/observerGroup'
+import _ from 'underscore'
+
+const REACTIVITY_DEBOUNCE = 20
```
jstarpl, line 15: Should the various reactive content caches use the same/more often reuse the same value here? I expect that this may be helpful.

meteor/server/publications/packageManager/expectedPackages/generate.ts
```diff
@@ -0,0 +1,250 @@
...
+import _ from 'underscore'
+import { getSideEffect } from '../../../../lib/collections/ExpectedPackages'
+import { Studio, StudioLight, StudioPackageContainer } from '../../../../lib/collections/Studios'
+import { clone, omit } from '../../../../lib/lib'
+import { CustomPublishCollection } from '../../../lib/customPublication'
+import { logger } from '../../../logging'
+import { ExpectedPackagesContentCache } from './contentCache'
+import type { StudioFields } from './publication'
+
+export async function regenerateForExpectedPackages(
```
jstarpl, line 21: I think it would be good to document what this method is doing. Maybe the name could be more descriptive? regenerate what? Especially, since the "output" of this method, is actually the 4th argument.

meteor/server/publications/packageManager/expectedPackages/generate.ts
```diff
@@ -0,0 +1,250 @@
...
+	deviceId: PeripheralDeviceId,
+	pieceInstanceId: PieceInstanceId | null,
+	priority: Priorities
+): PackageManagerExpectedPackage {
+	// Lookup Package sources:
+	const combinedSources: PackageContainerOnPackage[] = []
+
+	for (const packageSource of expectedPackage.sources) {
+		const lookedUpSource = studio.packageContainers[packageSource.containerId]
+		if (lookedUpSource) {
```
jstarpl, line 150: Maybe this could be reversed, to handle the error condition first?

meteor/server/publications/packageManager/expectedPackages/generate.ts
```diff
@@ -0,0 +1,250 @@
...
+
+				const packageAccessor: AccessorOnPackage.Any | undefined = packageSource.accessors?.[accessorId]
+
+				if (packageAccessor && sourceAccessor && packageAccessor.type === sourceAccessor.type) {
+					combinedSource.accessors[accessorId] = deepExtend({}, sourceAccessor, packageAccessor)
+				} else if (packageAccessor) {
+					combinedSource.accessors[accessorId] = clone<AccessorOnPackage.Any>(packageAccessor)
+				} else if (sourceAccessor) {
+					combinedSource.accessors[accessorId] = clone<Accessor.Any>(sourceAccessor) as AccessorOnPackage.Any
+				}
```
jstarpl, line 174: Maybe this could be refactored to a separate method, to lower the complexity a bit?

meteor/server/publications/packageManager/playoutContext.ts
```diff
@@ -0,0 +1,131 @@
...
+				PackageManagerPlayoutContextArgs,
+				PackageManagerPlayoutContextState,
+				PackageManagerPlayoutContextUpdateProps
+			>(
+				`${PubSub.packageManagerPlayoutContext}_${studioId}_${deviceId}`,
+				{ studioId, deviceId },
+				setupExpectedPackagesPublicationObservers,
+				manipulateExpectedPackagesPublicationData,
+				pub,
+				500 // ms, wait this time before sending an update
```
jstarpl, line 125: Should these magical numbers be `const` somewhere?

jstarpl (approved):
I think maybe the code could be documented a little bit better? The case is esoteric as it is, And the boatload of observers and caches and stuff makes it quite difficult to follow.

PR #965, opened 2023-06-20
ci: add trivy scanning

siljekristensen:
Add Trivy image scanning to the pipeline. This is just informational at this point, it will not break the pipeline. 

Example output: https://github.com/nrkno/sofie-core/actions/runs/5334276977

* **Other information**:

**Status**

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,99 @@
+name: Scheduled Trivy Scan
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 10 * * 1'
+
+jobs:
+  trivy:
+    name: Trivy scan
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image: ["server-core", "playout-gateway", "mos-gateway"]
+    steps:
+      - name: Run Trivy vulnerability scanner (json)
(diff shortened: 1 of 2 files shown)


siljekristensen:
> Should we have this re-running for master on a schedule, so that when new CVEs are reported, we get an updated report?

I could add a separate workflow that only does Trivy scanning, run on a schedule.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,107 @@
...
+    name: Generate SBOM via Trivy 
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Run Trivy in GitHub SBOM mode and submit results to Dependency Graph
+        uses: aquasecurity/trivy-action@0.11.2
+        with:
+          scan-type: 'fs'
```
Julusian, line 18: does this find any dependencies, as they wont have be installed yet?
I expect this will produce a different result to the docker image checks, as it definitely wont be finding some of the dependencies tucked away deep inside of meteor

siljekristensen: You're right, I was fooled by testing it locally where I have all the node_modules. I'll just add SBOM generation from a scan of whatever image is latest.

PR #972, opened 2023-06-26
fix: Duration update while on air trips up autonext (Sofie 1599)

baltedewit:

This PR makes sure a blueprint can not update the length of an autonext part instance to be shorter than the time it has been on-air for

Setting a shorter time is allowed and doing this will cause the playout-gateway to believe the next part should have been playing for x seconds already, which looks _quite_ ugly on air.

* **Other information**:

**Status**

### DIFF
--- a/packages/job-worker/src/blueprints/context/SyncIngestUpdateToPartInstanceContext.ts
+++ b/packages/job-worker/src/blueprints/context/SyncIngestUpdateToPartInstanceContext.ts
@@ -221,6 +221,14 @@ export class SyncIngestUpdateToPartInstanceContext
 
 		if (!this.partInstance) throw new Error(`PartInstance has been removed`)
 
+		// for autoNext, the new expectedDuration cannot be shorter than the time a part has been on-air for
+		if (trimmedProps.expectedDuration && (trimmedProps.autoNext ?? this.partInstance.part.autoNext)) {
+			const onAir = this.partInstance.timings?.reportedStartedPlayback
+			if (onAir && Date.now() - onAir > trimmedProps.expectedDuration) {
+				trimmedProps.expectedDuration = Date.now() - onAir
+			}
+		}
+
 		this._partInstanceCache.updateOne(this.partInstance._id, (p) => {
 			return {
 				...p,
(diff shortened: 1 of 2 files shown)


packages/job-worker/src/blueprints/context/SyncIngestUpdateToPartInstanceContext.ts
```diff
@@ -221,6 +221,14 @@ export class SyncIngestUpdateToPartInstanceContext
 
 		if (!this.partInstance) throw new Error(`PartInstance has been removed`)
 
+		// for autoNext, the new expectedDuration cannot be shorter than the time a part has been on-air for
+		if (trimmedProps.expectedDuration && (trimmedProps.autoNext ?? this.partInstance.part.autoNext)) {
+			const onAir = this.partInstance.timings?.reportedStartedPlayback
+			if (onAir && Date.now() - onAir > trimmedProps.expectedDuration) {
+				trimmedProps.expectedDuration = Date.now() - onAir
```
Julusian, line 228: I think it would be good to add a tolerance here. So that it sets the end to now + 500ms or something.  
My concern is that when this runs, there is still often 80ms+ until TSR has resolved this timeline, so it wont be a bad jump, but TSR will still be playing catchup.

Julusian (approved):

packages/job-worker/src/blueprints/context/SyncIngestUpdateToPartInstanceContext.ts
baltedewit: I agree and had a todo initially, but then I realised it's a bit of a magic number and I'm not sure how to handle that in a good way

Julusian: We already some magic constants in the code, doing things like blocking the next part from changing when within a second of an autonext, so Im not opposed to this being hardcoded.

Alternatively, maybe this should be a setting on the studio? That makes it slightly less hardcoded.
The optimal solution would be something using the latency reported by playout-gateway, but we would still need to factor in the time taken by core, so will be hard to do

PR #983, opened 2023-07-07
feat: remove supertimeline from Parts and Pieces resolving SOFIE-2373

Julusian:

feature

`processAndPrunePieceInstanceTimings` is performed on the Pieces to prune out any pieces which will end up as zero length, calculate start times and a maximum duration from the conflicts between the Pieces on each layer/exclusivity group.

Supertimeline was then used to determine the actual duration from the 3 duration/end properties that each Piece has.

Because `processAndPrunePieceInstanceTimings` is already doing most of the work to 'resolving' the Pieces, we can apply a simple post-step to calculate the accurate durations and replace any 'now' times.

Some bugs have been discovered in the old implementation, fixed by this change.

* **Other information**:

The implementation of `getResolvedPiecesForPartInstancesOnTimeline` mirrors some of the logic done when building the timeline. It isn't directly usable when building the timeline, as it hides any now values, but some further refactoring could split this into two steps so that the logic can be shared with timeline generation.

This has been tested with some basic playout. It has the potential to cause some less obvious playout issues.

The change includes some new unit tests, which were written against the old implementation to help verify the new implementation

**Status**


meteor/client/lib/rundown.ts
```diff
@@ -508,37 +501,26 @@ export namespace RundownUtils {
 					includeDisabledPieces
 				)
 
+				// furthestDuration is used to figure out how much content (in terms of time) is there in the Part
+				let furthestDuration = 0
```
baltedewit, line 505: What is this variable being used for? I can see it being assigned but I don't see where it is used after.

Julusian: I have no idea, it looks to have been abandoned a while ago. It has been removed now

PR #990, opened 2023-07-27
fix(playout-gateway): improve handling and typing of TSR events (fixes some logging issues)

alvancamp:

Bug fix

Some of the event listeners in `playout-gateway` for TSR are wrong, and they're wrong because the types are messed up, which led a previous developer to opt out of types for this section of the code.

This PR restores typing to this section of the code and fixes the relevant type errors, which fixes thing like error logging.

* **Other information**:

This is a hack to work around limitations of the `Promisify` type used by `threadedclass`. Thought should be given to better alternatives before merging this.

**Status**


Julusian (commented):
perhaps this could be done with a helper function, so that there is only one occurance of the 'unnecessary' await. That would make me more confident that it wont get accidentally removed or missed for new uses as the comment will be next to the single usage.

I suspect the typings could become a pain for this, if it becomes too painful I am happy to rethink

```
async function addDeviceEventListener(device, name, cb): Promise<void> {
....
}

```

Julusian:
I have made the changes I was after, but this has conflicted with other changes in release50 so I think it could do with a retest before merging as each change took a different approach to fixing the logging.

nytamin (approved):

PR #993, opened 2023-08-03
fix: Changes the wrong title "Delete" to "Restart".

jesperstarkar:
Also set all other properties of the modal to match a similar modal with the same message, presented somewhere else in the system - for consistency.

Fix

Modal title says "Delete"

Modal title says "Restart"
Also, conforms this modal to another similar one for consistency.

* **Other information**:

**Status**


jesperstarkar:
The wrapping `<p>` has been dropped. Just want a thumbs up from @jstarpl that that is OK and correct.

The new lines are all copied from another place in the code base that produces a similar modal, so now the two end up being identical.

jstarpl (approved):

Issue #857, opened 2023-02-21
RFC: JSON-schema for AdLib Action editable fields


thomasslee97:
The issue has been opened by SuperFly.tv on behalf of EVS Broadcast Equipment (henceforth EVS).

# Motivation

As part of a project being developed by EVS there is a desire to execute AdLib Actions with data provided by an external system. This requires the external system to understand the parameters of an AdLib Action and to have some way of modifying these parameters.

# Proposal

This comes in two parts: How to expose data, and how to execute an action with some data from an external system.

## Exposing Data

Currently, the `IBlueprintActionManifest` interface has the `userDataManifest` property which contains an `editableFields` property. This looks like exactly the property that an external system should be inspecting to know what can and can't be modified and what constitutes valid data.

However, this currently uses the `ConfigManifestEntry` type, which is an "internal" Sofie type, and would require a 1:1 implementation in an external system. As such, we propose that the `editableFields` property instead use JSON-schema to represent the editable data. This follows nicely from recent work that has [moved device config and mappings](https://github.com/nrkno/sofie-core/pull/837) to use JSON-schema.

This would allow the Live-status Gateway from the [API PR](https://github.com/nrkno/sofie-core/pull/777) to expose this JSON-schema on AdLib Actions, providing external systems with a description of the "arguments" to an AdLib Action in a more widely-understood format than the `ConfigManifestEntry` format that Sofie uses.

## Executing Actions

This work follows naturally from the [Stable API](https://github.com/nrkno/sofie-core/issues/773) discussions and would be a change made after the merging of the [open PR](https://github.com/nrkno/sofie-core/pull/777) (we promise!) to modify the `executeAdLib` method to optionally take a JSON object as an argument. For AdLib Pieces, providing a JSON object would result in the rejection of the request. For AdLib Actions, the object could be validated against the schema and used to extend/override the `userData` property passed to `StudioJobs.ExecuteAction`.

## Scope

As far as we can tell, the `editableFields` property is currently only used within the `ShelfInspector` component, which we believe is deprecated, or in the very least rarely used. We do not fully understand this aspect of Sofie, but if it is still in use then the UI components developed by [#837](https://github.com/nrkno/sofie-core/pull/837) should be sufficient as drop-in replacements for the existing `ConfigManifestEntryComponent` used inside of `ActionItemRenderer.tsx`.

The proposed change to the API would be non-breaking, as it would add an optional property to the body of the request, therefore impact to users of the API would be minimal.

# Request For Comment

[...]

PR #1005, opened 2023-08-25
feat!: Editable fields as JSON schema

sbaudlr:

Feature (Closes https://github.com/nrkno/sofie-core/issues/857).

Editable fields use the deprecated config manifest format.

Editable fields now use JSON-schema.

**Other information**:

The ability to change the content of editable fields from the inspector in the shelf has been removed, as this is seemingly unused at the moment. It may be possible to use the components that have been developed for displaying JSON schemas in the config UIs for this purpose if desired, but this has not been tested.

**Status**


Julusian:
I think that this could do with some cleanup to be done too. This should mean that the `ConfigManifestEntry` and related types from blueprints-integration can be removed, as well as the ui components which draw them. I think that cleanup should be done in this PR, as this does make it all dead code.

I do not know enough about the inspector and potential usages of it to say whether it is ok to remove the editable fields section in the ui. But I think it would be better to leave the current ui bits in place, and to replace some of the inside of `renderConfigFields` with something like `<span>Not implemented!</span>`, so that when someone tries to use the `editableFields` property next they don't have to spend time trying to figure out why it isn't working and are instead reminded that it isnt implemented.

meteor/client/ui/Shelf/Inspector/ItemRenderers/ActionItemRenderer.tsx
```diff
@@ -258,13 +233,7 @@ export default translateWithTracker<IProps, {}, ITrackedProps>((props: IProps) =
...
-								<>
-									{this.renderConfigFields(
-										action.userDataManifest.editableFields,
-										targetAction,
-										'transformedUserData.'
-									)}
-								</>
+								<span>Editable Fields are not currently supported.</span>
 							) : null}
 						</div>
```
jstarpl, line 238: I guess the "Editable Fields ..." text should be passed through the `t` function, but more importantly: does this section make any sense now? Should we just not show anything, instead of showing a spinner only to then later show "Editable Fields are not currently supported"?

jstarpl (changes requested):

meteor/client/ui/Shelf/Inspector/ItemRenderers/ActionItemRenderer.tsx
Julusian: I suggested doing this:

> But I think it would be better to leave the current ui bits in place, and to replace some of the inside of renderConfigFields with something like <span>Not implemented!</span>, so that when someone tries to use the editableFields property next they don't have to spend time trying to figure out why it isn't working and are instead reminded that it isnt implemented.

PR #1006, opened 2023-08-28
chore: add some architecture and PartInstance documentation

Julusian:

documentation

* **Other information**:

**Status**


mint-dewit (approved):
Awesome, this adds some great information to the docs.

Perhaps a note on the usage of the live status gateway as a means to receive real time information might be nice. I'm just thinking of a reader who wants to write a custom application and ends up on the publications page. They might end up getting away with the idea that they need to implement a DDP client when the live status gateway will probably be easier for them.

PR #1007, opened 2023-08-28
fix: zero duration pieces breaking ab resolver

Julusian:

bug fix

Sofie resolves the piece timeline, to be able to provide concrete times to the ab resolved, `onTimelineGenerate` and other places.  
If a piece is resolved to have a duration of 0, that gets lost during the resolving, so that it ends up with no duration.

It now treats a duration of 0 as a valid value, preserving it rather than replacing it with undefined.

* **Other information**:

Some minor tweaks have been made so that lookahead ab sessions use a start time that is json serializable, as this confused us while debugging due to infinity being converted to `null` when stringified for json

**Status**


mint-dewit (approved):

Julusian:
As this area has been rewritten in r51 by #983, I have added a test to confirm it is working there https://github.com/nrkno/sofie-core/commit/13c881ded38ad74c1689be5a8430d0fc824ca1b3. It was already fixed by the changes in #983

PR #1020, opened 2023-09-13
Promise button (SOFIE-2573)

nytamin:
Adds a new component, the PromiseButton

No feedback in the button itself when clicking Take a Snapshot

* The button is disabled and indicates that we're waiting for the promise to resolve.
* The button animates and indicates whether the action succeeded or not:

ezgif-4-eaf2830c68


meteor/client/lib/Components/PromiseButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+				setActionResult(undefined)
+			}, 3000)
+			return () => {
+				clearTimeout(timeout)
+			}
+		}
+	}, [isPending, actionResult])
+
+	const overlayContent = isPending ? (
+		<FontAwesomeIcon icon={faHourglass} />
```
jstarpl, line 52: I don't think this should use the hourglass icon. The hourglass means "Reserved clip". For processes in action, I think we should use a spinner, as that's what we use to convey "working".

jstarpl (changes requested):

meteor/client/lib/Components/PromiseButton.tsx
nytamin: Good observation, I've changed the hourglass to the Spinner now!

hummelstrand (approved):

PR #1036, opened 2023-09-27
fix: AdLibs don't appear in Live Status Gateway until first take

sbaudlr:

Bug fix.

The Live Status Gateway doesn't populate a list of AdLibs until the first Part of a Rundown is taken.

The list is now populated on activate.

**Other information**:

This PR also includes a fix for a bug where the Live Status Gateway would not restart when requested.

**Status**

### DIFF
--- a/packages/live-status-gateway/src/collections/adLibActions.ts
+++ b/packages/live-status-gateway/src/collections/adLibActions.ts
@@ -42,7 +42,7 @@ export class AdLibActionsHandler
 		await new Promise(process.nextTick.bind(this))
 		if (!this._collectionName) return
 		if (!this._publicationName) return
-		if (prevRundownId !== this._curRundownId || prevCurPartInstance !== this._curPartInstance)) {
+		if (prevRundownId !== this._curRundownId || prevCurPartInstance !== this._curPartInstance) {
 			if (this._subscriptionId) this._coreHandler.unsubscribe(this._subscriptionId)
 			if (this._dbObserver) this._dbObserver.stop()
 			if (this._curRundownId && this._curPartInstance) {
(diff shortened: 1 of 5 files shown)


packages/live-status-gateway/src/collections/adLibActions.ts
```diff
@@ -34,15 +35,17 @@ export class AdLibActionsHandler
...
-		this._curRundownId = data ? unprotectString(data.get(PartInstanceName.current)?.rundownId) : undefined
+		const prevCurPartInstance = this._curPartInstance
+		this._curPartInstance = data ? data.get(PartInstanceName.current) ?? data.get(PartInstanceName.next) : undefined
+		this._curRundownId = this._curPartInstance ? unprotectString(this._curPartInstance.rundownId) : undefined
 
 		await new Promise(process.nextTick.bind(this))
 		if (!this._collection) return
 		if (!this._publication) return
-		if (prevRundownId !== this._curRundownId) {
+		if (!(prevRundownId === this._curRundownId && prevCurPartInstance === this._curPartInstance)) {
```
Julusian, line 45:
```suggestion
		if (prevRundownId !== this._curRundownId || prevCurPartInstance !== this._curPartInstance)) {
```
I think this is easier to quickly interpret

Julusian: Should this be comparing object references of `_curPartInstance`, or should it do a deeper equality check?
It looks like it only really needs the `PartId` of the PartInstance. Should it store and compare just that value instead to reduce the amount of invalidations?

packages/live-status-gateway/src/collections/adLibs.ts
Julusian, line 1: see comments for `adlibActions.ts`

Julusian (approved):

PR #1040, opened 2023-09-29
fix(live-status-gw): AdLib Pieces were missing in `activePlaylist` topic

ianshade:

Bug fix: exposing previously missing data

AdLib Pieces and Rundown Baseline AdLib Pieces are missing in their respective arrays in the events broadcast by the `activePlaylist` topic. Only AdLib Actions (incl. baseline) are present.

The missing data is now present: `AdLibStatus` objects coresponding to AdLib Pieces and Rundown Baseline AdLib Pieces are now available to subscribers of the `activePlaylist` topic.

* **Other information**:
Note: The same convention for replacing magic strings with class names is used as in https://github.com/nrkno/sofie-core/pull/1037

**Status**


packages/live-status-gateway/src/topics/activePlaylist.ts
```diff
@@ -287,6 +289,18 @@ export class ActivePlaylistTopic
 				this._globalAdLibActions = globalAdLibActions
 				break
 			}
+			case AdLibsHandler.name: {
```
Julusian, line 292: I feel like it would be a good idea to be consistent with the other cases for now and use a fixed string.
I do agree that something like this is a good idea, but I think it might need to be done as an enum. Importing the other files like this could result in an import loop which can cause issues

ianshade: An import loop is something I have not considered. I hope eventually those names won't be needed. I'd like us to eventually get rid of those large switches that needs a type assertion in every case, and do it in a way that makes easy to make sure the necessary piping is set up, which with the current implementation in `liveStatusServer.ts` is not.

ianshade: I think a circular dependency is actually unlikely, because the data flow in the handlers and topics is designed to be unidirectional.

ianshade:
I resolved the conflicts.
Now the [Files changed](https://github.com/nrkno/sofie-core/pull/1040/files) tab reflects exactly what comes in this PR.

Julusian (approved):

PR #1042, opened 2023-09-29
fix: CasparCG restart error: Timeout after 10000ms SOFIE-2588

Julusian:

fix

The handling of cronjob CasparCG restarts is improved to produce more better errors, and minimise logging at the error level.

Additionally, some typings for the ui have improved for executing tsr actions.

This relies on https://github.com/nrkno/sofie-timeline-state-resolver/pull/295 for accurate logging

* **Other information**:

**Status**

### DIFF
--- a/meteor/server/cronjobs.ts
+++ b/meteor/server/cronjobs.ts
@@ -171,13 +171,15 @@ async function restartCasparCG(system: ICoreSystem | undefined, previousLastNigh
 		)
 	}
 
-	await Promise.all(ps)
-
-	if (attemptFailed && failedRetries < 5) {
-		failedRetries++
-		lastNightlyCronjob = previousLastNightlyCronjob // try again later
-	} else {
-		failedRetries = 0
+	try {
+		await Promise.all(ps)
+	} finally {
+		if (attemptFailed && failedRetries < 5) {
+			failedRetries++
+			lastNightlyCronjob = previousLastNightlyCronjob // try again later
+		} else {
+			failedRetries = 0
+		}
 	}
 }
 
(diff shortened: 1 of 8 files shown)


meteor/server/cronjobs.ts
```diff
@@ -70,59 +79,101 @@ async function cleanupOldDataCronjob() {
...
+		}
+
+		if (!device.parentDeviceId) {
+			logger.info(`Cronjob: Skipping CasparCG device "${device._id}" without parentDeviceId`)
+			// Misconfiguration, don't retry
+			continue
+		}
+		const parentDevice = deviceMap.get(device.parentDeviceId)
+		if (!parentDevice) {
+			logger.info(`Cronjob: Skipping CasparCG device "${device._id}" without missing parent device`)
```
jstarpl, line 123:
```suggestion
			logger.info(`Cronjob: Skipping CasparCG device "${device._id}" with a missing parent device`)
```

meteor/server/cronjobs.ts
```diff
@@ -70,59 +79,101 @@ async function cleanupOldDataCronjob() {
...
+		logger.info(`Cronjob: Trying to restart CasparCG on device "${device._id}"`)
+
+		ps.push(
+			executePeripheralDeviceFunctionWithCustomTimeout(device._id, DEFAULT_TSR_ACTION_TIMEOUT_TIME, {
+				actionId: TSR.CasparCGActions.RestartServer,
+				payload: {},
+			})
+				.then((res) => {
+					if (res.result === TSR.ActionExecutionResultCode.Ok) {
+						logger.info(`Cronjob: "${device._id}": CasparCG restart done`)
```
jstarpl, line 143: Shouldn't a succesful restart of one device be recorded, so that it can be skipped if another device fails and `shouldRetryAttempt` be true? Otherwise, won't be restarting all CasparCG multiple times, even though only one of them fails to restart?

jstarpl (changes requested):

meteor/server/cronjobs.ts
Julusian: Yes, I agree that it should be more intelligent. But I would class that as being more of a feature than a fix, so I wrote up a new task to improve that SOFIE-2684. I think it will end up being a bit fiddly, so am not comfortable doing that as part of a fix for r50.

jstarpl:
I think the tests still need fixing: https://github.com/nrkno/sofie-core/actions/runs/6430122872/job/17460498391?pr=1042

Julusian:
@jstarpl fixed

PR #1050, opened 2023-10-10
feat: blueprint config fixup step SOFIE-2258

Julusian:

feature

This builds on #1049 

I am not entirely convinced by the naming, so am open to suggestions on a better name

When using the new blueprint config upgrade flow #808 instead of the old migrations flow, it is not possible to handle restructuring of the config without some external tooling or manual fixup of the data.

It is now possible to provide a method to perform some fixup of the users configuration overrides.  
This method is optional, and if defined must be run after uploading blueprints before the config can be edited or re-validated.

It has been added as an explicit step as it has the ability to show warning messages to the user, some of which might need manual addressing. It is also possible to ignore/skip the run of fixupConfig, intended to be an escape hatch in case the fixupConfig function gets stuck in an error loop.  
This utilises the changed from #1049 for this, so that the fixup function can be triggered from both the blueprint config page, and from the main migrations page. When the fixup needs running, the configuration section is hidden, to ensure changes are not made.

Once run, it is possible to edit configuration and apply/validate the config as usual.

Because this is modifying an object which is of type `ObjectWithOverrides<T>`, we don't want to expose it directly to the blueprints. In particular, we don't want to expose every type of operation as that will make it more breaking when we modify/add more operation types. Instead we provide an abstract view of this via the context.  
This means the blueprints are aware that the key identifiers are paths, but they must perform mutations via function calls so that we can update the overrides correctly.

* **Other information**:

Some examples of things that could be done.

Fix a typo in a property name
```
export function fixUpConfig(context: IFixUpConfigContext<ShowStyleConfig>): void {
	if (context.hasOperations('MediaManagetBasePath')) {
		if (context.hasOperations('MediaManagerBasePath')) {
			// Both exist, discard the incorrect one
			context.removeOperations('MediaManagetBasePath')
		} else {
			context.renameOperations('MediaManagetBasePath', 'MediaManagerBasePath')
		}
	}
}
```

**Status**


meteor/server/migration/upgrades/showStyleBase.ts
```diff
@@ -15,39 +19,86 @@ import { evalBlueprint } from '../../api/blueprints/cache'
 import { logger } from '../../logging'
 import { CommonContext } from './context'
 import type { AnyBulkWriteOperation } from 'mongodb'
+import { FixUpBlueprintConfigContext } from '@sofie-automation/corelib/dist/fixUpBlueprintConfig/context'
+import { Blueprint } from '@sofie-automation/corelib/dist/dataModel/Blueprint'
+import { BlueprintFixUpConfigMessage } from '../../../lib/api/migration'
 
-export async function validateConfigForShowStyleBase(
+export async function fixupConfigForShowStyleBase(
```
mint-dewit, line 26: Why is the studio config fixup step ran from the worker whereas the showstyle step is ran from meteor?

meteor/client/ui/Settings/Upgrades/Components.tsx
```diff
@@ -152,21 +174,117 @@ export function UpgradeStatusButtons({ upgradeResult }: UpgradeStatusButtonsProp
...
+			acceptOnly: true,
+			yes: t('Confirm'),
+			onAccept: () => {
+				ignoreFixupConfig()
+					.then(() => {
+						NotificationCenter.push(
+							new Notification(
+								undefined,
+								NoticeLevel.NOTIFICATION,
+								t('Config for {{name}} fixed successfully', { name: upgradeResult.name }),
```
mint-dewit, line 194: This says the config has been fixed when in fact the fix step has been skipped.

mint-dewit (approved):

meteor/server/migration/upgrades/showStyleBase.ts
Julusian: Following the pattern I started with the other config methods.

I am not 100% sold on it anymore, but my original justification was that we have the job-worker which handles the studio, and doing it there will avoid needing to load the studio blueprints in meteor.  
But a showstyle can be used by multiple studios, so can't be pushed to the job-worker currently.

PR #1051, opened 2023-10-13
feat: support packageinfo update flow for buckets and studio baseline SOFIE-2655

Julusian:

feature

Rundown pieces and adlibs are able to register a dependency on an expected package, and be regenerated when the packageinfos for that package change.  
This was not implemented for buckets

It is now possible for bucket items to register a dependency on an expected package, and be regenerated when the packageinfos for that package change.  

Prior to this we were not storing the ingest data used to generate a bucket item. This is now being done on each bucket item, which is a bit messy, but unless we want a new collection I am not sure how better to do it.

* **Other information**:

**Status**


meteor/server/publications/buckets.ts
```diff
@@ -45,7 +45,9 @@ meteorPublish(
 meteorPublish(PubSub.bucketAdLibPieces, async function (selector: MongoQuery<BucketAdLib>, _token: string | undefined) {
 	if (!selector) throw new Meteor.Error(400, 'selector argument missing')
 	const modifier: FindOptions<BucketAdLib> = {
-		fields: {},
+		fields: {
+			ingestInfo: 0,
```
nytamin, line 49: Minor detail: Perhaps add a comment explaining why this is excluded?

nytamin (approved):

PR #1056, opened 2023-10-26
feat: typed publications in gateways/peripheraldevices SOFIE-1183

Julusian:

feature

Subscribing to publications in peripheraldevices is loosely typed, and requires each gateway to know the type of each collection it is accessing.

the `PubSub` enum has been split up to reflect what is available at different levels, and each gateway is able to define what subset it wants to access, resulting in the correct typings being available for publications.  
One portion resides in `shared-lib`, and is available to all peripheral devices (and is the default, unless they specify another set).  
Another portion resides in `corelib`, and contains publications for the Rundown/RundownPlaylist and its contents. This is intended to be used by the live-status gateway.
The final portion resides in the old location, and is intended to be used solely by the Meteor UI. The Meteor UI also has access to the rest of the publication types.

As part of this, some typings had to be adjusted inside LSG to make the compiler happy, resulting in it having some more usage of our id types, and some additional generics to get the publications typed properly.

Additionally, the `observe` method in server-core-integration has been made typed to match the collections. This did not have much impact and has improved type safety

* **Other information**:

I have prototyped the required changes for package-manager and input-gateway, PRs will be raised when this gets merged.

**Status**

### DIFF
--- a/meteor/server/api/rest/v0/index.ts
+++ b/meteor/server/api/rest/v0/index.ts
@@ -115,13 +115,13 @@ export function createLegacyApiRouter(): KoaRouter {
 	}
 
 	// Expose publications:
-	for (const [pubName, pubValue] of Object.entries<any>(MeteorPubSub)) {
+	for (const [pubName, pubValue] of Object.entries<string>(MeteorPubSub)) {
 		exposePublication(pubName, pubValue)
 	}
-	for (const [pubName, pubValue] of Object.entries<any>(CorelibPubSub)) {
+	for (const [pubName, pubValue] of Object.entries<string>(CorelibPubSub)) {
 		exposePublication(pubName, pubValue)
 	}
-	for (const [pubName, pubValue] of Object.entries<any>(PeripheralDevicePubSub)) {
+	for (const [pubName, pubValue] of Object.entries<string>(PeripheralDevicePubSub)) {
 		exposePublication(pubName, pubValue)
 	}
 
--- a/meteor/server/api/rest/v0/index.ts
+++ b/meteor/server/api/rest/v0/index.ts
@@ -7,13 +7,15 @@
 import * as _ from 'underscore'
 import { Meteor } from 'meteor/meteor'
 import { MeteorMethodSignatures } from '../../../methods'
-import { PubSub } from '../../../../lib/api/pubsub'
+import { MeteorPubSub } from '../../../../lib/api/pubsub'
 import { MeteorPublications, MeteorPublicationSignatures } from '../../../publications/lib'
 import { UserActionAPIMethods } from '../../../../lib/api/userActions'
 import { logger } from '../../../../lib/logging'
 import { ClientAPI } from '../../../../lib/api/client'
 import Koa from 'koa'
 import KoaRouter from '@koa/router'
+import { CorelibPubSub } from '@sofie-automation/corelib/dist/pubsub'
+import { PeripheralDevicePubSub } from '@sofie-automation/shared-lib/dist/pubsub/peripheralDevice'
 
 const LEGACY_API_VERSION = 0
 
@@ -82,8 +84,7 @@ export function createLegacyApiRouter(): KoaRouter {
 		})
 	}
 
-	// Expose publications:
-	for (const [pubName, pubValue] of Object.entries<any>(PubSub)) {
+	function exposePublication(pubName: string, pubValue: string) {
 		const signature = MeteorPublicationSignatures[pubValue] || []
 
 		const f = MeteorPublications[pubValue]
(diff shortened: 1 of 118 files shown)


meteor/server/api/rest/v0/index.ts
```diff
@@ -113,6 +114,17 @@ export function createLegacyApiRouter(): KoaRouter {
 		}
 	}
 
+	// Expose publications:
+	for (const [pubName, pubValue] of Object.entries<any>(MeteorPubSub)) {
```
nytamin, line 118: Could we use something else than the `<any>` here? `<string>`?

meteor/server/publications/_publications.ts
```diff
@@ -23,3 +24,26 @@ import './translationsBundles'
...
+			if (!MeteorPublications[pubName]) {
+				logger.error(`Publication "${pubName}" is not setup!`)
+			}
+		}
+	}
+
+	Meteor.startup(() => {
+		checkPublications(Object.values<string>(MeteorPubSub))
+		checkPublications(Object.values<string>(CorelibPubSub))
+		checkPublications(Object.values<string>(PeripheralDevicePubSub))
```
nytamin, line 47: This is slightly dangerous in the future, since someone might add another PubSub and forget to add it here.
Is there a way to expose some joined enum or list of values (like a `AllPubSub` from `lib/api/pubsub` and use that instead?

meteor/server/publications/rundown.ts
```diff
@@ -367,9 +375,29 @@ meteorPublish(
 		return null
 	}
 )
+
+meteorPublish(
+	PeripheralDevicePubSub.expectedPlayoutItemsForDevice,
+	async function (deviceId: PeripheralDeviceId, token: string | undefined) {
+		if (!deviceId) throw new Meteor.Error(400, 'deviceId argument missing')
```
nytamin, line 382: Should this publication use a better authentication? Like in `PeripheralDevicePubSub.rundownsForDevice`?

https://github.com/nrkno/sofie-core/pull/1056/files#diff-0a1e64f58116bd28a7682f7327410f9d24631a4b0c877257e54eeecb917cb4f6R59

packages/corelib/src/pubsub.ts
```diff
@@ -0,0 +1,209 @@
...
+import { TimelineComplete } from './dataModel/Timeline'
+import {
+	RundownId,
+	RundownPlaylistId,
+	ShowStyleBaseId,
+	StudioId,
+} from '@sofie-automation/shared-lib/dist/core/model/Ids'
+import { RundownPlaylistActivationId } from './dataModel/Ids'
+
+export enum CorelibPubSub {
```
nytamin, line 37: We should write a desciption for this. Something like "List of the publications used by the Sofie Core GUI" or similar?

packages/playout-gateway/src/coreHandler.ts
```diff
@@ -206,43 +202,6 @@ export class CoreHandler {
 			const studioId = device?.studioId
 			if (!studioId) throw new Error('PeripheralDevice has no studio!')
 
-			if (studioId !== this._studioId) {
```
nytamin, line 209: Getting rid of this tech debt feels good, right? :)

packages/corelib/src/pubsub.ts
```diff
@@ -0,0 +1,209 @@
...
+
+	externalMessageQueue = 'externalMessageQueue',
+
+	blueprints = 'blueprints',
+	showStyleBases = 'showStyleBases',
+	showStyleVariants = 'showStyleVariants',
+	studios = 'studios',
+
+	peripheralDevices = 'peripheralDevices',
+	peripheralDevicesAndSubDevices = ' peripheralDevicesAndSubDevices',
```
nytamin, line 76: typo? (there's a space in there)

packages/server-core-integration/README.md
```diff
@@ -72,7 +72,7 @@ ddp.connect((err: any) => {
 		return
 	}
 	let subId = ddp.subscribe('expectedMediaItems')
```
nytamin, line 74: this should be a PubSub reference, not a string, right?

packages/server-core-integration/README.md
```diff
@@ -72,7 +72,7 @@ ddp.connect((err: any) => {
...
-	ddp.observe('expectedMediaItems', /* added cb */, /* changed cb */, /* removed cb */)
+	ddp.observe(PeripheralDevicePubSubCollectionsNames.expectedMediaItems, /* added cb */, /* changed cb */, /* removed cb */)
 
 	/* ... then later ... */
```
nytamin, line 77: While you're at it, could you add to the example an example of using the collection data as well?

packages/corelib/src/pubsub.ts
Julusian: That appears to be an existing typo, but it may as well be fixed now

packages/server-core-integration/README.md
Julusian: Actually no. This is the `DDPClient`, describing how to use the connection class independently from Sofie. 
Instead I've changed the line below back to not be using an enum

nytamin (approved):

PR #1063, opened 2023-11-01
fix: Playout ignores 'Play from here' offset SOFIE-2356

Julusian:

bug fix

Using the 'play from here' context menu option will always play from the start of the selected part

It now plays from the correct point within the part

* **Other information**:

**Status**


nytamin (approved):
a simple and elegant solution!

PR #1070, opened 2023-11-13
feat: Buckets in the HTTP API (SOFIE-2795)

ianshade:

Exposes existing features regarding Buckets over the HTTP API

Buckets and Bucket AdLib management is only available over DDP. Only execution of bucket adlibs is available over HTTP.

The following features are made available over HTTP:

  - Buckets can be fetched, created, deleted and emptied.
  - Bucket AdLibs can be imported (through the blueprint ingest flow) and deleted.

* **Other information**:
It is assumed that validation of the payload when importing an adlib is performed by the blueprints.

**Status**


packages/openapi/api/definitions/buckets.yaml
```diff
@@ -0,0 +1,292 @@
...
+          $ref: '#/components/responses/bucketNotFound'
+        500:
+          $ref: '#/components/responses/internalServerError'
+  bucketAdlib:
+    delete:
+      operationId: deleteBucketAdlib
+      tags:
+        - buckets
+      summary: Deletes a bucket adlib
+      parameters:
```
Julusian, line 177: It looks like this should have an id for the item to be removed?

meteor/lib/api/rest/v1/buckets.ts
```diff
@@ -0,0 +1,124 @@
...
+	 * Deletes a Bucket AdLib.
+	 *
+	 * @param connection Connection data including client and header details
+	 * @param event User event string
+	 * @param adLibId Id of the bucket adlib to delete
+	 */
+	deleteBucketAdLib(
+		connection: Meteor.Connection,
+		event: string,
+		adLibId: BucketAdLibId | BucketAdLibActionId
```
Julusian, line 86: I wonder if this is the correct id to use, of if it should be the `externalId`.

Internally, when the import happens it can generate one or more bucket adlibs, one for each ShowStyleVariant. We 'group' them by the externalId when necessary.  
Perhaps this api should do the same?

This is nothing like how the operation is done today, but the current implementation is not great in this regard. It would be nice to not bake this limitation/quirk into the stable api too

Julusian (commented):
I don't see anything here for retrieving the contents of a bucket. Should that be here?

When it comes to exposing the adlibs in the LSG, I would suggest considering grouping the different versions of an adlib up as one document with 'variants'. I am not sure how that will look (or if it will work well), but the way we have multiple documents for each bucket item was a performance problem for us at one point.

packages/openapi/api/definitions/buckets.yaml
ianshade: Updated.
Technically the bucketId is not needed because adlibId is already unique, but I'm not sure how the path to the resource should look like then

ianshade:
@Julusian Those are some very good points about the externalId and variants. Do you agree that when executing a bucket adlib it should also be possible to do so by the `externalId`? I'd say that it should be the backend's responsibility to find the actual adlib by given externalId, that is also matching the current variant (if set). In that case we could use the externalId as THE `id` for bucket adlibs both in the LSG and the HTTP API. But that would be a breaking change in regards to playlist's execute-adlib endpoint.

> I don't see anything here for retrieving the contents of a bucket. Should that be here?

I'm working under the assumption that since regular adlibs can't be retrieved over HTTP either, the LSG will be used for that purpose, but I'm open to adding it if you think that's necessary.

meteor/server/api/rest/v1/buckets.ts
```diff
@@ -0,0 +1,280 @@
...
+			logger.info(`API DELETE: Empty Bucket`)
+			const bucketId = protectString(params.bucketId)
+			check(bucketId, String)
+			return await serverAPI.emptyBucket(connection, event, bucketId)
+		}
+	)
+
+	registerRoute<{ adLibId: string }, never, void>(
+		'delete',
+		'/buckets/:bucketId/adlibs/:adLibId',
```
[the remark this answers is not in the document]
ianshade: This has been updated in the code and API documentation

meteor/lib/api/rest/v1/buckets.ts
ianshade: This has been updated to allow deleting by externalId

ianshade:
Thank you all for your input! I believe I have now addressed all the comments.

meteor/server/api/rest/v1/playlists.ts
```diff
@@ -210,6 +213,57 @@ class PlaylistsServerAPI implements PlaylistsRestAPI {
...
+	async executeBucketAdLib(
+		connection: Meteor.Connection,
+		event: string,
+		rundownPlaylistId: RundownPlaylistId,
+		bucketId: BucketId,
+		externalId: string,
+		triggerMode?: string | null
+	): Promise<ClientAPI.ClientResponse<object>> {
+		const bucketPromise = Buckets.findOneAsync({}, { projection: { _id: 1, name: 1, studioId: 1 } })
+		const bucketAdlibPromise = BucketAdLibs.findOneAsync({ bucketId, externalId }, { projection: { _id: 1 } })
```
Julusian, line 225: I think it would be better to `await Promise.all(` these promises.

If a promise rejects but it was not awaited or had `.catch(` used on it, then it becomes an unhandledPromiseRejection and can result in an application crash

packages/job-worker/src/playout/adlibAction.ts
```diff
@@ -38,81 +38,93 @@ export async function handleExecuteAdlibAction(
...
 		const initCache = await loadPlayoutModelPreInit(context, lock, playlist, false)
 
-		const rundown = initCache.getRundown(playlist.currentPartInfo.rundownId)
-		if (!rundown) throw new Error(`Current Rundown "${playlist.currentPartInfo.rundownId}" could not be found`)
+		return executeAdlibAction(context, playlist, initCache, data)
+	})
+}
 
-		const showStyle = await context.getShowStyleCompound(rundown.showStyleVariantId, rundown.showStyleBaseId)
+export async function executeAdlibAction(
```
Julusian, line 45:
```suggestion
export async function executeAdlibActionAndSaveModel(
```
This takes the uncommon step of saving the model too, almost everywhere else it the caller that is expected to save the model

packages/job-worker/src/playout/bucketAdlibJobs.ts
```diff
@@ -0,0 +1,94 @@
...
+			rundown.showStyleBaseId,
+			rundown.showStyleVariantId
+		)
+
+		if (bucketAdLib) {
+			if (playlist.holdState === RundownHoldState.ACTIVE || playlist.holdState === RundownHoldState.PENDING) {
+				throw UserError.create(UserErrorMessage.DuringHold)
+			}
+
+			const playoutModel = await createPlayoutModelfromInitModel(context, initCache)
```
Julusian, line 43: I don't think this ever gets saved back to mongo?

Julusian (approved):

meteor/server/api/rest/v1/playlists.ts
ianshade: I also think I used the wrong selector for the buckets here...

packages/job-worker/src/playout/adlibAction.ts
ianshade: Updated

packages/job-worker/src/playout/bucketAdlibJobs.ts
ianshade: I added a call saving the model at the end

meteor/server/api/rest/v1/playlists.ts
ianshade: Updated

ianshade:
I have now addressed the remaining comments, resolved merge conflicts, and fixed a bug that I found (incorrect selectors when fetching buckets).

PR #1083, opened 2023-12-08
chore: refactor SegmentTimelineContainer to avoid `MeteorReactComponent` SOFIE-2751

Julusian:
## About the Contributor

## Type of Contribution

This is a: Bug fix

## New Behavior

## Status

### DIFF
--- a/meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
+++ b/meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
@@ -1,19 +1,16 @@
-import * as React from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import * as PropTypes from 'prop-types'
 import * as _ from 'underscore'
 import { SegmentTimeline, SegmentTimelineClass } from './SegmentTimeline'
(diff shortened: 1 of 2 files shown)


meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
```diff
@@ -79,8 +77,90 @@ interface IProps extends IResolvedSegmentProps {
...
+					return partInstanceIds
+				}
+			})
+		}, 40)
+
+		return () => {
+			// Abort if changed again before it fires
+			clearTimeout(timeout)
+		}
+	}, [partInstanceIds])
```
jstarpl, line 136: We already have a `useDebounce` hook in `client/lib/lib.tsx`, maybe we could have an additional version of that hook with a comparator function (that this would feed `equivalentArrays` as), to shorten the body of this function?

jstarpl (approved):

meteor/client/ui/SegmentTimeline/SegmentTimelineContainer.tsx
Julusian: Good idea. This is done in 21f7bb00a202c5b8fd808ae47eaeebe11598b2b3

PR #1084, opened 2023-12-08
chore: refactor `RundownSystemStatus` to avoid `MeteorReactComponent` SOFIE-2751

Julusian:
## About the Contributor

## Type of Contribution

This is a bug fix

## New Behavior

## Status


nytamin (approved):
Looks good.
I do think that it would look nicer to not mix functional and class components, perhaps refactor `MOSLastUpdateStatus` and `RundownSystemStatusContent` into functional components as well?

Julusian:
> I do think that it would look nicer to not mix functional and class components, perhaps refactor MOSLastUpdateStatus and RundownSystemStatusContent into functional components as well?

I agree, and I was going to say that I think it is out of scope for this as they both need more thought while refactoring (to understand internal state flows). But they were both easier than I expected. The state of `RundownSystemStatusContent` was not actually used, and there is already the `useCurrentTime` needed for `MOSLastUpdateStatus`, so I have done them now

PR #1085, opened 2023-12-08
chore: refactor PrompterView to avoid MeteorReactComponent SOFIE-2751

Julusian:
## About the Contributor

## Type of Contribution
This is a bug fix

## New Behavior

## Status

### DIFF
--- a/meteor/client/ui/Prompter/PrompterView.tsx
+++ b/meteor/client/ui/Prompter/PrompterView.tsx
@@ -101,419 +101,418 @@ function asArray<T>(value: T | T[] | null): T[] {
 	}
 }
 
-export const PrompterViewContent = withTranslation()(
-	class PrompterViewContent extends React.Component<Translated<IProps & ITrackedProps>> {
-		autoScrollPreviousPartInstanceId: PartInstanceId | null = null
+export class PrompterViewContent extends React.Component<Translated<IProps & ITrackedProps>> {
+	autoScrollPreviousPartInstanceId: PartInstanceId | null = null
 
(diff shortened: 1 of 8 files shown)


meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -543,27 +510,65 @@ export class PrompterViewInner extends MeteorReactComponent<Translated<IProps &
...
+				}
+			) as Omit<DBRundownPlaylist, 'trackedAbSessions'> | undefined,
+		[props.studioId]
+	)
+
+	return (
+		<PrompterViewInner
+			{...props}
+			t={i18nProps.t}
+			i18n={i18nProps.i18n}
```
jstarpl, line 564: This seems weird, wouldn't it make more sense to provide `t`, `i18n`, `tReady` via a `withTranslation` HOC (i.e. replacing the `translateWithTracker` with a `withTranslation`?

jstarpl (approved):

meteor/client/ui/Prompter/PrompterView.tsx
Julusian: I think I did this because doing it with `withTranslation` requires doing that wrap in its own line, as the `PromperViewInner` is exported and types break if wrapped with `withTranslation`.
And this way avoid the extra wrapping component that `withTranslation` adds.

But I don't feel strongly that it should be done this way, and do agree that the manual props feels a little messy, so I've changed it to have the line `const PrompterViewContentWithTranslation = withTranslation()(PrompterViewContent)` which then gets used here

PR #1091, opened 2023-12-08
chore: refactor RundownView to avoid MeteorReactComponent SOFIE-2751

Julusian:
## About the Contributor

## Type of Contribution
This is a bug fix

## New Behavior

## Status

### DIFF
--- a/meteor/client/ui/RundownView.tsx
+++ b/meteor/client/ui/RundownView.tsx
@@ -3,7 +3,13 @@ import * as React from 'react'
 import { parse as queryStringParse } from 'query-string'
 // @ts-expect-error No types available
 import * as VelocityReact from 'velocity-react'
-import { Translated, translateWithTracker } from '../lib/ReactMeteorData/react-meteor-data'
+import {
+	Translated,
+	translateWithTracker,
+	useSubscription,
+	useSubscriptions,
+	useTracker,
+} from '../lib/ReactMeteorData/react-meteor-data'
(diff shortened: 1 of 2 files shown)


meteor/client/ui/RundownView.tsx
```diff
@@ -1205,7 +1209,128 @@ interface ITrackedProps {
 	currentSegmentPartIds: PartId[]
 	nextSegmentPartIds: PartId[]
 }
-export const RundownView = translateWithTracker<IProps, IState, ITrackedProps>((props: Translated<IProps>) => {
+export function RundownView(props: Readonly<IProps>): JSX.Element {
+	const playlistId = props.playlistId
+
```
nytamin, line 1214: I wonder if it might be better to group the `useSubscription` together and instead use a few `useSubscriptions`?

When initially loading, a few of these subscriptions will be invalid anyway (like when `playlistStudioId` is undefined). By instead using `useSubscriptions`, no initial subscriptions will be set up, which might give a slight performance boost on initial load?

nytamin (approved):

meteor/client/ui/RundownView.tsx
Julusian: I've added `useSubscriptionIfEnabled` and made use of that here.
There should be a follow up PR to consider moving all usages of `useSubscription` and some uses of `useSubscriptions` across to use this.

Thinking about it, are all these usages like this correct? it seems like it will get stuck if there are no rundowns in the playlist, with the component never loading?

nytamin (approved):

PR #1123, opened 2024-01-15
feat: allow blueprints to specify a custom resolution for NORA hover preview SOFIE-2840

Julusian:
## About the Contributor

## Type of Contribution

This is a:  Feature

## Current Behavior
The NORA hover preview is currently fixed at being run at a resolution of 1920x1080.

## New Behavior
The blueprints are able to specify a different resolution for the hover preview window, so that it can more accurately represent the content as it will be shown on video walls with other aspect ratios

## Testing Instructions
This requires custom blueprints which set the properties

## Status

### DIFF
--- a/meteor/client/ui/FloatingInspectors/NoraFloatingInspector.tsx
+++ b/meteor/client/ui/FloatingInspectors/NoraFloatingInspector.tsx
@@ -181,9 +181,20 @@ export class NoraPreviewRenderer extends React.Component<{}, IStateHeader> {
 		this.rootElement = e
 	}
 
-	private getElStyle() {
+	private getElStyle(width: number, height: number) {
 		const style = { ...this.state.style }
 		style.visibility = this.state.show ? 'visible' : 'hidden'
+		style.width = `${width / 4}px`
+		style.height = `${height / 4}px`
+		return style
+	}
+
+	private getPreviewStyle(width: number, height: number) {
+		const style: React.CSSProperties = {
+			width: `${width}px`,
+			height: `${height}px`,
+		}
+
 		return style
 	}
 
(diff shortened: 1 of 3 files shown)


packages/blueprints-integration/src/content.ts
```diff
@@ -128,6 +128,7 @@ export interface NoraPayload {
 export interface NoraContent extends BaseContent {
 	payload: NoraPayload
 	previewRenderer: string
+	previewRendererDimensions?: { width: number; height: number }
```
nytamin, line 131:
```suggestion
	/** Dimensions of the rendered hover-preview viewport in pixels. Defaults to 1920x1080. */
	previewRendererDimensions?: { width: number; height: number }
```

meteor/client/ui/FloatingInspectors/NoraFloatingInspector.tsx
```diff
@@ -152,32 +152,50 @@ export class NoraPreviewRenderer extends React.Component<{}, IStateHeader> {
 		this.rootElement = e
 	}
 
-	private getElStyle() {
+	private getElStyle(width: number, height: number) {
 		const style = { ...this.state.style }
 		style.visibility = this.state.show ? 'visible' : 'hidden'
+		style.width = `${width / 4}px`
+		style.height = `${height / 4}px`
```
nytamin, line 159: I think we should make the viewed size be independent of the absolute rendered values.
So something like this instead?

```typescript
const ratio = height / width
const idealWidth = Math.max(300, browserWidth/4)
style.width = `${idealWidth}px`
style.height = `${idealWidth * ratio}px`

```

nytamin (changes requested):

meteor/client/ui/FloatingInspectors/NoraFloatingInspector.tsx
Julusian: I am open to this, but I think it needs a team/UX discussion to figure out what rule it should follow.

Julusian: After discussion with the team, we agreed on making the preview be a maximum of 480px in either dimension

PR #1131, opened 2024-01-25
feat(live-status-gw): expose pieces and split adlibs to a separate topic

ianshade:
## About the Contributor

## Type of Contribution

This is a: 

Feature

## Current Behavior

- The `activePlaylist` topic in the LSG contains a (usually long) array of adLibs that are changing relatively rarely compared to other information in that topic like the current/next parts.
- Information about pieces is not available through the LSG.
- Events from the `activePlaylist` topic may be sent more often than necessary, sometimes featuring incomplete data when the collection data hasn't settled yet e.g. during a Take.

## New Behavior

- Adlibs for the active playlist are now available in the `adLibs` topic.
- Currently playing Pieces from the On Air Part are available in the `activePieces` topic:
  - Updated in real time (driven by the properties set by Core based on the pieceInstance callbacks from the Playout Gateway).
- Throttling and incomplete data prevention are implemented.
- `publicData` property on the AdLibs, Pieces, Parts and Segments, and the Playlist is exposed

## Other Information

This is a breaking change in the Live Status Gateway API and wasn't preceded by an RFC so I'll appreciate any feedback on whether this is a step in the right direction for the users of the LSG

## Status


jstarpl:
Hello!

Thank you for contributing to the Sofie Project!

If you haven’t already, please give our [contribution guidelines](https://nrkno.github.io/sofie-core/docs/for-developers/contribution-guidelines/) a read. We will reach out to you to schedule an initial discussion regarding this (more details to follow).

jstarpl:
Hi!

The NRK team has no objections to this change _per se_ - moving the AdLibs onto a separate topic seems a sensible decision. One could argue that having the _Active Pieces_ exposed on the `activePlaylist` topic might also be too much information, and maybe could instead be branched off to their own `activePlaylistPieces` topic, allowing the consumer a more fine-grained control over the data being pushed. That being said, NRK is not using this feature set at the moment, so we don't feel very strongly about what will make sense.

The more important case is that this is a breaking change, targeting 1.51 in a feature that lands in 1.50, breaking any software that might integrate against Live Status Gateway. Because of that, we would like to invite others using Live Status Gateway - especially the original contributor - to bring forward their positions on this change.

We'll keep this PR open until 2024/02/06.

jstarpl:
Actually, I do have one proper comment: since this would be a breaking change, I think this version number should be major-bumped as well:

https://github.com/nrkno/sofie-core/pull/1131/files#diff-29807ab8c853f30bb20871ec7181b4002c60a75f9fc186d0c9690fa54b495a42R5

scriptorian:
This all looks like good progress. There are many other changes before this set that my customer is not up-to-date with - I have reviewed these as well and have no concerns.
I would prefer to see the active pieces in their own topic as well - it seems more logical to me - but I imagine there are few cases where there are a large number of pieces per part so this is not a real concern.
Yes please to a major version bump. We could consider adding a version string to the connect path if we need to support multiple versions in the future.

jstarpl:
@ianshade How would you feel about splitting off the active pieces to their own topic. Would that be okay with you?

ianshade:
@jstarpl @scriptorian I like this idea. I'm all for splitting the active pieces to their own topic.

PeterC89:
This looks great, we're about to start working on some bits that will integrate with live status gateway and this brings some much needed functionality on our end!
I can't quite tell if this will include the newly implemented `publicData `property on the parts / pieces? Is that part of this PR or a future concern?

ianshade:
@PeterC89, `publicData` is not included, but if there is no objection, I can add it as part of this PR

jstarpl:
> @PeterC89, `publicData` is not included, but if there is no objection, I can add it as part of this PR

I'm gonna go out on a limb, and say that I don't expect anyone from the NRK team to object.

ianshade:
I extracted active pieces to a separate topic, exposed `publicData` where applicable, and bumped the version.

jstarpl:
@ianshade I've updaed the "New behavior" section of the PR description, please make sure I didn't mischaracterize something.

ianshade:
> I've updated the "New behavior" section of the PR description, please make sure I didn't mischaracterize something

@jstarpl, thanks, I added that the Playlist is among the documents that have their `publicData` exposed.

The openapi docs now reflect the `publicData` addition.

jstarpl (approved):

PR #1135, opened 2024-01-29
feat: add `node-fetch` as a global to blueprint execution SOFIE-1367

Julusian:
## About the Contributor

## Type of Contribution

This is a: Feature

## New Behavior

This exposes a `fetch` global into blueprints, using `node-fetch` as the implementation.

While we are using `got` elsewhere, it feels bad to introduce `got` v11 into new usages, given that the current release of got is v14. It does appear to be getting security patches, but for how long?
By using `fetch`, it should provide a more seamless experience once we update to node18, as the builtin `fetch` implementation will be usable instead. `node-fetch` also looks to be better maintained, even though this is using the previous major version.

## Other Information

This is not necessary for R50, but it makes testing with storybuilder much easier as it allows for developing without needing package-manager to be setup and running.

## Status

### DIFF
--- a/meteor/yarn.lock
+++ b/meteor/yarn.lock
@@ -2174,6 +2174,7 @@ __metadata:
     elastic-apm-node: ^3.43.0
     eventemitter3: ^4.0.7
     mongodb: ^4.13.0
+    node-fetch: ^2.7.0
     p-lazy: ^3.1.0
     p-timeout: ^4.1.0
     superfly-timeline: ^8.3.1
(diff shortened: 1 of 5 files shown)


PeterC89:
@Julusian Does this mean `fetch` would become available for the Blueprint ingest methods etc.?

Julusian:
> Does this mean `fetch` would become available for the Blueprint ingest methods etc.?

@PeterC89 yes, I intending to use this during ingest operations. I don't think it is a good idea to do it, instead doing the fetches via package-manager should be preferred), but for debugging/prototyping doing the fetch directly is much simpler.

Julusian:
I've just found a task in our Jira which says:
> Extends ICommonContext to include a WebAPI-like (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) fetch method, with a built-in, reasonable timeout.

I am wondering if that would be a better idea (mostly to force the default timeout)

jstarpl:
I think probably having the `fetch` on the `global` makes more sense, because that's much more generally-compatible (rather than a method on the context object), but I suppose the default timeout is probably a good idea, so that we don't get hang up during ingest/execution forever.

jstarpl (approved):
This looks fine, but I think it would make more sense, and be more "standard" if `fetch` was a part of the global object, rather than exposed on the context object - but still with some sort of a default timeout/abort controller.

Issue #1125, opened 2024-01-16
RFC: Improve GUI Accessibility 


hummelstrand:
### About me

This RFC is posted by _SuperFly.tv_ on behalf of the _BBC_.

### Use case

Improving the accessibility of the Sofie GUI in order to comply with regulatory requirements. Recommendations for changes  are outlined in the attached report.
[Public Summary of Accessibility Audit report Sofie.pdf](https://github.com/nrkno/sofie-core/files/13953044/Public.Summary.of.Accessibility.Audit.report.Sofie.pdf)

### Proposal

We propose to modify the GUI in order to better cater to accessibility tools. We intend to target _Release 51_.

### Process

The Sofie Team will evaluate this RFC and open up a discussion about it, usually within a week.
- [x] RFC created
- [x] Sofie Team has evaluated the RFC
- [x] A workshop has been planned
- [x] RFC has been discussed in a workshop
- [x] A conclusion has been reached, see comments in thread

jstarpl:
Thank you for submitting this RFC!
(If you haven’t already, please give our [contribution guidelines](https://nrkno.github.io/sofie-core/docs/for-developers/contribution-guidelines/) a read.)

The Sofie NRK Team welcomes this improvement and sees no point in setting up a workshop. We'll keep this RFC open for comments until a PR is opened.

hummelstrand:
### Question to other Sofie users: 
If, as part of the PR, we changed the HTML page title from the currently static "Sofie" to something like "Sofie: Name of Playlist", would that break any existing scripts or logging workflows for any users?

hummelstrand:
PR: https://github.com/nrkno/sofie-core/pull/1138

PR #1138, opened 2024-02-02
feat: a11y improvements

jstarpl:
## About the Contributor

## Type of Contribution

This is a: 

Feature / Other: enhancements and refactoring for better accessibility.

## Current Behavior

There is a number of accessibility issues, as presented in the original RFC: https://github.com/nrkno/sofie-core/issues/1125

## New Behavior

Some of the issues were addressed. Some major changes were needed to the `ModalDialog` system in order to allow both _Escape/Enter_ key dismissal/confirmation of them and at the same time support multi-choice ModalDialogs. Also, _Tab_ key is no longer "disabled" in code, but instead will only be disabled if a _Tab_ Action Trigger is assigned.

## Testing Instructions

* Check if the After Broadcast Form (Evaluations) still works as expected
* Check if the _Tab_ key for opening/closing shelf still works as previously
* Check if there are any regressions around "Enter" key function, as some elements that used to be `<div>` are now `<buttons>` and will behave as if clicked if they have focus and "Enter" is pressed.

## Status

### DIFF
--- a/meteor/client/lib/ModalDialog.tsx
+++ b/meteor/client/lib/ModalDialog.tsx
@@ -1,17 +1,17 @@
-import * as React from 'react'
+import React, { useLayoutEffect, useRef } from 'react'
 import CoreIcons from '@nrk/core-icons/jsx'
 import Escape from './Escape'
+import FocusBounder from 'react-focus-bounder'
+import { useTranslation } from 'react-i18next'
 
 import ClassNames from 'classnames'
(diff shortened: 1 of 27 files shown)


nytamin (approved):
Looks good!

A question for later:
Should we enable some linting rule that helps us with following recommended accessability practices?
something like https://github.com/jsx-eslint/eslint-plugin-jsx-a11y ?

PR #1142, opened 2024-02-06
fix: restore ALL data from snapshot, for debugging

nytamin:
## About the Contributor

## Type of Contribution

This is a: Feature

## Current Behavior

When restoring a rundownPlaylist snapshot, only mediaObjects are restored, no other package-related data.

This is a problem when troubleshooting issues with packages, such as zebra stripes in the GUI.

## New Behavior

Added a parameter to the restoreSnapshot method to allow the user to choose whether to only restore the rundownPlaylist, or also the studio-related data from the snapshot:

image

When uploading for debugging, the following additional data is restored:

* mediaObjects
* expectedPackageWorkStatuses
* packageContainerPackageStatuses
* packageInfos

When uploading normally, no additional data is restored.

## Testing Instructions

* Just test that restoring of snapshots still work

## Status

### DIFF
--- a/meteor/client/ui/Settings/SnapshotsView.tsx
+++ b/meteor/client/ui/Settings/SnapshotsView.tsx
@@ -29,7 +29,8 @@ interface IProps {
 	}
 }
 interface IState {
-	uploadFileKey: number // Used to force clear the input after use
+	uploadFileKey: string // Used to force clear the input after use
+	uploadFileKey2: string // Used to force clear the input after use
 	editSnapshotId: SnapshotId | null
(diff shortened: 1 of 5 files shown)


ianshade:
I would like to suggest one more thing that the option of restoring a Rundown Playlist snapshot for debugging could do, which is not deleting Playlist's `activationId`. It would be perfect for investigating issues where something went visibly wrong in the UI (or on the output) and the snapshot was taken while experiencing the issue (and while the Playlist was still active). It should probably be optional, and the UI should clearly state the consequences that restoring an active Playlist might have.

PR #1146, opened 2024-02-13
feat: support external url dropzones

mint-dewit:
## About the Contributor

## Type of Contribution

This is a  Feature

## New Behavior

As of this PR is is possible for plugins to veer off the MOS plugin spec and provide a special drop page to be used for drag and drop operations. Sofie will listen for messages from both the external panel and the dropzone page. This mechanism allows us to work around limitations for drag and drop under cross origin policies.

## Other Information

We intend to add public documentation for this mechanism as part of another task.

### DIFF
--- a/meteor/client/ui/Shelf/BucketPanel.tsx
+++ b/meteor/client/ui/Shelf/BucketPanel.tsx
@@ -73,7 +73,6 @@ import {
 	ShowStyleVariantId,
 } from '@sofie-automation/corelib/dist/dataModel/Ids'
 import { RundownPlaylistCollectionUtil } from '../../../lib/collections/rundownPlaylistUtil'
-import { TFunction } from 'react-i18next'
 
 interface IBucketPanelDragObject {
 	id: BucketId
@@ -424,7 +423,7 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
 					window.addEventListener(MOSEvents.dragleave, this.onDragLeave)
 
 					RundownViewEventBus.on(RundownViewEvents.REVEAL_IN_SHELF, this.onRevealInShelf)
-					RundownViewEventBus.on(RundownViewEvents.TOGGLE_SHELF_DROPZONE, this.onToggleDropFrame.bind(this))
+					RundownViewEventBus.on(RundownViewEvents.TOGGLE_SHELF_DROPZONE, this.onToggleDropFrame)
 				}
 
 				componentDidUpdate(prevProps: IBucketPanelProps & IBucketPanelTrackedProps) {
@@ -800,7 +799,7 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
 
 				private onToggleDropFrame = (e: ToggleShelfDropzoneEvent) => {
 					this.setState({
-						dropActive: e.display,
+						// dropActive: e.display,
 						dropFrameActive: e.display ? e.id : null,
 					})
 				}
@@ -844,7 +843,8 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
 								<div
 									className={ClassNames('dashboard-panel', 'dashboard-panel__panel--bucket', {
 										'dashboard-panel__panel--bucket-active': this.state.dropActive,
-										'dashboard-panel__panel--sort-dragging': this.props.isDragging,
+										'dashboard-panel__panel--sort-dragging':
+											(this.props.isDragging || this.state.dropFrameActive) && !this.state.dropActive,
 									})}
 									data-bucket-id={this.props.bucket._id}
 									ref={this.setRef}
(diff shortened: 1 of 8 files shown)


meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -419,6 +424,7 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
 					window.addEventListener(MOSEvents.dragleave, this.onDragLeave)
 
 					RundownViewEventBus.on(RundownViewEvents.REVEAL_IN_SHELF, this.onRevealInShelf)
+					RundownViewEventBus.on(RundownViewEvents.TOGGLE_SHELF_DROPZONE, this.onToggleDropFrame.bind(this))
```
jstarpl, line 427:
```suggestion
					RundownViewEventBus.on(RundownViewEvents.TOGGLE_SHELF_DROPZONE, this.onToggleDropFrame)
```

Since this is using the = lambda way of setting up member functions, I don't think the `.bind` does anything?

meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -919,3 +944,81 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
...
+	}
+	const registerHandlers = () => {
+		window.addEventListener('message', onMessage)
+	}
+	const unregisterHandlers = () => {
+		window.removeEventListener('message', onMessage)
+	}
+
+	const onMessage = (event: MessageEvent) => {
+		if (event.source === frame?.contentWindow) {
```
jstarpl, line 978:
```suggestion
		if (event.source !== frame?.contentWindow) return
```

Maybe this could be inverted for less indentation?

meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -919,3 +944,81 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
...
+			}
+		}
+	}
+
+	React.useEffect(
+		() => () => {
+			unregisterHandlers()
+		},
+		[]
+	)
```
jstarpl, line 1012: I would suggest re-arranging this to a pattern where the register- and unregisterHandlers are joined in a single hook, something like: 

```suggestion
	const [dropzoneElementRef, setDropzoneElementRef] = React.useState<HTMLElement>(null)
	const mounted = !!dropzoneElementRef
	
	React.useEffect(
		() => {
			if (!dropzoneElementRef) return
			
			const registerHandlers = () => {
				window.addEventListener('message', onMessage)
			}
			const unregisterHandlers = () => {
				window.removeEventListener('message', onMessage)
			}
			
			registerHandlers()
			
			return () => {
				unregisterHandlers()
			}
		},
		[dropzoneElementRef]
	)
```

meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -919,3 +944,81 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
...
+	const setDropzoneElement = (frameRef: HTMLIFrameElement | null) => {
+		frame = frameRef
+		if (frame && !mounted) {
+			registerHandlers()
+			mounted = true
+		} else {
+			unregisterHandlers()
+			mounted = false
+		}
+	}
```
jstarpl, line 969: I think component state in React should only really be held and mutated via the `React.useState` hook. I feel like this way of doing it can very quickly end up in some undefined situations, if there are spontaneous component re-renders. The `mounted` will be false then, but the `setDropzoneElement` will not re-run, because as far as react will be concerned, the reference has already been set.

See a note below on how to solve this.

meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -919,3 +944,81 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
...
+	const registerHandlers = () => {
+		window.addEventListener('message', onMessage)
+	}
+	const unregisterHandlers = () => {
+		window.removeEventListener('message', onMessage)
+	}
+
+	const onMessage = (event: MessageEvent) => {
```
jstarpl, line 977:
```suggestion
	const onMessage = React.useCallback((event: MessageEvent) => {
```

I suggest wrapping this in a callback for reference stability.

meteor/client/ui/Shelf/BucketPanel.tsx
```diff
@@ -919,3 +944,81 @@ export const BucketPanel = translateWithTracker<Translated<IBucketPanelProps>, I
 		)
 	)
 )
+
+interface DropzoneHolderProps {
+	id: string
+	bucketId: BucketId
+	url: string
+	hidden: boolean
+	t: TFunction<'translation', undefined>
```
jstarpl, line 953: There's really no reason to pass the t function via props, just use

```
const {t} = useTranslation()
```

jstarpl (changes requested):

PR #1165, opened 2024-03-08
fix: add a way to take a heap snapshot (patch for release49)

nytamin:
## About the Contributor

This pull request is posted on behalf of the NRK.

## Type of Contribution

This is a hack, to allow for memory leak troubleshooting.

## New Behavior

This PR adds a GET method that returns a NodeJs heap snapshot, which can be inspected in the Chrome devtools.

## Testing Instructions

1. Go to Settings / Core System Settings
2. At the bottom of the page, you find a new button under **Memory troublshooting**

## Time Frame

This is urgent for us, please review and merge it as soon as possible.

## Status

- [x] PR is ready to be reviewed.
- [x] The functionality has been tested by the author.
- [ ] Relevant unit tests has been added / updated.
- [ ] Relevant documentation (code comments, [system documentation](https://nrkno.github.io/sofie-core/)) has been added / updated.

PR #1166, opened 2024-03-08
fix: add a way to take a heap snapshot.

nytamin:
This is a refactor of https://github.com/nrkno/sofie-core/pull/1165, targeting release51

### DIFF
--- a/meteor/client/ui/Settings/SystemManagement.tsx
+++ b/meteor/client/ui/Settings/SystemManagement.tsx
@@ -526,14 +526,14 @@ export function checkForOldDataAndCleanUp(t: TFunction, retriesLeft = 0): void {
 function SystemManagementHeapSnapshot() {
 	const { t } = useTranslation()
 
-	const [displayWarning, setdisplayWarning] = React.useState(false)
+	const [displayWarning, setDisplayWarning] = React.useState(false)
 	const [active, setActive] = React.useState(false)
 
 	const onAreYouSure = useCallback(() => {
-		setdisplayWarning(true)
+		setDisplayWarning(true)
 	}, [])
 	const onReset = useCallback(() => {
-		setdisplayWarning(false)
+		setDisplayWarning(false)
 		setActive(false)
 	}, [])
 	const onConfirm = useCallback(() => {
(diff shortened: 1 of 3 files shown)


meteor/server/api/heapSnapshot.ts
```diff
@@ -0,0 +1,69 @@
...
+	// Retrieve heap snapshot:
+	heapSnapshotPrivateApiRouter.get('/retrieve', async (ctx) => {
+		return handleKoaResponse(ctx, async () => {
+			return retrieveHeapSnapshot({ userId: null })
+		})
+	})
+}
+
+// Retrieve heap snapshot:
+heapSnapshotPrivateApiRouter.get('/:token/retrieve', async (ctx) => {
```
Julusian, line 65: I'm not sure that this route should exist.

In an environment where security has been enabled, I think this provides risk of being an accidental backdoor and I'm not 100% sure how it fits into the auth flow we have defined.

So until we address the user accounts setup, I think it would be better to omit this entirely. But I don't feel that strongly, so feel free to ignore this :)

meteor/client/ui/Settings/SystemManagement.tsx
```diff
@@ -522,3 +523,51 @@ export function checkForOldDataAndCleanUp(t: TFunction, retriesLeft = 0): void {
 		})
 		.catch(catchError('system.cleanupOldData'))
 }
+function SystemManagementHeapSnapshot() {
+	const { t } = useTranslation()
+
+	const [displayWarning, setdisplayWarning] = React.useState(false)
```
Julusian, line 529:
```suggestion
	const [displayWarning, setDisplayWarning] = React.useState(false)
```

Julusian (approved):

meteor/server/api/heapSnapshot.ts
nytamin: This is analog to [how the snapshots are retrieved](https://github.com/nrkno/sofie-core/blob/fix/take-heap-snapshot/meteor/server/api/snapshot.ts#L771), so I think that it should be fine to add it like this for now and they should both be addressed  when security is eventually looked into.

PR #1173, opened 2024-04-12
feat: New Prompter scroll algorithm (SOFIE-3082)

nytamin:
## About the Contributor

## Type of Contribution

This is a: Bug fix

## Current Behavior

We have reported issues in prod where the prompter "jump" in case of content changes (ie doesn't scroll to the new position correctly).
We also see quite a few `Read anchor could not be found after update` errors in our logs.

I haven't been able to properly reproduce the issue, but I believe it is related to the `getScrollAnchor` method returning an anchor which doesn't exist after the update.

## New Behavior

I've rewritten the `getScrollAnchor` algorithm to no return a list of anchors instead of a single one, order in falling "order of importance".

When the content changes, if `restoreScrollAnchor` fails to find the first anchor, it can now continue and try the next ones after that.

## Testing Instructions

* Use the Prompter, scroll to various positions.
* Modify the  rundown content (edit text, add / remove Parts & segments).
* Observe how the Prompter behaves, it should never jump to a strange/unexpected position after an update.

## Status

### DIFF
--- a/meteor/client/ui/Prompter/PrompterView.tsx
+++ b/meteor/client/ui/Prompter/PrompterView.tsx
@@ -1,5 +1,5 @@
 import React, { PropsWithChildren } from 'react'
-import _ from 'underscore'
+import _, { filter } from 'underscore'
 import Velocity from 'velocity-animate'
 import ClassNames from 'classnames'
 import { Meteor } from 'meteor/meteor'
@@ -574,7 +574,11 @@ interface IPrompterTrackedProps {
 	prompterData: PrompterData | null
 }
 
-type ScrollAnchor = [number, string] | null
+interface ScrollAnchor {
+	offset: number
+	anchorId: string
+}
+type PrompterSnapshot = ScrollAnchor[] | null
 
 export const Prompter = translateWithTracker<PropsWithChildren<IPrompterProps>, {}, IPrompterTrackedProps>(
 	(props: IPrompterProps) => ({
@@ -639,7 +643,7 @@ export const Prompter = translateWithTracker<PropsWithChildren<IPrompterProps>, 
 			})
 		}
 
-		getScrollAnchor = () => {
+		getScrollAnchors = (): ScrollAnchor[] => {
 			let readPosition = ((this.props.config.margin || 0) / 100) * window.innerHeight
 			switch (this.props.config.marker) {
 				case 'top':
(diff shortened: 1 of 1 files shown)


meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -653,46 +657,134 @@ export const Prompter = translateWithTracker<PropsWithChildren<IPrompterProps>,
...
+			{
+				const scrollAnchors = Array.from(document.querySelectorAll('.prompter .scroll-anchor'))
+					.map((anchor) => {
+						const { top, bottom } = anchor.getBoundingClientRect()
+						return { id: anchor.id, top, bottom }
+					})
+					.sort((a, b) => a.top - b.top)
+				for (const anchor of scrollAnchors) {
+					// find a prompter scroll anchor that is just above the prompter read position:
+					const distanceToReadPosition = Math.abs(anchor.top - readPosition)
```
jstarpl, line 713: I think this is too generic. The fact is we don't want to just find the closest match, we want to maintain whatever the reader is most likely focusing on, which may in fact not be the one with it's `top` being the closest. Consider a situation where one is reading the very end of a very long paragraph in Part A, and seeing Part B in view port. Part A's `anchor.top` will in fact be further away than Part B's `anchor.top`, But we want to keep Part A static, and Part B can just move around, if for example the editor moves a new Part C in front of Part B.

meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -653,46 +657,134 @@ export const Prompter = translateWithTracker<PropsWithChildren<IPrompterProps>,
...
+
+						for (const piece of part.pieces) {
+							if (scrollAnchor.anchorId === `line_${piece.id}`) {
+								logger.error(`Read anchor troubleshooting: piece "${piece.id}" was found in prompterData!`)
+								return
+							}
+						}
+					}
+				}
 			}
```
jstarpl, line 788: I don't see how any of this can happen, considering React component life cycle, but I've said that already.

jstarpl (changes requested):

meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -652,47 +657,147 @@ export const Prompter = translateWithTracker<PropsWithChildren<IPrompterProps>,
...
-			if (scrollAnchor === null) return
-			const anchor = document.getElementById(scrollAnchor[1])
-			if (anchor) {
-				const { top } = anchor.getBoundingClientRect()
+		private restoreScrollAnchor = (scrollAnchors: ScrollAnchor[] | null) => {
+			if (scrollAnchors === null) return
+			if (!scrollAnchors.length) {
+				if (this.props.prompterData?.segments.length) {
+					// That's wierd, log the issue:
+					logger.error(`No read anchors to use after update`)
```
jstarpl, line 736: I'm not sure if this is that weird - all it takes is a rundown activation and not getting all segments+parts yet from the publication. `scrollAnchors` will be `null` only on mount, at every update later, it will be an array, and if the data isn't loaded yet, it will be empty, so I think this is to be expected. Anyway, this doesn't provide any context (playlistId, etc.), so maybe it should?

meteor/client/ui/Prompter/PrompterView.tsx
```diff
@@ -652,47 +657,147 @@ export const Prompter = translateWithTracker<PropsWithChildren<IPrompterProps>,
...
+						}
+
+						for (const piece of part.pieces) {
+							if (scrollAnchor.anchorId === `line_${piece.id}`) {
+								logger.error(`Read anchor troubleshooting: piece "${piece.id}" was found in prompterData!`)
+								return
+							}
+						}
+					}
+				}
```
jstarpl, line 800: Maybe this could be taken out to a separate method, it doesn't really have anything to do with `restoreScrollAnchor` ?

jstarpl (approved):
Some low-priority comments only.

PR #1176, opened 2024-04-18
Fix: various cases where only one document would be updated (SOFIE-3100)

nytamin:
## About the Contributor

## Type of Contribution

This is a: Bug fix

## Current Behavior

Because the option `{ multi: true }` was missing in some places when calling `mongoCollection.update() .updateAsync() .upsert() .upsertAsync()`, instead of updating all objects, only the first matching document is updated.

## New Behavior

New types disallow calling an update method which might affect multiple documents without specifying the `{ multi: boolean }` option.
All objects are now updated properly.

### Affected areas

* Nightly cleanup of orphaned `PackageInfos` only marked a single document for later removal
* (unverified) When a PeripheralDevice goes offline, subdevices might not get their status set to offline
* (unverified) When a ShowStylebase changes its blueprint or config preset, not all variant got their blueprintConfigPresetIdUnlinked set
* (unverified) Migrations might not update all data correctly (so they might have to be run multiple times)

## Time Frame

I intend to merge this asap

## Status

…mullti: true } is required when calling update methods that can update multiple documents

### DIFF
--- a/meteor/lib/collections/lib.ts
+++ b/meteor/lib/collections/lib.ts
@@ -180,7 +180,7 @@ export class WrappedMongoCollection<DBInterface extends { _id: ProtectedString<a
 		}
 	}
 	update(
-		selector: MongoQuery<DBInterface> | DBInterface['_id'],
+		selector: MongoQuery<DBInterface> | DBInterface['_id'] | { _id: DBInterface['_id'] },
 		modifier: MongoModifier<DBInterface>,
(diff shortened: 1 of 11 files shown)


Julusian (approved):
Maybe long term we should consider following what mongodb has done, and splitting this method into `updateOne` and `updateMany`, to make the behaviour explicit and easy

jstarpl (approved):

PR #1182, opened 2024-05-01
Chore/openapi ci

mint-dewit:
## Type of Contribution

This is a:
CI Improvement

## New Behavior
An OpenAPI client library will be published to NPM in tandem with our other libraries.


nytamin (approved):

Julusian (approved):
I don't mind this using an action, but it might be better to do the needed yarn calls instead, seeing as it is fairly simple to do? 
https://github.com/nrkno/sofie-core/blob/33ed635534164e6d2c62c4722850f4946626cfa4/.github/workflows/node.yaml#L553-L580

Perhaps that ci step should be backported to r50 too? I don't mind though, we shouldn't be making any notable changes to the api at this stage.

PR #1209, opened 2024-06-19
Fix: live status gateway send message performance

nytamin:
## About the Contributor

## Type of Contribution

This is a:

Performance improvement

## Current Behavior

When testing locally, the size of an ActivePlaylist message is about 170kB. This takes around 1ms to stringify, so if LSG has several/many recipients, those numbers quickly add up.

## New Behavior

when sending a message, the message is only stringified once, then sent to the recipients.

## Testing

## Status

### DIFF
--- a/packages/live-status-gateway/src/wsHandler.ts
+++ b/packages/live-status-gateway/src/wsHandler.ts
@@ -34,15 +34,22 @@ export abstract class WebSocketTopicBase {
 		this._logger.error(`Process ${this._name} message not expected '${JSON.stringify(msg)}'`)
 	}
 
-	sendMessage(ws: WebSocket, msg: object): void {
+	sendMessage(recipients: WebSocket | Iterable<WebSocket>, msg: object): void {
 		const msgStr = JSON.stringify(msg)
-		this._logger.debug(`Send ${this._name} message '${msgStr}'`)
-		ws.send(msgStr)
+
+		recipients = isIterable(recipients) ? recipients : [recipients]
+
+		let count = 0
+		for (const ws of recipients) {
+			count++
+			ws.send(msgStr)
+		}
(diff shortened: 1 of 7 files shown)


packages/live-status-gateway/src/wsHandler.ts
```diff
@@ -34,15 +34,22 @@ export abstract class WebSocketTopicBase {
 		this._logger.error(`Process ${this._name} message not expected '${JSON.stringify(msg)}'`)
 	}
 
-	sendMessage(ws: WebSocket, msg: object): void {
+	sendMessage(recipients: WebSocket | Iterable<WebSocket>, msg: object): void {
 		const msgStr = JSON.stringify(msg)
```
Julusian, line 38: Perhaps this could be optimised further, by only doing this JSON.stringify if there are any recipients.

The new implementation of this means that if there are none, it will be stringifying when it wouldnt previously

Julusian (approved):

packages/live-status-gateway/src/wsHandler.ts
```diff
@@ -34,15 +34,22 @@ export abstract class WebSocketTopicBase {
...
+			count++
+			ws.send(msgStr)
+		}
+		this._logger.silly(`Send ${this._name} message '${msgStr}' to ${count} recipients`)
 	}
 
 	sendHeartbeat(ws: WebSocket): void {
 		const msgStr = JSON.stringify({ event: 'heartbeat' })
-		this._logger.silly(`Send ${this._name} message '${msgStr}'`)
+		// this._logger.silly(`Send ${this._name} message '${msgStr}'`)
```
jstarpl, line 52: Should this be removed? Or just kept, but easier to filter out?

nytamin: I think we should remove it, unless we have any specific issue that we need to be able to troubleshoot in prod.

PR #1210, opened 2024-06-21
feat: Rundown `source` property SOFIE-2963

Julusian:
## About the Contributor
This pull request is posted on behalf of the NRK.

## Type of Contribution

This is a: Code improvement 

## New Behavior
This changes the `Rundown` document, replacing the `peripheralDeviceId`, `restoredFromSnapshotId` and `externalNRCSName` properties with a new combined `source` property.
This clarifies the relationship and valid states of these properties, and allows for defining more states

## Testing

- [ ] I have added one or more unit tests for this PR
- [x] I have updated the relevant unit tests
- [ ] No unit test changes are needed for this PR

## Status

- [x] PR is ready to be reviewed.
- [x] The functionality has been tested by the author.
- [x] Relevant unit tests has been added / updated.
- [ ] Relevant documentation (code comments, [system documentation](https://nrkno.github.io/sofie-core/)) has been added / updated.

PR #1211, opened 2024-06-21
feat: create adlib testing rundowns SOFIE-2963

Julusian:
## About the Contributor

## Type of Contribution

This is a: Feature 

## New Behavior

This adds support for having buttons on the rundown list page, to create empty rundowns to be used with the 'adlib testing' mode.
The buttons to add the rundowns are only shown to studio level users, and they can also see the delete button to remove them

A blueprint method has been added `generateTestingRundown` which allows the blueprints to define the `IngestRundown` that should be used, if not implemented then sofie will fallback to using a minimal empty rundown.

Support for generating these rundowns has to be enabled on each show style variant that wants it, making this be opt-in and not visible for installations where not explicitly enabled.

UX isn't amazing, this can be refined with further iterations
image

This builds upon #1210, which allows for the 'reload nrcs data' to re-run the create operation to regenerate the rundown including blueprint defined ingest data.

## Testing

## Status


meteor/client/ui/RundownList/CreateTestingRundownPanel.tsx
```diff
@@ -0,0 +1,60 @@
...
+				})
+		}
+
+		const interval = setInterval(doPoll, 30000)
+		doPoll()
+
+		return () => {
+			isDisposed = true
+			clearInterval(interval)
+		}
```
nytamin, line 33: Not a fan of this polling.
Can't we expose this as a publication instead? Or possibly just calculate it client side, since it just uses the studio, showStyleBase and showStyleVariant?

meteor/client/ui/RundownList.tsx
```diff
@@ -185,6 +186,8 @@ export function RundownList(): JSX.Element {
 				)}
 			</section>
 
+			{getAllowStudio() && <CreateTestingRundownPanel />}
```
nytamin, line 189: About naming of this feature:
An idea is to name this "Adlib Rundowns", since that would align with the idea of "unplanned things, created/played/instanciated by the user"

packages/job-worker/src/ingest/ingestRundownJobs.ts
```diff
@@ -84,13 +79,19 @@ export async function handleUserRemoveRundown(context: JobContext, data: UserRem
 				}
 			}
 		})
+	} else {
+		// Its a real rundown, so defer to the proper route for deletion
```
nytamin, line 83: To avoide future issues, I think it would be better to list the other alternatives in an `else if()`, ending with an `else assertNever(tmpRundown.source.type)`, this way one would be alerted of this code if adding other alternatives to `source.type` in the future.


Also (minor note): I think this comment is slightly confusing, what does "real" mean? (a testing-rundown is "real", isn't it?). Just changing this to "These follow the normal deletion flow" is good enough for me. :)

nytamin (changes requested):

meteor/client/ui/RundownList/CreateTestingRundownPanel.tsx
Julusian: My thinking was that this is not worth the effort/complexity of a publication, as it does not need to be at all reactive it will only ever change when configuring Sofie, and it is cheap to calculate so doesn't benefit from any shared computation.
So it didn't feel worth the effort of setting up and building a full publication
I wouldn't be opposed to dropping the polling and having it loaded only on component mount

> Or possibly just calculate it client side

We currently only load the showstyles for the current rundowns, so the data isn't already there client side.

packages/job-worker/src/ingest/ingestRundownJobs.ts
Julusian: The 'real' here means not sourced from a snapshot, which in other words means it was created by an ingest operation and not the snapshot restore operation. A testing rundown qualifies as 'real' here, as it was generation by Sofie from an IngestRundown, using the exact same logic to do so as an nrcs rundown.

This special case exists because of the comment at the top of this method:
```
	 * As the user requested a delete, it may not be from an ingest gateway, and have a bad relationship between _id and externalId.
	 * Because of this, we must do some more manual steps to ensure it is done correctly, and with as close to correct locking as is reasonable
```

But if you would prefer, that if could probably be changed to `tmpRundown._id !== getRundownId(context.studioId, tmpRundown.externalId)`, as that is really what this is guarding against.

meteor/client/ui/RundownList/CreateTestingRundownPanel.tsx
nytamin: Yeah, perhaps dropping the polling would be good enough

packages/job-worker/src/ingest/ingestRundownJobs.ts
nytamin: >But if you would prefer, that if could probably be changed to tmpRundown._id !== getRundownId(context.studioId, tmpRundown.externalId), as that is really what this is guarding against.

Yeah, that would make it easier to understand from an initial read, I think.

nytamin (approved):

PR #1215, opened 2024-06-25
chore: check CURRENT_SYSTEM_VERSION matches package.json

jstarpl:
## About the Contributor

## Type of Contribution

This is a:

Code improvement

## Current Behavior

CURRENT_SYSTEM_VERSION can be mismatched with package.json, potentially leading to _weird_ issues.

## New Behavior

There is a test checking if the two values match for a real release (not a pre-release version string). This should ensure that we don't publish versions with a mismatch.

## Testing

No testing necessary.

- [x] I have updated the relevant unit tests

## Status

### DIFF
--- a/meteor/server/migration/__tests__/currentSystemVersion.test.ts
+++ b/meteor/server/migration/__tests__/currentSystemVersion.test.ts
@@ -1,0 +1,30 @@
+import fs from 'fs/promises'
+import path from 'path'
+import { CURRENT_SYSTEM_VERSION } from '../currentSystemVersion'
+
+describe('CURRENT_SYSTEM_VERSION', () => {
+	it('Matches the package.json version', async () => {
+		const currentSystemVersion = CURRENT_SYSTEM_VERSION
+
+		const packageJson = JSON.parse(
+			await fs.readFile(path.join(__dirname, '../../../package.json'), {
+				encoding: 'utf8',
+			})
+		)
+
+		const packageVersion = packageJson.version
+
+		if (packageVersion.includes('-')) {
+			console.warn(`Package version is pre-release. Mismatch is fine.`)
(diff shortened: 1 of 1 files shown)


Julusian:
I'm not immediately sold on this test currently. I suspect it will only fail once a tag has been pushed, meaning either replacing the tag, or re-running the portion of the release workflow to bump the version again in order to generate a new tag.

meteor/server/migration/__tests__/currentSystemVersion.test.ts
```diff
@@ -0,0 +1,30 @@
...
+			console.log(`package.json version: ${packageVersion}`)
+			console.log(`currentSystemVersion.ts version: ${currentSystemVersion}`)
+			return
+		}
+
+		if (packageVersion !== currentSystemVersion) {
+			throw new Error(
+				`Package version ("${packageVersion}") does not match CURRENT_SYSTEM_VERSION ("${currentSystemVersion}")!`
+			)
+		}
```
nytamin, line 28: Is there a particular reason to `throw` instead of using the `expect()` assertion?

Do we need to emit the warnings? If not, perhaps we could simply do this instead?

```suggestion
		if (isPrerelease(packageVersion)) return
		expect(packageVersion).toEqual(CURRENT_SYSTEM_VERSION)
```

Julusian:
> Well, since the release procedure for creating the tag in the first place for Core includes running these tests

As far as I'm aware, those tests will be run before the version number is changed, meaning it will always contain the prerelease suffix and will skip the check.
At the point where the version number is updated (using `meteor npm run release`), the tag is also created (locally), and the current instructions don't include running tests after that point, other than 'make sure that the new tag builds correctly in jenkins' at which point the tag has been pushed

PR #1221, opened 2024-07-05
Add LSG support for templateAdLibs

scriptorian:

Feature

LSG offers adLibActions and rundownBaselineAdLibActions in the adLib topic
Each adLibAction optionally includes an editableFields property to offer a schema for the adLibOptions property in the REST API for executeAdLib

A new type of adLibAction has been added to the adLib topic - exposed as an (optional) additional array of templateAdLibs.
The templateAdLibActions are intended for actions where important data required to complete the action is not available during getRundown/getSegment processing, for example to support playing a video clip that is stored on an external asset management system.
To support this change and to allow robust parameter validation the editableFields property has been renamed optionsSchema and this has been tested to support the important difference between ingest created adlibs where for example a video clip is defined and shouldn't be changeable and 'ad-hoc' adlibs where the video clip information _must_ be provided through the executeAdLib adlibOptions properties.

#### Status

- [x] Relevant unit tests have been checked.

#### Affected Areas

* This PR affects the Live Status Gateway.
* TemplateAdLibs and the optionsSchema property will not appear unless blueprints create them.


nytamin:

Julusian (approved):
I want a second opinion on the concept and how it gets enabled in `userDataManifest`, but I am happy with the code change

packages/live-status-gateway/src/topics/adLibsTopic.ts
```diff
@@ -32,6 +32,7 @@ export interface AdLibsStatus {
 	rundownPlaylistId: string | null
 	adLibs: AdLibStatus[]
 	globalAdLibs: GlobalAdLibStatus[]
+	templateAdLibs?: TemplateAdLibStatus[]
```
jstarpl, line 35: Are `templateAdLibs` only global? Or are they always Segment- or Part-local? I think it would be good to make that clear. Or should they still exists on individual `adLibs` and `globalAdLibs` topics and have a `template` or `optionsRequired` flag turned on?

jstarpl (changes requested):
One thing that is not quite clear to me from the spec is that `templateAdLibs` are Global, which I think will be even more confusing because there is no separate place for these "Template" AdLibs in the GUI. Maybe the "template"/"optionsRequired" aspect should just be a part of the topic object (as it is, in fact in IBlueprintActionManifest), and still just be inserted into two individual topics? As-is, one can create a `template` Part-local AdLib that will be pushed to the `adLib` topic, but a `template` Global AdLib will go to a different topic?

jstarpl:
We'd like to invite the Contributor and anyone else interested in this PR to a Workshop on _Thu 2024-08-22 13:00 - 14:00_ to discuss this and it's sibling PRs. Please e-mail me at jan.starzak@nrk.no for a Microsoft Teams invitation.

packages/live-status-gateway/src/topics/adLibsTopic.ts
scriptorian: templateAdLibs are no longer a separate array. They can now be discovered by finding a template property in the public data.

meteor/client/ui/Shelf/AdLibPanel.tsx
```diff
@@ -487,6 +487,7 @@ export function fetchAndFilter(props: IFetchAndFilterProps): AdLibFetchAndFilter
 								}
 							)
 								.fetch()
+								.filter((action) => !(action.publicData && ((action.publicData as any).template as boolean)))
```
jstarpl, line 490: `publicData` is a blueprints-defined blob, I don't think Core should impose any rules on what various properties can be set there. I'm quite sure someone might want to set `publicData.template` to be the name of a graphics template that the AdLib is using. I don't think they would expect the AdLib to disappear from the Shelf because of that.

```suggestion
```

jstarpl (changes requested):

meteor/client/ui/Shelf/AdLibPanel.tsx
scriptorian: Updated as suggested

jstarpl (approved):

PR #1223, opened 2024-07-08
Support for schema validation of blueprintConfiguration objects

scriptorian:

Feature

When updating blueprintConfiguration objects through the REST API the new configuration is written to the database without any validation that the values conform to any schema.

This feature supports blueprints that define blueprintConfiguration object JSON schemas and provides methods to support automatic, detailed validation of updated configurations against the published schema before being written to the database. Any invalid properties are reported back to the client.

Included is a small addition to the API that allows the blueprintConfiguration objects to be read and updated without including the rest of the studio or showStyle configuration properties.

#### Status

- [x] No unit test changes are needed for this PR.

#### Affected Areas

* This PR affects the REST API implementation.
* Without blueprint changes there will be no change in behaviour.


nytamin:

meteor/server/api/rest/v1/index.ts
```diff
@@ -92,6 +93,19 @@ function extractErrorDetails(e: unknown): string[] | undefined {
 	}
 }
 
+export const checkValidation = (method: string, configValidationMsgs: IConfigMessage[]): void => {
```
nytamin, line 96:
```suggestion
/**
 * Throws if any of the configValidationMsgs indicates that the config is not valid.
 */
export const checkValidation = (method: string, configValidationMsgs: IConfigMessage[]): void => {
```

meteor/server/api/rest/v1/index.ts
```diff
@@ -92,6 +93,19 @@ function extractErrorDetails(e: unknown): string[] | undefined {
...
+	const configValidationOK = configValidationMsgs.reduce((acc, msg) => acc && msg.level === NoteSeverity.INFO, true)
```
nytamin, line 97: What about `NoteSeverity.WARNING`? My initial assumption would be to still allow them, but log the warnings or return the warnings to the caller?

meteor/server/api/rest/v1/showstyles.ts
```diff
@@ -97,17 +110,71 @@ class ShowStylesServerAPI implements ShowStylesRestAPI {
...
+		)
+		checkValidation(`updateShowStyleConfig ${showStyleBaseId}`, blueprintConfigValidation)
+
+		const showStyle = await showStyleBaseFrom(apiShowStyleBase, showStyleBaseId)
+		if (!showStyle) throw new Meteor.Error(400, `Invalid ShowStyleBase`)
+
+		await ShowStyleBases.upsertAsync(showStyleBaseId, showStyle)
+
+		const validation = await validateConfigForShowStyleBase(showStyleBaseId)
+		checkValidation(`updateShowStyleConfig ${showStyleBaseId}`, validation.messages)
```
nytamin, line 174: Why do we do this validation _after_ the `upsertAsync` to write to db? Wouldn't it have been better to have made this validation before the write?
Or am I missing the point? (If so, perhaps add a comment to the code explaining why?)

packages/blueprints-integration/src/api/showStyle.ts
```diff
@@ -169,6 +169,17 @@ export interface ShowStyleBlueprintManifest<TRawConfig = IBlueprintConfig, TProc
 		coreConfig: BlueprintConfigCoreConfig
 	) => TProcessedConfig
 
+	/**
+	 * Validate the blueprint config passed to this blueprint according to the API schema, returning a list of messages to display to the user.
+	 */
+	validateConfigFromAPI?: (context: ICommonContext, apiConfig: object) => Array<IConfigMessage>
+
+	/** transform API blueprint config to the database format */
+	blueprintConfigFromAPI?: (context: ICommonContext, config: object) => TRawConfig
```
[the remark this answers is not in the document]
nytamin: I don't have strong opinions on this one, but since I know there has been discussions regarding internal vs external APIs before, I think the addition of this makes sense. It would be good to add to the description of this that `If not defined, the config will be passed through as a TRawConfig`, though.

packages/blueprints-integration/src/api/showStyle.ts
```diff
@@ -169,6 +169,17 @@ export interface ShowStyleBlueprintManifest<TRawConfig = IBlueprintConfig, TProc
 		coreConfig: BlueprintConfigCoreConfig
 	) => TProcessedConfig
 
+	/**
+	 * Validate the blueprint config passed to this blueprint according to the API schema, returning a list of messages to display to the user.
+	 */
+	validateConfigFromAPI?: (context: ICommonContext, apiConfig: object) => Array<IConfigMessage>
```
nytamin, line 175: We should add to the description what the impact of the returned notes will be.
(This applies to the other methods as well)

meteor/server/api/rest/v1/index.ts
nytamin: Notes from a chat with Simon: Let's consider WARNINGs to be "okay" but log the warnings somewhere.

packages/blueprints-integration/src/api/showStyle.ts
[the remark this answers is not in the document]
scriptorian: Comments have been updated

PR #1226, opened 2024-07-09
Add support for adLibOptions property in executeAdLib

scriptorian:

Feature

The playlist API request executeAdLib allows the client to trigger adLibs that have been created by the blueprints while loading a rundown. The endpoint offers some options for the way these adLibs will be executed though the use of the actionType property that sets a particular trigger mode.

The client discovers the available adLibs and actionTypes by subscribing to the adLibs subscription in the Live Status Gateway (LSG).

This change adds a new parameter adLibOptions to the executeAdLib request, allowing the client to modify specific properties of the adLib not available through trigger modes, for example the required duration of a video clip.

The adLibOptions property should conform to the optionsSchema that is provided in the adLibs subscription of the LSG thanks to PR #1221. The changes in combination allow the options property to be passed to the blueprints and validated before being used to modify the adLib execution.

The optionsSchema is expected to support only simple changes to the rundown-created adLibs and the properties will normally be optional. The exception to this is templateAdLibs where some required properties are not known during rundown loading. The options validation in this case will ensure that all necessary properties are available before executing the adLib.

#### Status

- [x] Relevant unit tests have been checked.

#### Affected Areas

* This PR affects the REST API implementation.
* Without blueprint changes there will be no change in behaviour.


nytamin:

packages/job-worker/src/playout/adlibAction.ts
```diff
@@ -102,6 +102,7 @@ export async function executeAdlibActionAndSaveModel(
 		userData: data.userData,
 		triggerMode: data.triggerMode,
 		privateData: adLibActionDoc?.privateData,
+		publicData: data.actionOptions,
```
jstarpl, line 105: I find this unusual. If I were a blueprint developer, when seeing a `publicData` incoming, I would expect that this is `adLibActionDoc?.publicData`, and not something injected by the user/consumer? Also, what's the difference between this implementation of `publicData` and `userData`? Both are incoming from the user/consumer, so why have them split? Why call it `publicData` if it has nothing to do with the entities `publicData`?

jstarpl (commented):
I think I need a bit more explanation on the purpose of this `adLibOptions`/`actionOptions`/`publicData` - the name of this structure changes through the code and leaves me confused. Would the same purpose be achieved by allowing the AdLib Execution REST endpoint to inject a "custom" `userData` on request? How are `adLibOptions`/`actionOptions`/`publicData` different from `userData`?

jstarpl:
We'd like to invite the Contributor and anyone else interested in this PR to a Workshop on Thu 2024-08-22 13:00 - 14:00 to discuss this and it's sibling PRs. Please e-mail me at [jan.starzak@nrk.no](mailto:jan.starzak@nrk.no) for a Microsoft Teams invitation.

jstarpl (approved):

jstarpl (approved):

PR #1228, opened 2024-07-11
fix: add PM properties to HTTP accessor

nytamin:
## About the Contributor

## Type of Contribution

This is a:

Feature 

## New Behavior

Adds types for the new[ Package Manager HTML-Template rendering](https://github.com/nrkno/sofie-package-manager/pull/193)

## Testing

### Affected areas

This PR only adds types and minor additions in GUI settings, does not affect much else.

## Status

### DIFF
--- a/meteor/client/ui/Settings/Studio/PackageManager.tsx
+++ b/meteor/client/ui/Settings/Studio/PackageManager.tsx
@@ -488,6 +488,36 @@ export const StudioPackageManagerSettings = withTranslation()(
 														{t('Base url to the resource (example: http://myserver/folder)')}
 													</span>
 												</label>
+												<label className="field">
+													<LabelActual label={t('Is Immutable')} />
+													<EditAttribute
+														modifiedClassName="bghl"
+														attribute={`packageContainers.${containerId}.container.accessors.${accessorId}.isImmutable`}
+														obj={this.props.studio}
+														type="checkbox"
+														collection={Studios}
+														className="input text-input input-l"
+													></EditAttribute>
+													<span className="text-s dimmed field-hint">
+														{t('When set, resources are considered immutable, ie they will not change')}
+													</span>
+												</label>
+												<label className="field">
+													<LabelActual label={t('Supports HEAD requests')} />
+													<EditAttribute
+														modifiedClassName="bghl"
+														attribute={`packageContainers.${containerId}.container.accessors.${accessorId}.supportHEAD`}
+														obj={this.props.studio}
+														type="checkbox"
+														collection={Studios}
+														className="input text-input input-l"
+													></EditAttribute>
+													<span className="text-s dimmed field-hint">
+														{t(
+															'Set to true if the HTTP server supports HEAD requests. (otherwise GET requests will be used to check availability)'
+														)}
+													</span>
+												</label>
 
 												<label className="field">
 													<LabelActual label={t('Network Id')} />
--- a/packages/shared-lib/src/package-manager/package.ts
+++ b/packages/shared-lib/src/package-manager/package.ts
@@ -245,10 +245,16 @@ export namespace Accessor {
 		allowWrite: false
 
 		/** Base url (url to the host), for example http://myhost.com/fileShare/ */
-		baseUrl: string
+		baseUrl?: string
 
 		/** Name/Id of the network the share exists on. Used to differ between different local networks. Leave empty if globally accessible. */
 		networkId?: string
+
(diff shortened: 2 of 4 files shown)


meteor/client/ui/Settings/Studio/PackageManager.tsx
```diff
@@ -488,6 +488,36 @@ export const StudioPackageManagerSettings = withTranslation()(
...
+														attribute={`packageContainers.${containerId}.container.accessors.${accessorId}.supportHEAD`}
+														obj={this.props.studio}
+														type="checkbox"
+														collection={Studios}
+														className="input text-input input-l"
+													></EditAttribute>
+													<span className="text-s dimmed field-hint">
+														{t(
+															'Set to true if the HTTP server supports HEAD requests. (otherwise GET requests will be used to check availability)'
+														)}
```
jstarpl, line 518: Any reasonable HTTP server should support HEAD in 2024 - can we switch the default around? To say, "Use GET requests instead of HEAD to check availability"?

`express.js` generates a HEAD handler automatically for GET routes: https://expressjs.com/en/api.html#app.METHOD - I think so does `koa-router`, so I think it's safe to assume that HEAD is supported, _unless_ it isn't, not the other way round.

packages/shared-lib/src/package-manager/package.ts
```diff
@@ -172,6 +177,78 @@ export namespace ExpectedPackage {
 			}
 		}[]
 	}
+	export interface ExpectedPackageHtmlTemplate extends Base {
+		type: PackageType.HTML_TEMPLATE
+		content: {
+			/** path to the HTML template */
+			path: string
+			/** Add prefix to output artifacts */
+			outputPrefix: string
```
jstarpl, line 186: What does this add a prefix _to_? the url filename of the artifact?

packages/shared-lib/src/package-manager/package.ts
```diff
@@ -172,6 +177,78 @@ export namespace ExpectedPackage {
...
+		version: {
+			renderer?: {
+				/** Renderer width, defaults to 1920 */
+				width?: number
+				/** Renderer width, defaults to 1080 */
+				height?: number
+				/** Zoom level, defaults to 1 */
+				zoom?: number
+				/** (defaults to black) */
+				backgroundColor?: string
```
jstarpl, line 197: How is this color specified? CSS Colors? Can this be set to "transparent"? Can this be `background` and not `backgroundColor` so that this is set to a `background-image` of choice?

packages/shared-lib/src/package-manager/package.ts
```diff
@@ -172,6 +177,78 @@ export namespace ExpectedPackage {
...
+				| { do: 'startRecording'; fileName: string }
+				| { do: 'stopRecording' }
+				| { do: 'cropRecording'; fileName: string }
+				| { do: 'executeJs'; js: string }
+				// Store an object in memory
+				| { do: 'storeObject'; key: string; value: Record<string, any> }
+				// Modify an object in memory. Path is a dot-separated string
+				| { do: 'modifyObject'; key: string; path: string; value: any }
+				// Send an object to the renderer as a postMessage
+				| { do: 'injectObject'; key: string }
```
jstarpl, line 238: I find the naming here a bit confusing - these actions are not related - `storeObject` and `injectObject`? Since there is no `retrieveObject`, maybe it would make more sense to call it `setObject` or even `setValue` if there's no reason to limit this to objects? and then change `injectObject` into `postMessage`, since that's what's going to happen?

Also I suggest not saying "memory" and instead say "page context".

```suggestion
				// Store an object in memory
				| { do: 'setValue'; key: string; value: any }
				// Modify an object in memory. Path is a dot-separated string
				| { do: 'modifyObject'; key: string; path: string; value: any }
				// Send an object to the renderer as a postMessage
				| { do: 'postMessage'; value: string }
```

meteor/client/ui/Settings/Studio/PackageManager.tsx
nytamin: Yeah, that's a good point, that would also be solve a backwards compatibility issue.

packages/shared-lib/src/package-manager/package.ts
nytamin: Yes, the output of this package are defined by the various defined steps. This adds a prefix to them. Hmm, perhaps we should remove this altogether..?

nytamin:
superseded by https://github.com/nrkno/sofie-core/pull/1247

PR #1232, opened 2024-07-29
feat: rest api clear multiple sourcelayers SOFIE-3314

Julusian:
## About the Contributor

## Type of Contribution

This is a: Feature 

## New Behavior
Adds a new endpoint to the rest api

## Testing

### Affected areas

* This PR affects the rest api

## Status


meteor/server/api/rest/v1/playlists.ts
```diff
@@ -786,6 +786,27 @@ export function registerRoutes(registerRoute: APIRegisterHook<PlaylistsRestAPI>)
...
+		playlistsAPIFactory,
+		async (serverAPI, connection, event, params, _) => {
+			const playlistId = protectString<RundownPlaylistId>(params.playlistId)
+			const sourceLayerIdsStr = params.sourceLayerIds
+			logger.info(`API DELETE: sourceLayers ${playlistId} ${sourceLayerIdsStr}`)
+
+			check(playlistId, String)
+			check(sourceLayerIdsStr, String)
+
+			const sourceLayerIds = sourceLayerIdsStr.split(',')
```
jesperstarkar, line 805: The comma-separated array format for URL params, is that a first for this API? Do we need explicit description of the expected format of the input in the API?

Also, in terms of REST, are we sure this goes as a param and not in the request body?

jesperstarkar (commented):
Unsure about the array-format for the URL param. 

Other than that, not able to verify more than just read through (static analysis). But I don't see any obvious flaws.

meteor/server/api/rest/v1/playlists.ts
Julusian: I think it is a first, I would expect that the syntax shouldn't be too unsurprising as comma separated is an openapi convention. https://swagger.io/docs/specification/serialization/

I don't know if it would make more sense in the body, I was trying to keep it match the existing api structure which doesn't make much use of the body

jesperstarkar: Yeah, I'm not questioning the format, just if we need to document how the param is intended to be used.

And for the question about the body, I don't have any opinion.

meteor/server/api/rest/v1/playlists.ts
```diff
@@ -786,6 +786,27 @@ export function registerRoutes(registerRoute: APIRegisterHook<PlaylistsRestAPI>)
 		}
 	)
 
+	registerRoute<{ playlistId: string; sourceLayerIds: string }, never, void>(
+		'delete',
+		'/playlists/:playlistId/sourceLayers/:sourceLayerIds',
```
ianshade, line 791: A DELETE on a sourceLayers resource(s) implies that (1) something will be deleted, and (2) it will be the sourceLayers, neither being the case here, because it will in fact result in _pieces_ being _stopped_. IMO this should be a POST to `/playlists/:playlistId/clear-source-layers` with a body containing an array of sourceLayerIds.

ianshade: ...and the existing endpoint that clears a single layer should be marked as deprecated.

jstarpl: I think it being a DELETE does make sense - re-running the same request will end up with the same result: the content will be cleared from the `sourceLayer` "bucket". It therefore matches a DELETE, even if it's not DELETING a piece but rather a collection of stuff.

One thing about marking stuff as deprecated: we need to make sure to properly version the change. I guess deprecating an endpoint is a breaking change, so we need to properly re-name, redirect, etc. For the new major version.

Julusian: >  I guess deprecating an endpoint is a breaking change, so we need to properly re-name, redirect, etc. For the new major version.

No, deprecating is merely a warning to consumers that this is not recommended and will likely be going away at some point, the old method will still be there and functional, unless we want to skip deprecating it and jump straight to removing it.
Keeping the old implementation of this around has barely any cost, so isn't worth the cost of making this a breaking change

Julusian: I discussed this with @nytamin, and the conclusion was that having this as `/clear-sourcelayers` is clearer about the intent and fits better with the other operations in this api.
So we shall go with that approach instead.

meteor/server/api/rest/v1/playlists.ts
Julusian: following other changes, this is now a body parameter

nytamin (approved):

PR #1243, opened 2024-08-19
Minor performance improvements (SOFIE-3388)

nytamin:
## About the Contributor

## Type of Contribution

This is a:

Minor performance improvements

## New Behavior

Slight improvements in performance.

## Testing

### Affected areas

This PR touches playout features, but should not have any impact on functionality.

## Status


Julusian (commented):
Would it be a good idea to only do the `logger.silly(` call when that level has somewhere to go? With this change we are still paying the cost of the JSON.strinigify, even if it doesn't go anywhere.

I am a little worried that changing the `Starting work .....` line to be silly only will make log reading more difficult. I think that is the line that I am often relying on to understand a sequence of events, although this could probably be solved by making sure each job type logs the relevant ids

PR #1290, opened 2024-10-17
fix(job-worker): queue a PartInstanceTimingEvent on deactivation (SOFIE-3470)

jstarpl:
## About the Contributor

## Type of Contribution

This is a:

Bug fix

## Current Behavior

It's impossible to send metadata using the `onRundownTimingEvent` callback for the very last PartInstance of the Rundown, since the callback is not called on deactivating the Rundown.

## New Behavior

`onRundownTimingEvent` is called on deactivating a Rundown with a currentPartInstance.

## Testing

### Affected areas

This PR affects Rundown deactivation and metadata flow.

## Time Frame

We intend to finish the development on this feature in two weeks time.

## Status

### DIFF
--- a/packages/job-worker/src/playout/model/PlayoutPartInstanceModel.ts
+++ b/packages/job-worker/src/playout/model/PlayoutPartInstanceModel.ts
@@ -176,6 +176,11 @@ export interface PlayoutPartInstanceModel {
 	 * @param time Reported stopped time
 	 */
 	setReportedStoppedPlayback(time: Time): boolean
+	/**
+	 * Set the Reported stopped playback time, including still-playing PieceInstances
+	 * @param time Reported stopped time on all available objects
+	 */
+	setReportedStoppedPlaybackWithPieceInstances(time: Time): boolean
 
 	/**
 	 * Set the rank of this PartInstance, to update it's position in the Segment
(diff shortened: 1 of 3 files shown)


nytamin (approved):

packages/job-worker/src/playout/model/PlayoutPartInstanceModel.ts
```diff
@@ -176,6 +176,11 @@ export interface PlayoutPartInstanceModel {
 	 * @param time Reported stopped time
 	 */
 	setReportedStoppedPlayback(time: Time): boolean
+	/**
+	 * Set the Reported stopped playback time, including still-playing PieceInstances
+	 * @param time Reported stopped time on all available objects
+	 */
+	setReportedStoppedPlaybackWithPieceInstances(time: Time): boolean
```
Julusian, line 183: should this be a public method on the model, or is this an internal implementation detail?

jstarpl: This is a very good question! I didn't want to leak the implementation detail of how `pieceInstances` are organized and managed to the outside, because it didn't feel like whoever is calling this _should_ need to know what else needs to be called for the desired effect.

Would you say it makes more sense to just reuse `setReportedStoppedPlayback` and then have the caller update the individual pieceInstances using their model methods?

Julusian: I don't mind the method existing, my only thought was that this is called only by `PlayoutModelImpl` which knows that it must be working with a `PlayoutPartInstanceModelImpl`.
I now realise that the types don't quite line up with this, because of `this.currentPartInstance` returning the interface, but something could be done so that where you are calling `setReportedStoppedPlaybackWithPieceInstances` it knows it is the Impl type. 

I don't really mind either way though, might be best to leave it as it avoids needing some type fiddling

PR #1365, opened 2025-01-24
fix(Cronjob): don't restart CasparCG using the nightly cronjob if there's a Rundown active in the Studio (SOFIE-3618)

jstarpl:
## Type of Contribution

This is a:

Bug fix

## Current Behavior

CasparCG servers will be restarted during "Low Season" (between 4:00 and 5:00) if the CasparCG restart cronjob is enabled

## New Behavior

CasparCG cronjob will check if there's an active rundown in the Studio that owns the CasparCG to be restarted. If there is, the restart will be skipped until next day.

## Testing

## Status

### DIFF
--- a/meteor/server/cronjobs.ts
+++ b/meteor/server/cronjobs.ts
@@ -27,6 +27,7 @@ import {
 } from '@sofie-automation/corelib/dist/TranslatableMessage'
 import { applyAndValidateOverrides } from '@sofie-automation/corelib/dist/settings/objectWithOverrides'
 import { DBRundownPlaylist } from '@sofie-automation/corelib/dist/dataModel/RundownPlaylist'
+import { StudioId } from '@sofie-automation/corelib/dist/dataModel/Ids'
 
 const lowPrioFcn = (fcn: () => any) => {
 	// Do it at a random time in the future:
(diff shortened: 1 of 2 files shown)


meteor/server/cronjobs.ts
```diff
@@ -90,20 +91,38 @@ async function restartCasparCG(systemSettings: ICoreSystemSettings | undefined,
...
+			}
+		) as Promise<
+			Array<Pick<PeripheralDevice, '_id' | 'subType' | 'parentDeviceId' | 'lastSeen' | 'studioAndConfigId'>>
+		>,
+		RundownPlaylists.findFetchAsync(
+			{
+				activationId: {
+					$exists: true,
+				},
+			},
```
nytamin, line 117: _This is perhaps nit-picking, as most/all Sofie installations we know of are single-studio at the moment, but since the data model allows for multi-studio:_

This query will fail to match any [indexes we've set up in MongoDB](https://github.com/nrkno/sofie-core/blob/6e04a7f53cbe088c3db88bdc394675c27b456706/meteor/server/collections/rundown.ts#L144). Could we instead first query for the PeripheralDevices, and then use their studioId to filter in the query?
```
{
  studioId: {$in: relevantStudioIds },
  activationId: { $exists: true }
}
```

jstarpl: Yeah, I guess I wanted to be fancy, but I'm good with doing this in series.

PR #1370, opened 2025-01-28
feat(LSG): add buckets topic

ianshade:
## About the Contributor

This pull request is posted on behalf of TV 2 Norge

## Type of Contribution

This is a:

Feature

## New Behavior

Buckets with their adLibs and actions are published as an LSG topic. Only adLibs and actions valid for all variants (showStyleVariantId === null) are included, but further contributions are welcome to lift this limitation, should anyone need it.

## Testing

### Affected areas

This PR affects the Live Status Gateway functionally. In order to achieve that, some types and a publication used by other packages in the core monorepo are moved around, some publications have their parameters changed

## Time Frame

Not urgent, but it would be nice to have in release 52

## Other Information

These changes were so far tested only on release50, with release51 LSG.
This branch also includes changes from https://github.com/nrkno/sofie-core/pull/1367. Before that one is merged, and this one is rebased, only the latest commit on this branch is to be reviewed.

## Status


packages/documentation/versioned_docs/version-1.50.0/for-developers/data-model.md
```diff
@@ -24,7 +24,7 @@ Currently, there is not a very clearly defined flow for modifying these document
 This includes:
 
 - [Blueprints](https://github.com/nrkno/sofie-core/blob/master/packages/corelib/src/dataModel/Blueprint.ts)
-- [Buckets](https://github.com/nrkno/sofie-core/blob/master/meteor/lib/collections/Buckets.ts)
+- [Buckets](https://github.com/nrkno/sofie-core/blob/master/packages/corelib/src/dataModel/Bucket.ts)
```
jstarpl, line 27: We should probably let this be? This refers to historical information anyway, so will not align with the rest of documentation.

packages/live-status-gateway/src/topics/bucketsTopic.ts
```diff
@@ -0,0 +1,136 @@
...
+					actionType: triggerModes,
+					tags: action.display.tags,
+					externalId: action.externalId,
+					publicData: action.publicData,
+				})
+			})
+			return {
+				id: bucketId,
+				name: bucket.name,
+				adLibs: [...bucketAdLibs, ...bucketAdLibActions],
```
jstarpl, line 98: Should this be sorted according to the `_rank` of the original objects? I feel like it probably should, considering this is returned as an ordered array? As a Sofie user, I would expect the order to be retained (since I can reorder them in the UI). Especially, since if they aren't sorted, any ordering is gone and there's no way of recovering it. I would expect the users to be using reordering for Bucket organization (keeping the "hottest" clip at the begining or whatever else makes sense in one's workflow). I see that `adLibsTopic` keeps sorting, so maybe we could use a similar thing here.

jstarpl (commented):
I don't have a lot of comments, but I think it would be nice if sorting of the buckets and bucket items was retained in the LSG output, like it is in other topics.

jstarpl (changes requested):
Pressed the wrong button - I think it would be good if the new topic retained the ordering of Buckets and Bucket Adlibs, like the Adlib topic does.

packages/live-status-gateway/src/topics/bucketsTopic.ts
ianshade: I added sorting in the latest commit

PR #1374, opened 2025-01-29
feat: implement tally for device trigger previews

ianshade:
## About the Contributor

This pull request is posted on behalf of TV 2 Norge

## Type of Contribution

This is a:

Feature

## New Behavior

Tally information, based on the existing tally-tags system (`currentPieceTags`) is added to the device trigger previews

## Testing

### Affected areas

This PR affects the device triggers publication that the Input Gateway consumes

## Time Frame

Not urgent, but it would be nice to have in release 52

## Status

### DIFF
--- a/meteor/server/api/deviceTriggers/PieceInstancesObserver.ts
+++ b/meteor/server/api/deviceTriggers/PieceInstancesObserver.ts
@@ -94,5 +94,6 @@ export class PieceInstancesObserver {
 		this.#cancelCache()
 		this.#observers.forEach((observer) => observer.stop())
 		this.#cleanup?.()
+		this.#cleanup = undefined
 	}
 }
--- a/meteor/server/api/deviceTriggers/PieceInstancesObserver.ts
+++ b/meteor/server/api/deviceTriggers/PieceInstancesObserver.ts
@@ -1,0 +1,98 @@
+import { Meteor } from 'meteor/meteor'
+import { RundownPlaylistActivationId, ShowStyleBaseId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+import { RundownPlaylists, ShowStyleBases, PieceInstances, PartInstances } from '../../collections'
+import { logger } from '../../logging'
+import { rundownPlaylistFieldSpecifier } from './reactiveContentCache'
+import {
+	ContentCache,
+	createReactiveContentCache,
+	partInstanceFieldSpecifier,
+	pieceInstanceFieldSpecifier,
+} from './reactiveContentCacheForPieceInstances'
+import { waitForAllObserversReady } from '../../publications/lib/lib'
+
+const REACTIVITY_DEBOUNCE = 20
+
+type ChangedHandler = (cache: ContentCache) => () => void
+
+export class PieceInstancesObserver {
+	#observers: Meteor.LiveQueryHandle[] = []
+	#cache: ContentCache
+	#cancelCache: () => void
+	#cleanup: (() => void) | undefined
+	#disposed = false
+
+	constructor(onChanged: ChangedHandler) {
+		const { cache, cancel: cancelCache } = createReactiveContentCache(() => {
+			this.#cleanup = onChanged(cache)
+			if (this.#disposed) this.#cleanup()
+		}, REACTIVITY_DEBOUNCE)
+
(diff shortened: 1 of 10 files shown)


meteor/server/api/deviceTriggers/TagsService.ts
```diff
@@ -0,0 +1,168 @@
...
+			(!this.areSetsEqual(this.onAirPiecesTags, activePieceInstancesTags) ||
+				!this.areSetsEqual(this.nextPiecesTags, nextPieceInstancesTags)) &&
+			(this.doSetsIntersect(activePieceInstancesTags, this.tagsObservedByTriggers) ||
+				this.doSetsIntersect(nextPieceInstancesTags, this.tagsObservedByTriggers) ||
+				this.doSetsIntersect(this.onAirPiecesTags, this.tagsObservedByTriggers) ||
+				this.doSetsIntersect(this.nextPiecesTags, this.tagsObservedByTriggers))
+		)
+	}
+
+	protected areSetsEqual<T>(a: Set<T>, b: Set<T>): boolean {
```
jstarpl, line 125: Why are these methods protected? I'm also wondering if they shouldn't live in some lib file somewhere, considering they are super-generic and don't reference the members of this class at all?

meteor/server/api/deviceTriggers/StudioDeviceTriggerManager.ts
```diff
@@ -224,6 +238,18 @@ export class StudioDeviceTriggerManager {
 		actionManager.deleteActionsOtherThan(touchedActionIds)
 	}
 
+	protected async updateTriggersFromLastCache(): Promise<void> {
```
jstarpl, line 241: Why is this method `protected`? I don't see this being subclassed anywhere?

Also, I feel that maybe if there's a need to have `cache` be persisted, it makes more sense to just change the signature of the `updateTriggers` altogether and add a `setCache` method? I think I'll become confused next time I'll be looking at this code, trying to figure out which of these (`updateTriggersFromLastCache` or `updateTriggers`) I should use. I find that this PR breaks the state-less nature of `StudioDeviceTriggerManager` - which is fine, but if that's the case, let's just go all in.

jstarpl (changes requested):

packages/shared-lib/src/input-gateway/deviceTriggerPreviews.ts
```diff
@@ -61,4 +61,6 @@ export type PreviewWrappedAdLib = Omit<IWrappedAdLibBase, '_id'> & {
 		  }
 		| undefined
 	styleClassNames: string | undefined
+	isCurrent: boolean | undefined
```
jstarpl, line 64: I'm not sure about calling this state `isCurrent`. My reading is that "current" is a state of the Part, whereas this doesn't just mean "AdLib is current", it actually means "AdLib is playing", since this will be reactive to actual playout and any limiations of the corresponding Piece's `enable` property. Maybe an `isOnAir` would make the meaning more clear?

meteor/server/api/deviceTriggers/PieceInstancesObserver.ts
```diff
@@ -0,0 +1,99 @@
...
+	public get cache(): ContentCache {
+		return this.#cache
+	}
+
+	public stop = (): void => {
+		this.#disposed = true
+		this.#cancelCache()
+		this.#observers.forEach((observer) => observer.stop())
+		this.#cleanup?.()
+		this.#cleanup = undefined
```
evenstensberg, line 97: Just a nitpick. If you assign this to undefined, the mental logic of this would be that the value was never set. However, if you assign it to null (a falsy value that works similar to undefined as you know) it implies more that the function resets the parameter.

jstarpl: > However, if you assign it to null (a falsy value that works similar to undefined as you know)

True, but I `undefined` meaning 'unset' is an established pattern in this codebase, and it doesn't really matter if the function resets the parameter or not - the behavior should be the same regardless if this is the value hasn't been set before or if it has been reset. With that said, since `typeof null` is `object`, using `undefined` gives less chance of it being mistaken for something else, even if it's falsy.

PR #1417, opened 2025-04-02
Sofie Core image should run as a non-root user

siljekristensen:
## About the Contributor

## Type of Contribution

This is a change in how the Dockerfile is constructed.

## Current Behavior

Currently the Sofie Core container image is run as the root user. When running Core in Kubernetes with a security context defined for the deployment (to make sure it runs as a non-root user), Core will not be able to start if any of the files needed for startup are only readable for the root user. We've currently seen this happen with the release52 image (ghcr.io/nrkno/sofie-core-server-core:release52, sha256:2498251605124571a6c4cac250d402b7912f9d8661136eef664585845d55ecb3). We don't see the issue in the release51 or the release53 image. 

## New Behavior

We make sure all the files can be accessed by user id 1000 and we set the `USER 1000` instruction to make the image run as that user.

## Testing

### Affected areas

This PR affects running of the Sofie Core docker image.

## Status


jstarpl:
Looks good to me, I think one final step in this PR would be to document that the docker images built from these repositories will run as UID 1000. Maybe here somewhere: https://nrkno.github.io/sofie-core/docs/user-guide/installation/installing-sofie-server-core ? I suppose this also means you _can't_ run them using `--user`?

jstarpl (approved):

Julusian:
This changes just the `Dockerfile.circle` versions and not the self-container building `Dockerfile`s. It would be good to have them both updated, to avoid confusion about different behaviour depending on which was used

jstarpl (changes requested):
Julian has a good point, so I'm gonna mark this as "Request changes"
> This changes just the Dockerfile.circle versions and not the more self-contained Dockerfiles. It would be good to have them both updated, to avoid confusion about different behaviour depending on which was used

jstarpl (approved):

jstarpl:
Considering that this changes requires _some_ oversight before updating a deployment, I propose that we merge this into `release52`, and that this will take effect once 1.52.0 is released.

PR #1495, opened 2025-07-07
feat: Automatically assign peripheral subdevices to parent during a studio upgrade

scriptorian:
## About the Contributor

## Type of Contribution

This is a:

Code improvement

## New Behaviour

During a studio upgrade a list of parent devices of each type is created. Sub-devices of each type being upgraded have their parent device id set if there is only one instance of that type of parent.

## Testing

### Affected areas

* This PR affects the studio blueprint upgrade.

## Time Frame

We would like to get this merged into the in-development release.

## Status

### DIFF
--- a/packages/job-worker/src/playout/upgrade.ts
+++ b/packages/job-worker/src/playout/upgrade.ts
@@ -28,6 +28,12 @@ import {
 import { compileCoreConfigValues } from '../blueprints/config.js'
 import { CommonContext } from '../blueprints/context/index.js'
 import { JobContext } from '../jobs/index.js'
+import {
+	PeripheralDevice,
+	PeripheralDeviceType,
+	PeripheralDeviceCategory,
+	PERIPHERAL_SUBTYPE_PROCESS,
+} from '@sofie-automation/corelib/dist/dataModel/PeripheralDevice'
 import { FixUpBlueprintConfigContext } from '@sofie-automation/corelib/dist/fixUpBlueprintConfig/context'
 import { DEFAULT_MINIMUM_TAKE_SPAN } from '@sofie-automation/shared-lib/dist/core/constants'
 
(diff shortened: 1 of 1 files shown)


packages/job-worker/src/playout/upgrade.ts
```diff
@@ -62,29 +68,54 @@ export async function handleBlueprintUpgradeForStudio(context: JobContext, _data
...
-				options: dev[1],
-			}),
-		])
+		Object.entries<unknown>(result.ingestDevices ?? {}).map((dev) => {
+			const { ingestDeviceType, ...payload } = dev[1] as any
+			return [
+				dev[0],
+				literal<Complete<StudioIngestDevice>>({
+					peripheralDeviceId:
+						ingestDeviceType === PeripheralDeviceType.SPREADSHEET ? spreadsheetGatewayId : mosGatewayId,
```
jstarpl, line 108: I think that this needs some code comments to explain why `PeripheralDeviceType.SPREADSHEET` needs specific handling? This also leaves `PeripheralDeviceType.INEWS` dangling, since it will not be handled by neither `spreadsheetGatewayId` nor `mosGatewayId`. Perhaps, the same check as for `StudioPlayoutDevice` would suffice? I.e.


```suggestion
					peripheralDeviceId:
						ingestIds.length === 1 ? ingestIds[0] : undefined,
```

L96 makes the assumption that if there is _a single playout device_, then the output of `result.playoutDevices` should be compatible with it. Same happens at L118 for Input Gateway. Would it be possible to make the same assumption here? If not, then I think this needs to be commented on in the code, so that this behavior is preserved, since it's not obvious (at least to me) why there's a difference.

jstarpl (changes requested):
It's not quite clear to me why there's a difference in behavior for Ingest Gateways and other Peripheral Devices.

scriptorian:
I will close this PR now and in due course submit a new PR that exploits the new parent device settings record of the upgrade object.

PR #1497, opened 2025-07-24
feat: allow part to be queued from onTake

ianshade:
## About the Contributor

This pull request is posted on behalf of TV 2 Norge.

## Type of Contribution

This is a:

Feature

## Current Behavior

Not possible to queue a part from `onTake()`. Not possible to queue an extra part after an adLib action queues one.

## New Behavior

Blueprints can queue a dynamic part during the `onTake()` callback. A `queuePartAfterTake()` method was added `IOnTakeContext`. Naming of the method is quite verbose, to make it clear that it queues the new part after the part currently being Taken. In the future there might be a need to add another method that e.g. queues a part instead of performing the Take.
Actions that take a part, are able to queue an additional one, using `queuePartAfterTake()` from `IActionExecutionContext`.

## Testing

### Affected areas

This PR affects the playout logic, specifically what happens during a Take. It introduces a non-breaking addition to the Blueprints API.

## Status

### DIFF
--- a/packages/job-worker/src/playout/take.ts
+++ b/packages/job-worker/src/playout/take.ts
@@ -258,16 +258,6 @@ export async function performTakeToNextedPart(
 	const wasLooping = playoutModel.playlist.quickLoop?.running
 	playoutModel.updateQuickLoopState()
 
-	const nextPart = selectNextPart(
-		context,
-		playoutModel.playlist,
-		takePartInstance.partInstance,
-		null,
-		playoutModel.getAllOrderedSegments(),
-		playoutModel.getAllOrderedParts(),
-		{ ignoreUnplayable: true, ignoreQuickLoop: false }
-	)
-
 	takePartInstance.setTaken(now, timeOffset)
 
 	if (wasLooping) {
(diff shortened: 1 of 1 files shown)


packages/job-worker/src/playout/take.ts
```diff
@@ -264,8 +270,12 @@ export async function performTakeToNextedPart(
 		resetPreviousSegmentIfLooping(context, playoutModel)
 	}
 
-	// Once everything is synced, we can choose the next part
-	await setNextPart(context, playoutModel, nextPart, false)
+	if (queuePart) {
+		await queuePart()
+	} else {
+		// Once everything is synced, we can choose the next part
+		await setNextPart(context, playoutModel, nextPart, false)
```
Julusian, line 288: The definition of nextPart would be good to move into this else block, so that it is only run when needed and make sure it doesnt get used elsewhere accidentally

packages/job-worker/src/playout/take.ts
```diff
@@ -323,6 +335,12 @@ async function executeOnTakeCallback(
 			await blueprint.blueprint.onTake(onSetAsNextContext, blueprintPersistentState)
 			await applyOnTakeSideEffects(context, playoutModel, onSetAsNextContext)
 			isTakeAborted = onSetAsNextContext.isTakeAborted
+			if (onSetAsNextContext.partToQueue) {
+				const partToQueue = onSetAsNextContext.partToQueue
+				queuePart = async () => {
+					await actionService.queuePart(partToQueue.rawPart, partToQueue.rawPieces)
```
Julusian: I don't like how this function gets returned, and am also wondering does this even work correctly?

Largely because this working is reliant on the `nextPartState` in the service as being `NONE`. So if any other changes are made to the current part in the callback then this will fail.

Being pickier (perhaps too much so), this relies on the implementation detail of how the actionService gets currentPartInstance. If that were changed to be a less dynamic reference, then this would break (I am pretty sure that nowhere else relies on this mutability).

PR #1505, opened 2025-07-31
Set sub-device peripheralDeviceId from deviceOptions parentDeviceName

scriptorian:
## About the Contributor

## Type of Contribution

This is a:

Feature

## Current Behaviour

When configuring a new peripheral sub-device in a Studio upgrade the peripheralDeviceId would be left undefined.

## New Behaviour

The updated code allows the device options of the new sub-device to specify a parentDeviceName property that is then used to find a peripheralDeviceId with a matching name.
Parent devices append an extra string to the configured name so the search for a match uses only the first characters of the parent device _id property.

## Testing

### Affected areas

This PR affects peripheral sub-devices configured through a studio upgrade operation triggered by the API.

## Time Frame

We would like to get this merged into the in-development release.

## Status

### DIFF
--- a/packages/job-worker/src/playout/upgrade.ts
+++ b/packages/job-worker/src/playout/upgrade.ts
@@ -18,7 +18,7 @@ import {
 	StudioRouteSetExclusivityGroup,
 } from '@sofie-automation/corelib/dist/dataModel/Studio'
 import { Complete, clone, literal } from '@sofie-automation/corelib/dist/lib'
-import { protectString } from '@sofie-automation/corelib/dist/protectedString'
+import { protectString, unprotectString } from '@sofie-automation/corelib/dist/protectedString'
 import { applyAndValidateOverrides } from '@sofie-automation/corelib/dist/settings/objectWithOverrides'
 import { wrapTranslatableMessageFromBlueprints } from '@sofie-automation/corelib/dist/TranslatableMessage'
 import {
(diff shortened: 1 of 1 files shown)


packages/job-worker/src/playout/upgrade.ts
```diff
@@ -58,37 +59,61 @@ export async function handleBlueprintUpgradeForStudio(context: JobContext, _data
...
 	const playoutDevices = Object.fromEntries(
-		Object.entries<TSR.DeviceOptionsAny>(result.playoutDevices ?? {}).map((dev) => [
-			dev[0],
-			literal<Complete<StudioPlayoutDevice>>({
-				peripheralDeviceId: undefined,
-				options: dev[1],
-			}),
-		])
+		Object.entries<TSR.DeviceOptionsAny>(result.playoutDevices ?? {}).map((dev) => {
+			const { parentDeviceName, ...payload } = dev[1] as any
```
Julusian: I'm not keen on these currently, it means that this feature will be rather secret as it is both not in any documentation and not in the types (It can't be in the types without a larger rework as most of these are typed as unknown).

Ideally the type of this `playoutDevices` object should become something like `Record<string, { parentDeviceName?: string, options: TSR.DeviceOptionsAny }>`. While that is a breaking change in these types, we don't guarantee stability, so should be ok?

packages/job-worker/src/playout/upgrade.ts
```diff
@@ -58,37 +59,61 @@ export async function handleBlueprintUpgradeForStudio(context: JobContext, _data
...
+			return [
+				dev[0],
+				literal<Complete<StudioPlayoutDevice>>({
+					peripheralDeviceId: peripheralDevices.find(
+						(p) => unprotectString(p._id).substring(0, parentDeviceName?.length) === parentDeviceName
```
Julusian: I'm not keen on this matching by name. 

Could this match against `p.studioAndConfigId?.configId` instead? That will make this more predictable, and tied to a value that the blueprints have some control over (those values will match the keys in `result.parentDevices`).  

I suppose that longer term, maybe these objects in the db should be changed to hold these configId values instead of peripheralDeviceId, I think that would make more sense than the current approach.

packages/job-worker/src/playout/upgrade.ts
scriptorian: I have updated the sub-device definitions as you proposed.

packages/job-worker/src/playout/upgrade.ts
scriptorian: I have used the configId as you proposed. The upgrade code is definitely clearer as a result

PR #1510, opened 2025-08-26
Fix: Mos-Gateway Implement required callbacks for profile 4

nytamin:
## About the Contributor

## Type of Contribution

This is a:

Bug fix

## Current Behavior

If running Mos Gateway in strict mode, there is an error from Mos-connection library that `onRequestAllRunningOrders` is missing.

This is technically required, and has always been. Sofie doesn't really support the functionality itself so I'll just leave the implementation a stub for now.

## New Behavior

No error.

## Testing

## Time Frame

Not urgent, but annoying.

## Status

### DIFF
--- a/packages/mos-gateway/src/mosHandler.ts
+++ b/packages/mos-gateway/src/mosHandler.ts
@@ -371,6 +371,11 @@ export class MosHandler {
 				// Profile 3: -------------------------------------------------
 				// Profile 4: -------------------------------------------------
 				// onStory: (cb: (story: IMOSROFullStory) => Promise<any>) => void
+				mosDevice.onRequestAllRunningOrders(async () => {
+					// MOSDevice >>>> Core
+					// Not implemented (yet)
+					return []
+				})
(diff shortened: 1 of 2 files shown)


packages/mos-gateway/src/mosHandler.ts
```diff
@@ -371,6 +371,11 @@ export class MosHandler {
 				// Profile 3: -------------------------------------------------
 				// Profile 4: -------------------------------------------------
 				// onStory: (cb: (story: IMOSROFullStory) => Promise<any>) => void
+				mosDevice.onRequestAllRunningOrders(async () => {
+					// MOSDevice >>>> Core
+					// Not implemented (yet)
```
Julusian: I feel like this comment is misleading, it suggests that this is something that we should be implementing at some point.
But is this something we should be implementing? The docs suggest to me that as we aren't a NCS, then we don't have anything to return. If my reading is correct, then it would be good to state this in the comment

PR #1513, opened 2025-08-28
fix: bug in onTimelineGenerate: Use same time for "now" everywhere

nytamin:
## About the Contributor

## Type of Contribution

This is a: Bug fix

## Current Behavior

We have found a issue where the timings of pieces are not monotonic.
This only affects infinite pieces, and probably only when `multiGatewayMode` is enabled.

How to reproduce:

In a blueprints adlibAction:
```typescript
// Stop the previous (infinite) piece
await context.stopPiecesOnLayers(['myLayer'])

await insertPiecesTyped(context, 'current', {
	[...],
	sourceLayerId: 'myLayer',
	enable: { start: 'now' },
	lifespan: PieceLifespan.OutOnShowStyleEnd,
})
```
This results in a virtual piece being inserted to stop the previous infinite `OutOnShowStyleEnd` piece, [at time ](https://github.com/Sofie-Automation/sofie-core/blob/f222545db5a68de63ce45ecff3c839b1b8a86311/packages/job-worker/src/playout/adlibUtils.ts#L282) `getCurrentTime()+calculateNowOffsetLatency` (this is correct).
Then a pieceInstance is inserted at time `"now"`, which is later resolved using `getCurrentTime()`.
In our case, this ends up in that the virtual piece ends up _after_ the inserted Piece, thus instantly ending it.

## New Behavior

The same `getCurrentTime()+calculateNowOffsetLatency` time is used in both `deNowifyMultiGatewayTimeline` and in `getTimelineRundown`.

## Testing

### Affected areas

This PR affects the playout logic in general.

## Time Frame

* This is a critical bug for us, but not urgent to merge into upstream.

## Other Information

Disclaimer: I have not gone through the other places where `getCurrentTime()` is used, so there _might_ be other places `calculateNowOffsetLatency` should be used.

## Status

### DIFF
--- a/packages/job-worker/src/playout/timeline/generate.ts
+++ b/packages/job-worker/src/playout/timeline/generate.ts
@@ -40,7 +40,7 @@ import { deserializePieceTimelineObjectsBlob } from '@sofie-automation/corelib/d
 import { convertResolvedPieceInstanceToBlueprints } from '../../blueprints/context/lib'
 import { buildTimelineObjsForRundown, RundownTimelineTimingContext } from './rundown'
 import { SourceLayers } from '@sofie-automation/corelib/dist/dataModel/ShowStyleBase'
-import { deNowifyMultiGatewayTimeline } from './multi-gateway'
+import { calculateNowOffsetLatency, deNowifyMultiGatewayTimeline } from './multi-gateway'
 import { validateTimeline } from 'superfly-timeline'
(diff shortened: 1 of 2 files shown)


nytamin:
@Julusian mentioned an issue with that we're now generating a `now`-time from a point-in-time _before_ `getTimelineRundown` has run, which might cause issues since the variability in execution-time of `getTimelineRundown` will now play a part, which might not be ideal.

jstarpl:
I've drafted an alternative implementation to address the comments here: https://github.com/Sofie-Automation/sofie-core/pull/1546

PR #1519, opened 2025-09-04
Fix/various fixes

nytamin:
## About the Contributor

## Type of Contribution

This is a: Bug fix 

## Current Behavior

This PR addresses a few issues we've seen in prod:

1. `cannot read reloadProps of undefined`
2. `MaxListenersExceededWarning` message related to `CoreConnectionChild.init()`
3. `Match failed:  expected string, got null`

## New Behavior

1. Error should now be gone.
2. Increases `setMaxListeners` depending of number of `children`
   Interestingly, the `this._children` property was unused. It must have been refactored away at some point?
3. No error (null is allowed in types)

## Time Frame

### DIFF
--- a/packages/server-core-integration/src/__tests__/index.spec.ts
+++ b/packages/server-core-integration/src/__tests__/index.spec.ts
@@ -482,21 +482,21 @@ describe('coreConnection', () => {
 		})
 
 		// Set child connection:
-		const coreChild = await coreParent.createChild({
+		const coreChild0 = await coreParent.createChild({
 			deviceId: protectString('JestTestChild'),
 			deviceSubType: PERIPHERAL_SUBTYPE_PROCESS,
 			deviceName: 'Jest test framework child',
 		})
 
 		const onChildError = jest.fn()
-		coreChild.on('error', onChildError)
+		coreChild0.on('error', onChildError)
 
-		expect(coreChild.connected).toEqual(true)
+		expect(coreChild0.connected).toEqual(true)
 
 		// Close parent connection:
-		await coreParent.destroy()
+		await coreParent.destroy() // This will also close all children
 
-		expect(coreChild.connected).toEqual(false)
+		expect(coreChild0.connected).toEqual(false)
 
 		// connect parent again:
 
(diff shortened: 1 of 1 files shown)


packages/playout-gateway/src/tsrHandler.ts
```diff
@@ -382,7 +383,7 @@ export class TSRHandler {
 		this.tsr.connectionManager.on('connectionEvent:debugState', (id, state) => {
 			const device = this._coreTsrHandlers[id]?._device
 
-			if (device?.debugState && this._coreHandler.logDebug) {
+			if (device && device.debugState && this._coreHandler.logDebug) {
```
Julusian: This change could be problematic, various linters will tell us to revert this.

Im curious why it needs to be split like this, and a comment stating this is a necessary change would be good to avoid it being reverted

packages/server-core-integration/src/lib/coreConnection.ts
```diff
@@ -212,6 +212,16 @@ export class CoreConnection<
 	async createChild(coreOptions: ChildCoreOptions): Promise<CoreConnectionChild<PubSubTypes, PubSubCollections>> {
 		const child = new CoreConnectionChild<PubSubTypes, PubSubCollections>(coreOptions)
 
+		this._children.push(child)
+
+		// Set the max listeners to a higher value, so that we don't get warnings:
+		this.setMaxListeners(
```
Julusian, line 218: I'm wondering if we should do this or should we simply set it to unlimited?

I can see pros and cons both ways, so I am fine with either

packages/playout-gateway/src/tsrHandler.ts
nytamin: This must have been a bad commit from my side. I'll revert that line!

packages/server-core-integration/src/lib/coreConnection.ts
nytamin: I was leaning towards that it'd be better to be conservative and keep the `maxListeners` in effect, rather than disabling it altogether.

PR #1524, opened 2025-09-10
feat: Rundown information in actions

rjmunro:
## About the Contributor

## Type of Contribution

This is a Feature

## Current Behavior

N/A

## New Behavior

ActionContexts now contain a method getUpcomingParts which will return a list of upcoming parts after the next one, allowing actions to be taken based on what items may be coming up further ahead than the next part instance.

## Testing

### Affected areas

This PR affects what information is easily available to blueprints.

## Status

### DIFF
--- a/packages/job-worker/src/blueprints/__tests__/context-OnTakeContext.test.ts
+++ b/packages/job-worker/src/blueprints/__tests__/context-OnTakeContext.test.ts
@@ -6,6 +6,8 @@ import { mock } from 'jest-mock-extended'
 import { PartAndPieceInstanceActionService } from '../context/services/PartAndPieceInstanceActionService.js'
 import { OnTakeContext } from '../context/index.js'
 import { DBPart } from '@sofie-automation/corelib/dist/dataModel/Part'
+import { PartId, RundownId, SegmentId } from '@sofie-automation/corelib/dist/dataModel/Ids'
+import { protectString } from '@sofie-automation/corelib/dist/protectedString'
 import { PlayoutModelImpl } from '../../playout/model/implementation/PlayoutModelImpl.js'
 
 describe('Test blueprint api context', () => {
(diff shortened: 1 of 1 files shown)


packages/blueprints-integration/src/context/partsAndPieceActionContext.ts
```diff
@@ -48,6 +49,12 @@ export interface IPartAndPieceActionContext {
 	/** Gets the Segment. This primarily allows for accessing metadata */
 	getSegment(segment: 'current' | 'next'): Promise<IBlueprintSegment | undefined>
 
+	/** Get a list of the upcoming Parts in the Rundown, in the order that they will be Taken */
+	getUpcomingParts(
+		/** The max number of parts returned. */
+		limit?: number
```
jstarpl: I like that the implementation of this interface has a reasonable default of `5`, but I think the fact that there is a default has to be documented here (maybe not necessarily what the default is, but that there is a default?).

Alternatively, just mandating that you need to provide your own limit would be fine too, I think.

rjmunro: I've updated the JSDoc

packages/job-worker/src/blueprints/context/adlibActions.ts
```diff
@@ -102,6 +103,10 @@ export class ActionExecutionContext extends ShowStyleUserContext implements IAct
 		super(contextInfo, _context, showStyle, watchedPackages)
 	}
 
+	async getUpcomingParts(limit: number = 5): Promise<ReadonlyDeep<IBlueprintPart[]>> {
+		return getOrderedPartsAfterPlayhead(this._context, this._playoutModel, limit)
```
Julusian: Both of these should be running the parts through `convertPartToBlueprints`, to sanitise/clone them and make them safe for blueprints.

I'm a  little surprised that typescript didn't flag this, I thought that some of the properties were incompatible.. weird

PR #1532, opened 2025-10-02
feat: add notifications to lsg

justandras:
## About the Contributor

This pull request is posted on behalf of CBC

## Type of Contribution

This is a: Feature

## Current Behavior

- Currently the Live Status Gateway doesn't provide a way to subscribe to notifications present in Sofie.
- The notifications are only exposed to the UI and not available to the gateways.

## New Behavior

- This PR adds a new topic to the Live Status Gateway that exposes the active notifications.
- The publications for the Notification collections are now available to the gateways.

## Testing

### Affected areas

* This PR affects the AsyncApi schema and Live Status Gateway API types.
* This PR affects the Live Status Gateway.
* This PR adds new publications to `corelib`
* This PR removes publications from `meteor-lib` that are still available, because now they will carry over from `corelib`
* This PR affects the `webui`

## Time Frame

* Not urgent

## Status

- [X] Relevant documentation (code comments, [system documentation](https://sofie-automation.github.io/sofie-core//)) has been added / updated.


packages/live-status-gateway/src/collections/notifications/notificationsHandler.ts
```diff
@@ -0,0 +1,101 @@
...
+		if (this._rundownNotificationsHandler) {
+			const rundownDocs = this._rundownNotificationsHandler.getPublishedDocs()
+			for (const d of rundownDocs) {
+				if (d._id && !merged.has(unprotectString(d._id))) {
+					merged.set(unprotectString(d._id), d)
+				}
+			}
+		}
+
+		const newNotifications = Array.from(merged.values())
```
ianshade: I'm curious why this rather complex merging logic is needed. Are id conflicts expected? Is there a particular rule that says the first one should win?

packages/live-status-gateway/src/topics/helpers/notification/notificationTarget/toNotificationTarget.ts
```diff
@@ -0,0 +1,98 @@
...
+	switch (dbTarget.type) {
+		case DBNotificationTargetType.PARTINSTANCE:
+			return toNotificationTargetPartInstance(dbTarget)
+		case DBNotificationTargetType.RUNDOWN:
+			return toNotificationTargetRundown(dbTarget)
+		case DBNotificationTargetType.PLAYLIST:
+			return toNotificationTargetPlaylist(dbTarget)
+		case DBNotificationTargetType.PIECEINSTANCE:
+			return toNotificationTargetPieceInstance(dbTarget)
+	}
```
Julusian, line 38: It would be good to have a default case here. While I think that typescript would flag this when new enum values get added, an already built lsg won't be aware will have issues due to this function returning undefined when it is not declared to do so.
```suggestion
    default:
      assertNever(dbTarget)
      return undefined
	}
```

packages/live-status-gateway/src/topics/helpers/notification/notificationTarget/toNotificationTarget.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function toNotificationTargetBase(dbTarget: DBNotificationTarget): Pick<NotificationTarget, 'type' | 'studioId'> {
+	return literal<Pick<NotificationTarget, 'type' | 'studioId'>>({
+		type: toNotificationTargetType(dbTarget.type),
+		studioId: unprotectString(dbTarget.studioId),
+	})
+}
+
+function toNotificationTargetType(dbTargetType: DBNotificationTargetType): NotificationTargetType {
```
Julusian: Nitpick: I am not a fan of this function, it feels wordy when it could be done simpler as part of building the objects inside of `toNotificationTargetRundown` (and others)

packages/live-status-gateway/src/collections/notifications/notificationsHandler.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	init(handlers: CollectionHandlers): void {
+		super.init(handlers)
+
+		this._playlistNotificationsHandler =
+			handlers.playlistNotificationsHandler as unknown as PlaylistNotificationsHandler
+		this._rundownNotificationsHandler =
+			handlers.rundownNotificationsHandler as unknown as RundownNotificationsHandler
+
+		if (this._playlistNotificationsHandler) {
```
Julusian: This may be a pattern that is already in use elsewhere, but is this good behaviour? (In which case, you dont need to answer)

I suspect that if this is ever false, then this handler won't initialise and it will never initialise as there is a bug with the sequence of operations at startup?

In which case it would be good to at least log that this has failed, perhaps even throwing an error would be reasonable?  
I am worried that if this does fail, it will do so silently, and leave the consumer confused as to why it doesnt work with no indication in the logs that it did fail

Julusian (approved):

Issue #1450, opened 2025-05-12
RFC: Simplify Studio & ShowStyle relationship


Julusian:
### About Me

This RFC is posted on behalf of Superfly.

### Use Case

We feel that Sofie has unnecessary complexity and confusion surrounding the handling of multiple studios. 

The main aim of this is to solve the issue of there not being any shared config between the ShowStyle and Studio during the applyConfig blueprint methods. This makes it incredibly hard to generate sensible SourceLayers and LayerMappings without either making assumptions or requiring config to be defined in both places.

For example, each Studio Monitor needs at least one SourceLayer to put the pieces on, and it translates to some hardware resource (such a CasparCG channel or chef window), so needs mappings to be created too.
Ideally this could be driven by a blueprint config table in the Studio settings, with the ShowStyle blueprints able to use this config table too. This is not possible today.

### Proposal

The challenge with the config, is that there is currently a M-N relationship between ShowStyles and Studios. Each Studio can use any number of ShowStyles, and each ShowStyle can be used in any number of Studios.

For this reason, the ability to access the Studio config was never given to the ShowStyles, as handling this correctly in blueprints will be complicated. The hope was that it could be simplified, then config would be exposed.

I see 3 possible approaches to making this possible:

Limit Sofie to only having one Studio

Limit ShowStyles to only be usable by one Studio

Provide an array of Studio configs to the ShowStyle blueprint

### 1) One Studio

As far as we are aware, no installation of Sofie has more than one Studio within the installation. So we want to question whether the current structure is correct, or if it should be simplified.

From what we have seen this is a combination of:

Wanting to roll out updates to a single Studio at a time

Isolation. Any studios that would make sense to group are often backups for each other. Sharing a Sofie installation adds risks that an issue in the shared installation would make both studios unusable.

So perhaps Sofie should enforce this and be limited to one studio?

As a developer of Sofie, I sometimes wonder how well Sofie will handle multiple studios. As no-one has really used it, are there areas of the code which are not checking the studio boundary correctly, and so could leak data/operations across? By introducing this limit, this would no longer be something to worry about.

Short term, this would be simple to limit the creation of additional studios and let developers make assumptions about there being one studio going forward.
Long term, various portions for handling multiple studios could be cleaned up.

### 2) One Studio for a ShowStyle

Another option is to limit it so that a ShowStyle is only being able to be used in one Studio.

This directly solves the blueprint config issue without trying to change Sofie much.

[...]

PR #1534, opened 2025-10-06
feat: limit to single studio #1450

Julusian:
## About the Contributor

This pull request is posted on behalf of Superfly

## Type of Contribution

This is a:  Feature / Code improvement / Documentation improvement 

## Current Behavior

Sofie currently supports any number of studios

## New Behavior

As discussed in #1450, this starts to remove support for having more than 1 studio.  
At this stage it simply attempts to stop it from being possible to create more than 1, or to delete the last studio, and adds a migration to complain if there is more than 1.

This is the first stage of this, once this is merged future changes could start to assume that there is only one studio in the system.

## Testing

## Time Frame

## Status

### DIFF
--- a/packages/webui/src/client/ui/Settings/SettingsMenu.tsx
+++ b/packages/webui/src/client/ui/Settings/SettingsMenu.tsx
@@ -86,12 +86,16 @@ function SettingsMenuStudios() {
 		MeteorCall.studio.insertStudio().catch(catchError('studio.insertStudio'))
 	}, [])
 
+	// An installation should have only one studio https://github.com/Sofie-Automation/sofie-core/issues/1450
+	const canAddStudio = studios.length === 0
+	const canDeleteStudio = studios.length > 1
+
 	return (
 		<>
-			<SectionHeading title={t('Studios')} addClick={onAddStudio} />
+			<SectionHeading title={t('Studios')} addClick={canAddStudio ? onAddStudio : undefined} />
 
 			{studios.map((studio) => (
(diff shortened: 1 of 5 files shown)


packages/webui/src/client/ui/Settings/SettingsMenu.tsx
```diff
@@ -86,12 +86,16 @@ function SettingsMenuStudios() {
...
+	const canDeleteStudio = studios.length > 1
+
 	return (
 		<>
-			<SectionHeading title={t('Studios')} addClick={onAddStudio} />
+			<SectionHeading title={t('Studios')} addClick={canAddStudio ? onAddStudio : undefined} />
 
 			{studios.map((studio) => (
-				<SettingsMenuStudio key={unprotectString(studio._id)} studio={studio} />
+				<SettingsMenuStudio key={unprotectString(studio._id)} studio={studio} canDelete={canDeleteStudio} />
```
jstarpl, line 98: It seems that after resolving migration steps, `canDelete` will always be false? If that's the case, does it still make sense to propagate this property rather than removing the button altogether?

Julusian: The migration relies on the user to 'manually' fixup the system if there is more than 1 studio.
This button is the easiest way to do that, the other being to do it directly in mongodb.

So at this stage of being a soft-enforcmenet of 1 studio, I thought this was best to make it easy to resolve.
But yes as some point when we it is strongly enforced that there is 1 studio, then yes this should go away

PR #1537, opened 2025-10-08
Fix determining TSR version when running in certain configurations

rjmunro:
## Type of Contribution

This is a Bug fix / Code improvement for compatibility with later node versions

## Current Behavior

Using `import` with a json file is banned in node 20 unless you use `with { type: 'json' }`, which is not supported in node 16, our current build target.

Question: Does using import freeze the value at sofie-core build time, not reflect the version that yarn has chosen to install? If so, is that the right thing to do?

## New Behavior

Use require to import the `package.json` file.

## Testing

### Affected areas

This PR affects building blueprints.

## Time Frame

## Status

### DIFF
--- a/packages/shared-lib/src/tsr.ts
+++ b/packages/shared-lib/src/tsr.ts
@@ -1,5 +1,10 @@
 import * as TSR from 'timeline-state-resolver-types'
 export { TSR }
 
-import * as tsrPkgInfo from 'timeline-state-resolver-types/package.json'
-export const TSR_VERSION: string = tsrPkgInfo.version
+// Dynamically import package.json to avoid import attributes error
+const TSR_VERSION = require('timeline-state-resolver-types/package.json').version
+
+// Below line should work in Node.js 18+ and with bundlers that support import attributes, and replace the above
+// import { version as TSR_VERSION } from 'timeline-state-resolver-types/package.json' with { type: 'json' }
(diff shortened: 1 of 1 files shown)


Julusian:
I'm wondering where/how this is an issue, as these libs officially only suport node22+ in release52+.
https://github.com/Sofie-Automation/sofie-core/blob/9bd232e8f0561a46db8cc6143c5353d7fa531206/packages/shared-lib/package.json#L31-L33

The build config might be doing so with support for older, but that will be changed whenever someone feels like it (it often lags a bit behind)

The latest r53 build of this file (from https://www.npmjs.com/package/@sofie-automation/shared-lib/v/1.53.0-nightly-release53-20250924-143640-007a9da.0?activeTab=code) is:
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.TSR_VERSION = exports.TSR = void 0;
const tslib_1 = require("tslib");
const TSR = tslib_1.__importStar(require("timeline-state-resolver-types"));
exports.TSR = TSR;
const tsrPkgInfo = tslib_1.__importStar(require("timeline-state-resolver-types/package.json"));
exports.TSR_VERSION = tsrPkgInfo.version;
//# sourceMappingURL=tsr.js.map
```
so I am curious what this PR will actually change

PR #1541, opened 2025-10-16
feat: Add ESLint rule to standardise file extensions on imports

rjmunro:
## Type of Contribution

This is a code improvement

## Current Behavior

Between release52 and release53, particularly in d02f4162, the standard for imports seemed to be changed to include `.js` on the end of the filename. This caused merge conflicts when I was trying to rebase a branch targeted at release52 onto release53.

## New Behavior

Use an eslint plugin to enforce extensions on file imports. This enabled me to automatically resolve a bunch of merge conflicts using my [git resolve-formatting-conflicts](https://github.com/rjmunro/jamies-git-tools/blob/main/doc/git-resolve-formatting-conflicts.md) tool.

Currently I have applied this is just to the webui package because that's where I was working. It could probably be done more widely.

## Testing

### Affected areas

This PR affects code formatting and it's enforcement.

## Time Frame

* Not urgent

## Other Information

It looks like there is a script to do the modification here: https://github.com/Sofie-Automation/sofie-core/pull/1432/commits/88c500fff3821b76ce27167bae9d01070711675c. The advantage of this is that it is part of normal ESLint formatting, so IDEs and will highlight and auto-correct, CI will catch it, and it will work with tools like [git resolve-formatting-conflicts](https://github.com/rjmunro/jamies-git-tools/blob/main/doc/git-resolve-formatting-conflicts.md) without any further customisation.

## Status

## Summary

This pull request introduces an ESLint rule to standardize and enforce file extensions (`.js`, `.jsx`) on import statements within the webui package, addressing recurring merge conflicts caused by inconsistent import formats between release versions.

### Key Changes

**ESLint Configuration**
- Added `eslint-plugin-n` to the ESLint configuration in `packages/webui/eslint.config.mjs`
- Configured the `n/file-extension-in-import` rule set to 'error' to enforce extensions on imports
- Extended the plugin settings with `tryExtensions` including `.js`, `.cjs`, `.mjs`, `.json`, `.node`, `.ts`, `.cts`, `.mts`, `.d.ts`, and `.tsx`

**Import Path Updates**
Systematically updated import statements across approximately 40 files in `packages/webui/src/client/` to include explicit file extensions:
- Updated relative imports to include `.js` extensions (e.g., `'./lib'` → `'./lib.js'`, `'../../collections'` → `'../../collections/index.js'`)
- Updated JSX/TSX component imports to include `.jsx` or `.js` extensions as appropriate
- Affected files include components, hooks, and utilities across various subdirectories (RundownView, Settings, TestTools, Shelf, etc.)

### Impact

- No functional logic changes; all modifications are module resolution path updates
- Leverages the ESLint plugin to automatically enforce and fix import extensions, integrating with IDE auto-correct and CI checks
- Helps prevent future merge conflicts caused by inconsistent import formatting between releases
- Changes are compatible with existing code structure and control flow


packages/webui/eslint.config.mjs
```diff
@@ -11,6 +12,8 @@ const tmpRules = {
 	'@typescript-eslint/unbound-method': 'off',
 	'@typescript-eslint/no-misused-promises': 'off',
 	'@typescript-eslint/no-unnecessary-type-assertion': 'off',
+	'import-extensions/require-extensions': ['error', { expectedExtensions: ['js'] }],
```
Julusian: These shouldn't be added here, as these are permanent additions, not `Temporary rules to be removed over time`

Julusian (commented):
It looks like for the backend code there is already a rule in place for this. Perhaps this should use the same plugin instead of adding another one?

I believe that rule is https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/file-extension-in-import.md

That plugin was omitted from the webui as its not using node at runtime, so some of the rules dont make sense. But maybe there are a handful of rules in that plugin which do make sense to enable?

Julusian:
> I wonder why we have it set at all in sofie-standard-code-preset. 

For any packages which are actually generated as ESM, the extensions are necessary. So I thought it easier to enforce everywhere rather than having that variation based on how the project is generated.  
We have a couple of things which are ESM, and I want to push more things across. Most notably, vite gets a bit upset that
meteor-lib and other dependencies in the mono-repo arent esm, which is why it doesnt always detect all changes without a restart.

> It incorrectly added a few more .js extensions where the import was a folder which caused errors.

Yeah, in ESM you can't import a folder, those imports will need to be adjusted to point to the index file within the folder

PR #1543, opened 2025-10-20
chore: refactor lsg asyncapi schema

justandras:
## About the Contributor

## Type of Contribution

This is a: Code improvement / Documentation improvement

## Current Behavior

1. Currently the Live Status Gateway's AsyncApi schemas do not reuse definitions, it's hard to maintain.
2. Examples have to be defined inline, without the ability to include already defined examples in a nested object.
3. All messages are logically grouped in the same `/` channel
4. Types for messages to be sent by the client are not generated

## New Behavior

1. Added a script to resolve example references as it's currently not supported by our AsyncApi version.
5. Split up the schemas into smaller chunks, established inheritance between schemas to minimize duplicate definitions.
6. Introduced the concept of channels to organize the schemas, improve documentation readability
7. Added a script that creates a single `asyncapi.yaml` file for better compatibility with tools like Postman and other parsers.
8. Types are now generated for messages that are sent by the client.

## Testing

### Affected areas

This PR is only an internal refactor of the schemas, there is no breaking change in the generated types.

This PR affects:
1. The schema source files for the Live Status Gateway.
2. The type, docs and server generation for the Live Status Gateway API.

## Time Frame

This Refactor is critical for us, we have other PRs, feature request that depend on it.

## Other Information

Due to AsyncApi limitations, some types now have to have an optional property `additionalProperties`. This will remain unused, there is no clean workaround to have inheritance between schemas that we know of and disallow `additionalProperties` at the same time.

We the SuperFly team have discussed it internally and feel like it's a worthy compromise.

## Status

- [X] Relevant unit tests has been added / updated.
- [X] Relevant documentation (code comments, [system documentation](https://sofie-automation.github.io/sofie-core//)) has been added / updated.


packages/live-status-gateway-api/scripts/resolve-nested-examples.mjs
ianshade, line 1: I didn't look too deep into this file, but could you please explain shortly why we need this implementation? What does it do that an existing package like @apidevtools/json-schema-ref-parser does not?

packages/live-status-gateway-api/src/generated/asyncapi.yaml
ianshade, line 1: After checking out this code and running `yarn gendocs`, this file got overwritten with slightly different formatting. It might become a mild annoyance if this file keeps getting reformatted back and forth. Can it be solved in the scripts?
image

packages/live-status-gateway-api/api/asyncapi.yaml
```diff
@@ -26,114 +26,66 @@ tags:
...
-      messageId: packages
-      description: Status of Packages expected by Pieces
-      payload:
-        $ref: './schemas/packages.yaml#/$defs/packages'
-    buckets:
-      name: buckets
-      description: Buckets in Studio
-      payload:
-        $ref: './schemas/buckets.yaml#/$defs/buckets'
+          - messageId: pong-typeGen
```
ianshade: I'm wondering where those `typeGen` suffixes come from, what do they mean, what do they do

packages/live-status-gateway-api/api/asyncapi.yaml
```diff
@@ -26,114 +26,66 @@ tags:
...
-      operationId: processRootMessage
-      message:
-        oneOf:
-          - $ref: '#/components/messages/ping'
-          - $ref: '#/components/messages/subscribe'
-          - $ref: '#/components/messages/unsubscribe'
     subscribe:
-      description: Messages that are received from the API
+      summary: Slash Messages
+      description: Messages that are received from the API - This channel is needed to retain the generated Slash type from the old schema structure. Eventually this should be removed.
```
ianshade: As a reader of those docs, I don't know what to do with this information, and the Slash Messages channel in general.
image

packages/live-status-gateway-api/api/asyncapi.yaml
justandras: Before the refactor all messages were grouped under the slash channel. We were also exporting a Slash type that is the union of all messages the client can receive. To retain this and the new cleaner structure, we need to have the slash channel also defined. Since the messages exist in the channel level already we are not allowed to reuse the same message Id in a different channel. The only purpose of the Slash channel is to keep the type changes minimal, in case someone is using the Slash type in their LSG client. I hope this also covers your other question regarding the Slash channel.

packages/live-status-gateway-api/api/asyncapi.yaml
justandras: I was considering adding a script that removes the slash channel for the doc generation, but I wasn't too sure about it. Do you think we should ignore it during doc generation?

PR #1567, opened 2025-11-27
feat: Add links to different studio screens and add a multiview for development and testing

rjmunro:
## Type of Contribution

This is a Feature

## Current Behavior

If you start at the URL of, for example, a directors screen (e.g. http://localhost:3005/countdowns/studio0/director) and go up a level (e.g. http://localhost:3005/countdowns/studio0/), it would redirect you to the home page of Sofie. Also there are several available views for a studio ([Presenter Screen](http://localhost:3005/countdowns/studio0/presenter), [Director Screen](http://localhost:3005/countdowns/studio0/director), [Overlay Screen](http://localhost:3005/countdowns/studio0/overlay), [Camera Screen](http://localhost:3005/countdowns/studio0/camera)
[Prompter](http://localhost:3005/prompter/studio0), [Active Rundown](http://localhost:3005/activeRundown/studio0) and it's hard to remember them all

## New Behavior

If you go to a URL like http://localhost:3005/countdowns/studio0/ you get a list of the main views available:

image

I've also made a multiView which shows most of them all in iframes, which is great for testing.

image

Currently this page is not itself listed anywhere. A page should be added to enumerate all the available studios and link to these views. But we may want to move some things around and add some redirects first because having most of these views under /countdowns/ possibly doesn't make sense.

## Testing

### Affected areas

This PR adds a couple of per-studio pages, neither of which has any active code, just links to already existing pages.

## Time Frame

## Status

### DIFF
--- a/packages/webui/src/client/ui/ClockView/ClockView.tsx
+++ b/packages/webui/src/client/ui/ClockView/ClockView.tsx
@@ -31,7 +31,7 @@ export function ClockView({ studioId }: Readonly<{ studioId: StudioId }>): JSX.E
 
 	return (
 		<Switch>
-			<Route path="/countdowns/:studioId/presenter">
+			<Route exact path="/countdowns/:studioId/presenter">
 				{playlist ? (
 					<RundownTimingProvider playlist={playlist}>
 						<PresenterScreen playlistId={playlist._id} studioId={studioId} />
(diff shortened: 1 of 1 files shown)


jstarpl (commented):
I don't see any particular problems, but I do see that while there's a header on the main "countdowns landing page":

Image

There is none if I go to a non-existant countdown variant, but I still get the list:

Image

It's not a big thing, but I think it would be nice if we either kept the fallback redirect, in case someone opens a page that doesn't exist, or have the header always if our new policy is to show this page now for anything starting with `/countdowns`.

PR #1593, opened 2026-01-08
Fix some SCSS warinings

rjmunro:
## Type of Contribution

This is a Code improvement

## Current Behavior

A lot of SCSS warnings are shown, e.g. when running `yarn dev` on sofie-core, mostly due to deprecation of certain SCSS features.

## New Behavior

Warnings from external libraries are suppressed; a few other warnings have been eliminated by changing the code where the changes are small. Many warnings still remain, mostly due to use of `@import` rather than `@use` but these are a bit harder to fix.

I have also removed a redundant and broken `hidden` feature from the PreviewPopup. When we hide the popup, we unmount it in React, we don't use the hidden feature. It wouldn't have worked anyway because use used `visibility: none`, which is not valid css, it's a confusion of `visibility: hidden` and `display: none` which are similar but slightly different concepts.

## Testing

### Affected areas

This affects webui.

## Time Frame

## Status

This PR reduces SCSS deprecation warnings during development by suppressing dependency warnings and fixing small invalid SCSS in the web UI.

### Key Changes

- packages/webui/vite.config.mts
	- Added css.preprocessorOptions.scss.quietDeps: true to silence deprecation warnings from external SCSS (e.g., Bootstrap) while still showing warnings from the repo's own SCSS.

- packages/webui/src/client/ui/PreviewPopUp/PreviewPopUp.scss
	- Replaced @import with @use '../../styles/itemTypeColors';
	- Updated mixin usage to the namespaced form: @include itemTypeColors.item-type-colors;
	- Removed the invalid/unused hidden modifier (no preview-popUp--hidden definition remains).

- packages/webui/src/client/ui/PreviewPopUp/PreviewPopUp.tsx
	- Removed the public hidden?: boolean prop and eliminated the conditional application of the preview-popUp--hidden class (the popup is unmounted by React when hidden).

### Notes

- Many SCSS warnings remain (primarily from files still using @import); this PR focuses on suppressing dependency warnings and making small, feasible fixes.
- Author marked the PR ready for review and tested functionality locally; no unit test changes were included.

### DIFF
--- a/packages/webui/src/client/ui/PreviewPopUp/PreviewPopUp.scss
+++ b/packages/webui/src/client/ui/PreviewPopUp/PreviewPopUp.scss
@@ -108,7 +108,7 @@
 			margin-left: 2%;
 			margin-top: 7px;
 			flex-shrink: 0;
-			@include itemTypeColors.item-type-colors();
+			@include itemTypeColors.item-type-colors;
 		}
 
 		.preview-popUp__element-with-time-info__text {
@@ -397,7 +397,7 @@
 		}
 
 		& > * {
-			@include itemTypeColors.item-type-colors();
+			@include itemTypeColors.item-type-colors;
 		}
 
 		.video-preview__label {


packages/webui/vite.config.mts
```diff
@@ -59,6 +59,16 @@ export default defineConfig(({ command }) => ({
 		},
 	},
 
+	css: {
+		preprocessorOptions: {
+			scss: {
+				// Silence deprecation warnings from Bootstrap and other libraries
+				// These are caused by Bootstrap 5.x not yet fully supporting Dart Sass 2.x
+				silenceDeprecations: ['import', 'global-builtin', 'color-functions', 'mixed-decls'],
```
Julusian: should this be `quietDeps: true,` instead? that seems to hide all the warnings, and might be better as it wont risk us adding any new silenced deprecations to our code?
but I'm not seeing any warnings from the other change you made, so maybe this hides too much too?

PR #1597, opened 2026-01-14
chore: reconfigure tsc to run at the root, not for each package individually

Julusian:
## About the Contributor

This pull request is posted on behalf of Superfly

## Type of Contribution

This is a: Code improvement

## Current Behavior

When building the TS, we are using lerna to spawn tsc for each package individually.

This makes it rather memory hungry, and if an error is encountered typically means running it all again

I find the watcher output quite confusing, you can never quite be sure if everything is happy or if only the package you just edited is happy and other ones dependent on it are still erroring.

## New Behavior

This now utilises the typescript projects and references functionality. There is a single 'root' tsconfig inside packages, which will build and typecheck everything in one go.

This means that there is one stream of type error output, which shows everything in one go with one count of number of errors. Fixing one error will reflect that in all referencing projects instead of burying errors higher up the terminal history.

A crude memory test locally, spawning in watch mode:
* Before: 6.3GB of memory used once it stabilises
* After: 1.5GB once stabilised peak at 1.8GB (this is also type checking meteor, which was not done before)

I have not touched the openapi package in this, as that was commonly excluded to avoid a dependency on java.

Some of the build workflows have been refined a bit so that every branch will build the docker images, just not push them unless the existing rules are met. I did this as I noticed that I had caused some build failures that only appeared on the superfly fork, as the main repositry is only setup to build images for releases

As a follow up, I intend to give the same treatment to linting and tests, as I find those have similar pain points

## Testing

## Status

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -130,6 +130,10 @@ jobs:
       - name: Build libs
         run: |
           yarn build:packages
+      - name: Build webui
+        run: |
+          cd packages/webui
+          yarn build
       - name: Persist Built Version information
         run: |
           cd meteor


packages/job-worker/tsconfig.build.json
```diff
@@ -14,6 +15,13 @@
 		"resolveJsonModule": true,
 		"types": ["node"],
 		"skipLibCheck": true,
-		"esModuleInterop": true
-	}
+		"esModuleInterop": true,
+		"composite": true
+	},
+	"references": [
+		//
```
nytamin, line 22: this is a syntax error, right?

nytamin (approved):

packages/job-worker/tsconfig.build.json
Julusian: No, typescript allows comments in its json files.

I am using comments here to force prettier to format this array across multiple lines, to make it easier to read

PR #1710, opened 2026-04-01
Fix for AB playback lookahead player assignment

scriptorian:
This PR is posted on behalf of EVS Broadcast Equipment.

## Type of Contribution

This is a:

Bug fix

## Current Behaviour

Assigning a player for a lookahead picked from an array of players that would be free soon. The logic didn't take correct account of current piece end times so the wrong player was assigned

## New Behaviour

The soon to be free players are now sorted by end time so that the lookahead is assigned to the player that will be free first

## Testing

### Affected areas

This PR affects the ab playout logic. It particularly affects systems that use an aux output to show the next part.

## Time Frame

## Status

### DIFF
--- a/packages/job-worker/src/playout/abPlayback/__tests__/abPlaybackResolver.spec.ts
+++ b/packages/job-worker/src/playout/abPlayback/__tests__/abPlaybackResolver.spec.ts
@@ -665,7 +665,7 @@ describe('resolveAbAssignmentsFromRequests', () => {
 				name: 'c',
 				start: 10000,
 				end: 15000,
-				playerId: 1,
+				playerId: 2,
 				pieceNames: ['p3'],
 			},
 			// lookaheads (in order of future use)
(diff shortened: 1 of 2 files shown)


packages/job-worker/src/playout/abPlayback/__tests__/abPlaybackResolver.spec.ts
```diff
@@ -639,6 +639,57 @@ describe('resolveAbAssignmentsFromRequests', () => {
 		expectGotPlayer(res, 'd', undefined)
 	})
 
+	test('Autonext pgm/pvw', () => {
```
Julusian: Isn't this the bad test that I was going to look into?
It seems to still be asserting that they get assigned as 2, 1, 1, 2

scriptorian: Test now updated to match now resolved case

Julusian: This could do with a more descriptive name of what it is testing, it isnt clear from this test that it is testing the assignment of the lookahead

Julusian: Also the values in this test make this be not a good test. If I comment out the fix, the test still passes.
Perhaps the player ids should simply be swapped around? Or maybe it needs to do some more assertions to ensure that it is actually assigning in a predictable order and not passing by luck

scriptorian: I've updated the test, changed the test name and it successfully fails without the fix

Julusian (approved):

Issue #1752, opened 2026-05-20
Deprecation RFC: Remove "Rewind" button in the Sidebar


hummelstrand:
### Suggested by

This RFC is championed by _SuperFly.tv_

### Use Case

<img width="120" height="126" alt="Image" src="https://github.com/user-attachments/assets/3e7ebfc4-6e03-4b50-a35f-7a17046fe1f5" />

This "Rewind" button in the Sofie GUI is a remnant of earlier times when played segments ended up scrolled all the way to the right. It was an inelegant quick fix to the big GUI shortcoming of not allowing the user to "zoom" segments, and should have been deprecated as soon as we added auto-scrolling segments back to the beginning as soon as they had been played.

Since the _per-segment zoom levels_ and controls were introduced many years ago, each segment now initially displays in its entirety, and the user can always return to this state by the means of the "Show all" button between the – and the + buttons.

Attaching an old screenshot were one can see all segments displayed at the same scale, and the mini-rendering of each segment below that would allow the user to scroll horizontally.

<img width="3754" height="2418" alt="Image" src="https://github.com/user-attachments/assets/cc9f7466-aab1-477a-871b-190ee4d2ecb4" />

### Proposal

This button should be deprecated on the following grounds:

- The functionality is no longer needed after two UX changes mentioned above.
- The icon is misleading and indicates that it "rewinds" something, which it no longer does.
- The button is always active, even though it often doesn't produce any results.

### Process

The Sofie Team will evaluate this Deprecation RFC following their approval workflow.

Stakeholders have until **2026-06-20** to comment or challenge the proposal

- [ ] Outcome: A unison approval from stakeholders has been reached.
- [ ] Outcome: No objections raised within the 1 month limit, so the proposal is implicitly approved.

PeterC89:
@hummelstrand TSC approved ✅

RannveigNC:
I will bring this up with our producers, hopefully by the end of this week

RannveigNC:
As expected none of our producers seems to use this button, so I would say go ahead an remove it

hummelstrand:
PR here: https://github.com/Sofie-Automation/sofie-core/pull/1770

PR #1770, opened 2026-06-11
Removed the user-facing GUI button "Rewind Segments" as per RFC #1752

hummelstrand:
## About the Contributor
This PR and RFC are championed by SuperFly.tv

## Type of Contribution

This is a:

GUI improvement

## Current Behavior

As per the [RFC](https://github.com/Sofie-Automation/sofie-core/issues/1752).

## New Behavior

As per the [RFC](https://github.com/Sofie-Automation/sofie-core/issues/1752).

## Testing

### Affected areas

* This PR affects the main Sofie GUI sidebar.

## Time Frame

## Other Information

The functionality still exists in the code, this PR only removes the user-facing button.

## Status

### DIFF
--- a/packages/webui/src/client/lib/ui/icons/rewindAllSegmentsIcon.tsx
+++ b/packages/webui/src/client/lib/ui/icons/rewindAllSegmentsIcon.tsx
@@ -1,8 +1,0 @@
-export const RewindAllSegmentsIcon = (): JSX.Element => (
-	<svg width="18px" height="18px" viewBox="0 0 512 300" fill="none" xmlns="http://www.w3.org/2000/svg">
-		<path
-			fill="currentColor"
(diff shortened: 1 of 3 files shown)


jstarpl (commented):
The code itself is fine, and the RFC is "Concluded", but I do have one note: the button did two things: rewind the segments & execute "zoom fit to view" on all Segments. I think it would be nice if there was still a way to reset the zoom state of all the segments, if that got messed up somehow. Or maybe we should make the Segment scroll & zoom level not persistent? So that all it takes to fix it, is to scroll away and come back?

hummelstrand:
While a "zoom all segments to fit" button might be useful for a power user, I'd argue against resetting the zoom level on each segment that is scrolled out of view. 

If the user deliberately sets the zoom level of a segment it would be really annoying to have Sofie change it back to default as soon as you scrolled. Use case: comparing two parts in separate segments. And by extension; if the user has deliberately zoomed in to a segment, that means the user has found the GUI elements for the segment zoom control, meaning that if they want to change it again they should already know where to find the controls for doing so.

A power user feature that could potentially be added would be to add a keyboard modifier to the "zoom to fit" button which would reset all segments.

PR #1792, opened 2026-07-13
fix: adlib testing segment not retained properly during reingest

justandras:
## About the Contributor

This pull request is posted on behalf of the CBC.

## Type of Contribution

This is a: Bug fix

## Current Behavior

When Adlib Testing is active and you reingest the upcoming segments, the testing segment is not retained properly in the Playout Model, causing an error due to missing infinites and the reingest fails

Additionally, all of the segments parts are being cleared, as a result the job worker cannot properly update the parts, it has to recreate them in all cases.

Additional bug: the testing segement is not cleared when deactivating a rundown (despite it's content being cleaned up), only when activating it again.

## New Behavior

The segment is now correctly retained when the playlist is active and it's cleared when deactivating the rundown.

Updating segments through the IngestAPI no longer causes the webui to flicker.

## Testing

- [X] I have added one or more unit tests for this PR

### Affected areas

This PR affects the playout logic.

## Time Frame

## Status

- [X] Relevant unit tests has been added / updated.
- [X] Relevant documentation (code comments, [system documentation](https://sofie-automation.github.io/sofie-core/)) has been added / updated.

### DIFF
--- a/meteor/server/api/rest/v1/ingest.ts
+++ b/meteor/server/api/rest/v1/ingest.ts
@@ -780,17 +780,6 @@ class IngestServerAPI implements IngestRestAPI {
 				if (!segment) {
 					return
 				}
-
-				const parts = await this.findParts(segment._id)
-				return Promise.all(
-					parts.map(async (part) =>
-						runIngestOperation(studio._id, IngestJobs.RemovePart, {
-							partExternalId: part.externalId,
-							rundownExternalId: rundown.externalId,
-							segmentExternalId: segment.externalId,
-						})
-					)
-				)
 			})
 		)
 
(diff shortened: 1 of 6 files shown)


rjmunro (approved):
Code looks good.

I'm not sure that the "TODO" is helpful, though, unless there is some specific reason we need to update the documentation on that particular thing more than on anything else, in which case it would be good to say what that is.

justandras:
> Code looks good.
> 
> I'm not sure that the "TODO" is helpful, though, unless there is some specific reason we need to update the documentation on that particular thing more than on anything else, in which case it would be good to say what that is.

There is a lot that this function does (although it does need to do a lot of clean up)

Maybe a future refactor would be a better solution to make it more obvious why and what it's doing.

rjmunro (approved):

ianshade (approved):

